PR #3, opened 2022-09-13
SDK Shift

jpwilliams:


src/index.ts
```diff
@@ -1,147 +1,172 @@
...
+     */
+    context: Context
+  ) => Promise<Response> | Response;
+
   /**
-   * Any user data associated with the event
-   * All fields ending in "_id" will be used to attribute the event to a particular user
+   * The event payload structure for sending data to Inngest
    */
-  user?: {
```
djfarrelly, line 65: I do really like the use of generics here, but our current generated types are for the entire Event Payload as to include data and user. That may be at odds with this, unless we perhaps change how the types are generated (or use `Pick` to extract the `data` and `user`).

src/index.ts
```diff
@@ -1,147 +1,172 @@
...
    */
-  private axiosConfig: AxiosRequestConfig;
+  private readonly client: AxiosInstance;
+
+  private readonly fns: Record<string, InngestFunction<Events>> = {};
 
   /**
    * @param apiKey - An API Key for the Inngest Source API
    */
   constructor(
```
djfarrelly, line 147: While we don't necessarily need to worry about backcompat since this is pre-stable, we're going to need 2 different keys for sending (Source API key) and registering (Inngest API key). I'm definitely open to new ways that we could improve this, but `send()` wants a different API key which could be a challenge.

My gut is also saying that the constructor is heading towards being more overloaded, so perhaps just a single object argument may be the best and most flexible going forward.

src/index.ts
jpwilliams: @djfarrelly This may be me trying to standardise something that's not standardised. 😄 We can shift this to just include the entire block instead? Whichever is easiest when porting the type generation to a CLI command.

src/index.ts
jpwilliams: Great shout - I had missed the need for two keys. Cool - I'll add this!

src/index.ts
djfarrelly: Yeah - I guess we use the entire payload right now, but we're still so early, we could def be subject to change some of this if you have better ideas!

jpwilliams:
Replacing with #4.

PR #4, opened 2022-09-14
SDK Shift

jpwilliams:
Supersedes #3, but keeping both branches around for posterity.


src/components/Inngest.ts
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * The name of this instance, most commonly the name of the application it
+     * resides in.
+     */
+    name: string,
+
+    /**
+     * Inngest Source API key, used to send events to Inngest Cloud.
+     */
+    apiKey: string,
```
tonyhb, line 52: Instead of API key, let's call this an `event key`.  API keys are keys to manage your account.

src/components/Inngest.ts
```diff
@@ -0,0 +1,164 @@
...
+      validateStatus: () => true, // all status codes return a response
+      maxRedirects: 0,
+    });
+  }
+
+  #getResponseError(response: AxiosResponse): Error {
+    let errorMessage = "Unknown error";
+    switch (response.status) {
+      case 401:
+        errorMessage = "API Key Not Found";
```
tonyhb, line 85: Same: "Event key not found" etc.

src/components/Inngest.ts
```diff
@@ -0,0 +1,164 @@
...
+    return new Error(`Inngest API Error: ${response.status} ${errorMessage}`);
+  }
+
+  /**
+   * Send event(s) to Inngest
+   */
+  public async send<Event extends keyof Events>(
+    name: Event,
+    payload: Omit<Events[Event], "name"> | Omit<Events[Event], "name">[]
+  ): Promise<boolean> {
```
tonyhb, line 119: 🤔 Could this still accept one parameter for the entire payload as well as this "partial" split?

tonyhb: And what if this is a new, unseen event?  This requires that we create events in your SDK prior to using, right?

src/handlers/default.ts
```diff
@@ -191,7 +191,11 @@ export class InngestCommHandler {
 
           const stepRes = await this.runStep(fnId, stepId, req.body);
 
-          return void res.json(stepRes);
+          if (stepRes.status === 500) {
+            return void res.status(stepRes.status).json(stepRes.error);
```
tonyhb, line 195: What if this is a 4xx?

src/types.ts
```diff
@@ -120,7 +120,7 @@ export type StepRunResponse =
     }
   | {
       status: 200;
```
tonyhb, line 122: Should/could this be any 2xx response?

src/handlers/default.ts
jpwilliams: @tonyhb For this and the `2xx` question, atm this is an implementation detail.

The user just returns data, and then we give that data `status: 200` if the step runs successfully, and `status: 500` if it throws. `status: 400` is reserved for the client believing it's being called incorrectly (i.e. `POST` without specifying a `fnId` or `stepId`).

This comes back to whether we expose the `status`/`body` pattern to the user or keep it abstracted.

My view has been that it's still an implementation detail; knowing which status codes produce which behaviour (e.g. `500` = retry) feels like an extra thing for the user to learn. We should discuss this!

src/helpers/func.ts
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @example
+   *
+   * "0 0 0 1 1 * 1" // At 12:00 AM, on day 1 of the month, only in January, only in 0001
+   * "0 0 0 1 1 * 1,2" // At 12:00 AM, on day 1 of the month, only in January, only in 0001 and 0002
+   * "0 0 0 1 1 * 1,2,3" // At 12:00 AM, on day 1 of the month, only in January, only in 0001, 0002, and 0003
+   * "0 0 0 1 * * 1/4" // At 12:00 AM, on day 1 of the month, every 4 years
+   * "0 0 0 * * 0 1-4" // At 12:00 AM, only on Sunday, 0001 through 0004
+   * "0 0 0 * * * 2/4" // At 12:00 AM, every 4 years, 0002 through 9999
+   * "0 0 * * * * *" // Every hour
```
tonyhb, line 62: We don't have seconds in our cron - there's a max of 5 parameters here, such as:
`0 * * * *` for every hour.

src/remix.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  public override createHandler() {
+    return async ({
+      request: req,
+    }: {
+      request: Request;
+    }): Promise<Response> => {
+      /**
+       * Specifically for CORS (browser->site requests), only allow PUT requests
+       * from the dashboard.
```
tonyhb, line 25: FYI, we don't need this — it's our backend API that kicks off the handshake so we can leave all of our integrations CORS free!

jpwilliams: Ah aye - forgot to remove this following our changes

PR #6, opened 2022-09-22
Ensure scheme exists and handle undefined signing keys

tonyhb:
This PR ensures that undefined signing keys are handled correctly, and ensures that we always have a scheme (required for node's URL parsing).

### DIFF
--- a/src/handlers/default.ts
+++ b/src/handlers/default.ts
@@ -98,7 +98,7 @@ export class InngestCommHandler {
   private readonly inngestRegisterUrl: URL;
 
   protected readonly frameworkName: string = "default";
-  protected readonly signingKey: string;
+  protected readonly signingKey: string | undefined;
 
   /**
    * An Axios instance used for communicating with Inngest Cloud.
@@ -158,6 +158,10 @@ export class InngestCommHandler {
   // hashedSigningKey creates a sha256 checksum of the signing key with the
   // same signing key prefix.
   private get hashedSigningKey(): string {
+    if (!this.signingKey) {
+      return "";
+    }
+
     const prefix =
       this.signingKey.match(/^signkey-(test|prod)-/)?.shift() || "";
     const key = Buffer.from(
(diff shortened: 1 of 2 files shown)


src/handlers/default.ts
```diff
@@ -171,7 +175,24 @@ export class InngestCommHandler {
...
-      const reqUrl = new URL(req.originalUrl, req.hostname);
+      const hostname = req.hostname || req.headers["host"];
+      if (!hostname) {
+        console.error("Unable to determine your site URL to host Inngest.");
+        return res.status(500).json("Unable to determine your site URL to host Inngest.");
+      }
+
+      let scheme = "";
+      if (hostname.indexOf("://") === -1) {
+        scheme = (process.env.NODE_ENV === "development" || !req.connection.encrypted) ? "http://" : "https://";
```
djfarrelly, line 186: I couldn't find `req.connection` cited in Express and it's marked as [deprecated in the Node docs](https://nodejs.org/dist/latest-v16.x/docs/api/http.html#requestconnection) - Can we use `req.protocol` instead?

* https://expressjs.com/en/api.html#req.protocol
* https://nodejs.org/dist/latest-v16.x/docs/api/http.html#requestprotocol

src/handlers/default.ts
```diff
@@ -171,7 +175,24 @@ export class InngestCommHandler {
 
   public createHandler(): any {
     return async (req: Request, res: Response) => {
-      const reqUrl = new URL(req.originalUrl, req.hostname);
+      const hostname = req.hostname || req.headers["host"];
+      if (!hostname) {
+        console.error("Unable to determine your site URL to host Inngest.");
+        return res.status(500).json("Unable to determine your site URL to host Inngest.");
```
djfarrelly, line 181: "host Inngest" feels "serve the Inngest handler"?

djfarrelly (commented):
Good catch on this issue @tonyhb - I left a couple comments, one that may be good to fix up before shipping. The other thought is that we could extract this into another function or only run this code within the `PUT` part of the switch statement as those other bits don't need it, but that change isn't necessary.

src/handlers/default.ts
```diff
@@ -98,7 +98,7 @@ export class InngestCommHandler {
   private readonly inngestRegisterUrl: URL;
 
   protected readonly frameworkName: string = "default";
-  protected readonly signingKey: string;
+  protected readonly signingKey: string | undefined;
```
jpwilliams, line 101: Is it valid to not provide a signing key? Do we not always need one?

tonyhb: we do need one, though if you use `proces.env.INNGEST_SIGNING_KEY` or similar env variables it's still possible to be undefined, so we need to denote this and handle appropriately.

src/handlers/default.ts
tonyhb: Done, thanks!

tonyhb:
@djfarrelly There's plan with @jpwilliams to create a single interface that each custom framework handler fulfils to abstract this code.  This will DRY up any URL creation and is going to be handled separately from this PR.

PR #7, opened 2022-09-23
Ensure `name`, `ts`, and `v` aren't always required when sending events

jpwilliams:
Depending on the typed event, `name`, `ts`, and `v` was always required when using `inngest.send`.

The following example would break, as the generated type shows that `ts` is always present in the event, so sending required the same.

image

This now works fine with this PR, still allowing the user to specify `ts` if they wish to.

image

It's worth doing this here rather than in type generation to ensure we handle custom event typing elegantly too.


tonyhb (approved):
I like your typescript-fu

PR #9, opened 2022-09-23
Clean up types to be less verbose inline

jpwilliams:
Cleaned up `createFunction` and `Inngest#send` types to be less verbose inline.


tonyhb (approved):
This looks solid

PR #10, opened 2022-09-23
Ensure we use app as fn prefix with SDK-based functions

tonyhb:
This reduces the chance of collisions when creating function IDs via the SDK.

### DIFF
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -118,7 +118,7 @@ export class InngestFunction<Events extends Record<string, EventPayload>> {
   #generateId(prefix?: string) {
     const join = "-";
 
-    return `${prefix || ""}-${this.#opts.name}`
+    return [prefix || "", this.#opts.name].join("-");
       .toLowerCase()
       .replaceAll(/[^a-z0-9-]+/g, join)
       .replaceAll(/-+/g, join)
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -44,9 +44,9 @@ export class InngestFunction<Events extends Record<string, EventPayload>> {
   /**
    * The generated or given ID for this function.
    */
-  public get id() {
+  public id(prefix?: string) {
     if (!this.#opts.id) {
-      this.#opts.id = this.#generateId();
+      this.#opts.id = this.#generateId(prefix);
     }
 
     return this.#opts.id;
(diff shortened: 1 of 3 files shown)


src/components/InngestFunction.ts
```diff
@@ -114,10 +115,10 @@ export class InngestFunction<Events extends Record<string, EventPayload>> {
   /**
    * Generate an ID based on the function's name.
    */
-  #generateId() {
+  #generateId(prefix?: string) {
     const join = "-";
 
-    return this.#opts.name
+    return `${prefix || ""}-${this.#opts.name}`
```
jpwilliams, line 121: Not that it's something that happens right now, but could this use `.join()` to omit a leading `-` if no `prefix` is passed?

```suggestion
    return [prefix || "", this.#opts.name].join("-");
```

src/components/InngestFunction.ts
```diff
@@ -44,9 +44,9 @@ export class InngestFunction<Events extends Record<string, EventPayload>> {
   /**
    * The generated or given ID for this function.
    */
-  public get id() {
+  public id(prefix?: string) {
```
jpwilliams, line 47: I see what you mean - bit weird that `id` now always needs to be given a prefix and can no longer be a getter, but it makes sense as to why.

src/components/InngestFunction.ts
```diff
@@ -68,16 +68,17 @@ export class InngestFunction<Events extends Record<string, EventPayload>> {
      * This function can't be expected to know how it will be accessed, so
      * relies on an outside method providing context.
      */
-    baseUrl: URL
+    baseUrl: URL,
+    appPrefix?: string
   ): FunctionConfig {
     return {
-      id: this.id,
+      id: this.id(appPrefix),
```
jpwilliams, line 75: Nit: Can we do this once to avoid an accidental mismatch of IDs? I made this mistake a while ago when I changed the top `id` but not the ID in the step URL 😆 

```ts
const id = this.id(appPrefix);

return {
  id,
  ...
```

src/handlers/default.ts
```diff
@@ -124,15 +124,15 @@ export class InngestCommHandler {
 
     this.fns = functions.reduce<Record<string, InngestFunction<any>>>(
       (acc, fn) => {
-        if (acc[fn.id]) {
+        if (acc[fn.id(this.name)]) {
```
jpwilliams, line 127: Nit: As one of the comments above, `const id = fn.id(this.name)`, then use it to avoid accidentally only changing one.

jpwilliams (approved):
Ace shout for avoiding conflicts - a requirement really!

Comments are just nits.

src/components/InngestFunction.ts
tonyhb: Aha, this is taken care of in your slugify regex which removes leading and trailing slashes.  I'll commit this, though, as it's safer and nicer!

PR #17, opened 2022-09-29
Add dev server integration

tonyhb:
This PR introduces integrations with the `inngest-cli` dev server.  It attempts to check if the dev server is running, and will ultimately swap out:

- The registration API for the dev server's registration URL
- The event API for the dev server's API

Based on whether the dev server is available.

Note: the dev server is never available in production.

### DIFF
--- a/src/express.ts
+++ b/src/express.ts
@@ -4,9 +4,9 @@ import { z } from "zod";
 import { Inngest } from "./components/Inngest";
 import { InngestFunction } from "./components/InngestFunction";
 import { envKeys, queryKeys } from "./helpers/consts";
+import { devServerAvailable, devServerUrl } from "./helpers/devserver";
 import { strBoolean } from "./helpers/scalar";
 import { landing } from "./landing";
-import { available, devserverURL } from "./helpers/devserver";
 import type {
   FunctionConfig,
   IntrospectRequest,
@@ -108,6 +108,14 @@ export class InngestCommHandler {
 
   protected readonly frameworkName: string = "default";
   protected signingKey: string | undefined;
+
+  /**
+   * A property that can be set to indicate whether or not we believe we are in
+   * production mode.
+   *
+   * Should be set every time a request is received.
+   */
+  protected _isProd = false;
   private readonly headers: Record<string, string>;
   private readonly fetch: FetchT;
 
@@ -201,6 +209,8 @@ export class InngestCommHandler {
           .set("x-inngest-sdk", `js/${this.frameworkName}`)
           .json({ message });
       }
+
+      this._isProd = process.env.ENVIRONMENT === "production";
 
       switch (req.method) {
         case "GET": {
@@ -325,8 +335,14 @@ export class InngestCommHandler {
     // Whenever we register, we check to see if the dev server is up.  This
     // is a noop and returns false in production.
     let registerURL = this.inngestRegisterUrl;
-    if (await available()) {
-      registerURL = devserverURL("/fn/register");
+    const devServerHost = process.env[envKeys.DevServerUrl];
+
+    if (!this.isProd) {
+      const hasDevServer = await devServerAvailable(devServerHost, this.fetch);
+
+      if (hasDevServer) {
+        registerURL = devServerUrl(devServerHost, "/fn/register");
+      }
     }
 
     try {
@@ -368,6 +384,10 @@ export class InngestCommHandler {
     );
 
     return { status, message: error };
+  }
+
+  private get isProd() {
+    return this._isProd;
   }
 
   protected shouldShowLandingPage(strEnvVar: string | undefined): boolean {
--- a/src/express.ts
+++ b/src/express.ts
@@ -168,7 +168,7 @@ export class InngestCommHandler {
 
     this.headers = {
       "Content-Type": "application/json",
-      "User-Agent": `InngestJS v${version} (${this.frameworkName})`,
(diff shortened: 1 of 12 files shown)


src/helpers/devserver.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Normalize the URL to be friendly here.  If the user hasn't added a scheme,
+  // default to http.
+  if (host.indexOf("://") === -1) {
+    host = `http://${host}`;
+  }
+
+  const url = new URL(host);
+  url.pathname = pathname || "";
+  return url;
+};
```
jpwilliams: Nitpicking - can this be more succinct? I think this is the same functionality

```suggestion
export const devserverURL = (pathname = ""): URL => {
  let host =
    process.env.INNGEST_DEVSERVER_URL?.trim() || "http://127.0.0.1:8288/";
  if (!host.includes("://")) {
    host = `http://${host}`;
  }

  return new URL(pathname, host);
};
```

src/express.ts
```diff
@@ -195,7 +197,7 @@ export class InngestCommHandler {
           "Unable to determine your site URL to serve the Inngest handler.";
         console.error(message);
 
-        return res.status(500).json({ message });
+        return res.status(500).set("x-inngest-sdk", `js/${this.frameworkName}`).json({ message });
```
jpwilliams: Can we move setting these headers to be the same as the `User-Agent` header? See [src/express.ts#299](https://github.com/inngest/inngest-js/blob/521a4d296758f5ea5560ee0c7679fe5c101ac3e0/src/express.ts#L299)

We could abstract creating that header, perhaps, and use it like `this.sdkHeader()`?

src/express.ts
```diff
@@ -306,8 +308,15 @@ export class InngestCommHandler {
 
     let res: globalThis.Response;
 
+    // Whenever we register, we check to see if the dev server is up.  This
+    // is a noop and returns false in production.
+    let registerURL = this.inngestRegisterUrl;
+    if (await available()) {
```
jpwilliams, line 328: Nit - I find `await available()` an odd statement to read - could we rename the `available` import to `isDevServerAvailable()` or something more specific?

src/express.ts
```diff
@@ -195,7 +196,10 @@ export class InngestCommHandler {
           "Unable to determine your site URL to serve the Inngest handler.";
         console.error(message);
 
-        return res.status(500).json({ message });
+        return res
+          .status(500)
+          .set("x-inngest-sdk", `js/${this.frameworkName}`)
```
jpwilliams, line 201: Can we refactor the `js/${this.frameworkName}` to use the same as the `User-Agent` header? See [src/express.ts#299](https://github.com/inngest/inngest-js/blob/521a4d296758f5ea5560ee0c7679fe5c101ac3e0/src/express.ts#L299)

Abstracting that and this out to `this.sdkHeader()` might be nice?

src/helpers/devserver.ts
```diff
@@ -0,0 +1,87 @@
+import fetch from "cross-fetch";
```
jpwilliams, line 1: I'll test this with Cloudflare - I think we'll have to be able to pass the desired `fetch` implementation in.

src/helpers/devserver.ts
```diff
@@ -0,0 +1,87 @@
...
+    return false;
+  }
+};
+
+// isProd compares any supported standard env variable for "production",
+// returning true on first match.
+export const isProd = (): boolean => {
+  return !!Object.values(prodEnvKeys).find(
+    (e) => process.env[e] === "production"
+  );
```
jpwilliams, line 25: Not all environments have `process.env`. 😢 

Maybe each specific framework should have a method for defining whether it's "in prod" or not and we use that before we get here?

src/helpers/devserver.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+// url returns the dev server URL, overriding to use the INNGEST_DEVSERVER_URL
+// env var if provided.
+export const devserverURL = (pathname?: string): URL => {
+  let host = process.env[envKeys.DevServerURL];
```
jpwilliams, line 31: Another `process.env` to remove. Maybe we allow `host` to be passed in instead? 🤔

src/helpers/devserver.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Normalize the URL to be friendly here.  If the user hasn't added a scheme,
+  // default to http.
+  if (host.indexOf("://") === -1) {
+    host = `http://${host}`;
+  }
+
+  const url = new URL(host);
+  url.pathname = pathname || "";
+  return url;
+};
```
jpwilliams, line 48: Pending `process.env` change, could this be more succinct?

```suggestion
export const devServerUrl = (pathname = ""): URL => {
  let host =
    process.env.INNGEST_DEVSERVER_URL?.trim() || "http://127.0.0.1:8288/";
  if (!host.includes("://")) {
    host = `http://${host}`;
  }

  return new URL(pathname, host);
};
```

src/helpers/devserver.ts
```diff
@@ -0,0 +1,87 @@
+import fetch from "cross-fetch";
+import { envKeys, prodEnvKeys } from "./consts";
+import type { FunctionConfig } from "../types";
```
jpwilliams, line 3: `import type` 😍

jpwilliams (approved):

PR #26, opened 2022-10-19
Automatically use dev server if available when sending events

tonyhb:
This PR changes the event URL to use the dev server, if the dev server is locally available.

Because events may be sent from the front and backend, we introduce a new `env` file _specifically_ for manaing process.env, which searches known variables including frontend prefixes transformed via webpack's basic DefinePlugin

### DIFF
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -91,7 +91,7 @@ export class Inngest<Events extends Record<string, EventPayload>> {
    */
   constructor({
     name,
-    eventKey = process.env[envKeys.EventKey],
+    eventKey,
     inngestBaseUrl = "https://inn.gs/",
   }: ClientOptions) {
     if (!name) {
@@ -105,7 +105,13 @@ export class Inngest<Events extends Record<string, EventPayload>> {
     }
 
     this.name = name;
-    this.eventKey = eventKey;
+
+    this.eventKey =
+      eventKey ||
+      (typeof process === "undefined"
+        ? ""
+        : process.env[envKeys.EventKey] || "");
+
     this.inngestBaseUrl = new URL(inngestBaseUrl);
     this.inngestApiUrl = new URL(`e/${this.eventKey}`, this.inngestBaseUrl);
 
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -268,7 +268,10 @@ export class Inngest<Events extends Record<string, EventPayload>> {
 
     if (!isProd()) {
       const host = devServerHost();
-      if (await devServerAvailable(host, fetch)) {
+      // If the dev server host env var has been set we always want to use
+      // the dev server - even if it's down.  Otherwise, optimistically use
+      // it for non-prod services.
+      if (host !== undefined || await devServerAvailable(host, fetch)) {
         url = devServerUrl(host, `e/${this.eventKey}`).href;
       }
     }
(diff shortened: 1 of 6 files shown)


src/components/Inngest.ts
```diff
@@ -89,7 +91,7 @@ export class Inngest<Events extends Record<string, EventPayload>> {
    */
   constructor({
     name,
-    eventKey = process.env[envKeys.EventKey],
```
jpwilliams, line 92: This was a bug unrelated to this PR, but needed fixing to test! We can't use `process.env` like this, as we're in shared code here.

src/components/Inngest.ts
```diff
@@ -254,7 +262,18 @@ export class Inngest<Events extends Record<string, EventPayload>> {
       );
     }
 
-    const response = await fetch(this.inngestApiUrl.href, {
+    // When sending events, check if the dev server is available.  If so, use the
+    // dev server.
+    let url = this.inngestApiUrl.href;
+
+    if (!isProd()) {
```
jpwilliams, line 269: An interesting point to note here is that we have two different checks for prod - this one and the one within handlers, such as `"inngest/next"`.

It makes sense as we can't reliably detect what environment we're in here, whereas handlers are explicitly imported, so we can assume the user has got it right.

Don't think there's an action here, but raising in case we have a bright idea to reduce any drift.

src/helpers/devserver.ts
```diff
@@ -11,13 +11,15 @@ type FetchT = typeof fetch;
 /**
  * Attempts to contact the dev server, returning a boolean indicating whether or
  * not it was successful.
+ *
+ * @example devServerUrl(process.env[envKeys.DevServerUrl], "/your-path")
```
jpwilliams, line 15: SpongebobWorshipGIF

src/helpers/env.ts
```diff
@@ -0,0 +1,44 @@
...
+  // text replacement instead of actually understanding the AST, despite webpack
+  // being fully capable of understanding the AST.
+  const values =
+    typeof process === "undefined"
+      ? []
+      : [
+          process.env.INNGEST_DEVSERVER_URL,
+          process.env.REACT_APP_INNGEST_DEVSERVER_URL,
+          process.env.NEXT_PUBLIC_INNGEST_DEVSERVER_URL,
+        ];
```
jpwilliams, line 32: I adjusted both functions to handle `process` being undefined as in non-Node environments. Right now, this means it'll return the `defaultDevServerHost` here and `false` for `isProd()`.

The only place this will affect atm is `"inngest/cloudflare"`. There are some complications here, but I think this is still fine to ship for now as that handler is still relatively alpha.

jpwilliams (approved):

jpwilliams:
@tonyhb I moved some code around a little to fix this for the Cloudflare (non-Node) environment. Looking good on Cloudflare/Next and Vercel/Next for me.

src/helpers/env.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return values.find((a) => !!a) || defaultDevServerHost;
+};
+
+export const isProd = (): boolean => {
```
djfarrelly, line 37: Is there a reason that we couldn't re-use this on the server side?

djfarrelly (changes requested):
I left a suggestion for a change that I think we should make - if the user explicitly sets the environment variable, we use that, even if the dev server is not available. I think we should only automatically use the dev server as an enhancement if the user hasn't set the variable.

src/helpers/env.ts
tonyhb: This code isn't necessarily server _or_ client side — it handles both.

src/components/Inngest.ts
tonyhb: Yea, this is a good idea.  We definitely want to stop prod receiving it in that case.  I'll make that change!

PR #27, opened 2022-10-19
Fix function registered log spam

tonyhb:
This fixes function registered log spam by hashing the contents of the registration request, then storing the hash of the most recently successful registration request.  If the hashes are the same consecutively, we don't log a success message.

This also allows us to track hashes of each deploy from the devserver to prevent log spam there :)

### DIFF
--- a/src/express.ts
+++ b/src/express.ts
@@ -17,16 +17,11 @@ import type {
 import { version } from "./version";
 
 /**
- * lastRegisteredHash stores a checksum of the last successful function hash
- * that was registered via this instance of the server.
- */
-let lastRegisteredHash: string | undefined = undefined;
-
-/**
  * A schema for the response from Inngest when registering.
  */
 const registerResSchema = z.object({
   status: z.number().default(200),
+  skipped: z.boolean().optional().default(false),
   error: z.string().default("Successfully registered"),
 });
 
@@ -396,15 +391,14 @@ export class InngestCommHandler {
     } catch (err) {
       console.warn("Couldn't unpack register response:", err);
     }
-    const { status, error } = registerResSchema.parse(data);
+    const { status, error, skipped } = registerResSchema.parse(data);
 
     // The dev server polls this endpoint to register functions every few
     // seconds, but we only want to log that we've registered functions if
     // the function definitions change.  Therefore, we compare the body sent
     // during registration with the body of the current functions and refuse
     // to register if the functions are the same.
-    if (lastRegisteredHash !== body.hash) {
-      lastRegisteredHash = body.hash;
+    if (!skipped) {
       console.log(
         "registered inngest functions:",
         res.status,
--- a/src/express.ts
+++ b/src/express.ts
@@ -17,6 +17,12 @@ import type {
 import { version } from "./version";
 
 /**
+ * lastRegisteredHash stores a checksum of the last successful function hash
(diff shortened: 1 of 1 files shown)


src/express.ts
```diff
@@ -335,6 +341,10 @@ export class InngestCommHandler {
       sdk: this.sdkHeader[1],
       v: "0.1",
     };
+
+    // Calculate the checksum of the body... without the checksum itself being included.
+    body.hash = shajs("sha256").update(JSON.stringify(body)).digest("hex");
```
djfarrelly, line 346: It might be a good idea to do a checksum of `functions` instead of the entire `body`. We talked in the past about adding a checksum in the body for registering ([code](https://github.com/inngest/inngest-js/blob/81a89539f0c13ebd5667be75f14f62a216eaa3b3/src/express.ts#L344-L362)) so the Inngest API could determine if any functions have changed (Future "deployments" feature). What do you think is better? checksum of just the functions (or the other fields which likely won't change.

djfarrelly (commented):
Quick work and great idea! I only have a question of if we should be using something else to generate the checksum.

src/express.ts
tonyhb: To be honest, I think it's safer to re-deploy if any of the JSON changes (barring hash), such as the SDK versions, deploy type, etc.  Though some of these attributes are in the function config not all of them are.  This means that we can properly show deploy information to users as they upgrade (or don't upgrade) the SDK — which will be important with breaking changes in the future.

src/express.ts
```diff
@@ -16,6 +16,12 @@ import type {
 } from "./types";
 import { version } from "./version";
 
+/**
+ * lastRegisteredHash stores a checksum of the last successful function hash
+ * that was registered via this instance of the server.
+ */
+let lastRegisteredHash: string | undefined = undefined;
```
jpwilliams, line 23: Placing this here assumes we'll only ever have a single serve handler whereas it's technically possible to have multiple; it might be better to make this a property of the class instead.

src/express.ts
```diff
@@ -387,12 +397,21 @@ export class InngestCommHandler {
...
-      res.statusText,
-      data
-    );
+
+    // The dev server polls this endpoint to register functions every few
+    // seconds, but we only want to log that we've registered functions if
+    // the function definitions change.  Therefore, we compare the body sent
+    // during registration with the body of the current functions and refuse
+    // to register if the functions are the same.
+    if (lastRegisteredHash !== body.hash) {
```
jpwilliams, line 406: I'm not sure this would work in a serverless environment where this file will be imported and run on every request. I'm not sure it'd be a problem for Express if it's a running process, but the other handlers would have a more difficult time.

Instead of comparing the hash here, can we have the dev server do the work and return a particular piece of data stating that no change has been made? It has a much better chance of being stateful than this code.

djfarrelly: You're right about serverless or a server that is restarted often during dev (think: [nodemon](https://nodemon.io/)). I think with Tony's dev server change (https://github.com/inngest/inngest/pull/263) it does do that as well so it is kind of a progressive enhancement on both sides of the system. When we add deployments to Inngest Cloud, we can emulate what the dev server is doing as well.

tonyhb: We'll handle 202s && "skipped: true" as if the functions were already registered.

PR #35, opened 2022-10-28
Add signature validation to incoming requests

tonyhb:
This ensures that incoming events are valid according to the PSK.

### DIFF
--- a/src/cloudflare.ts
+++ b/src/cloudflare.ts
@@ -92,8 +92,11 @@ class CloudflareCommHandler extends InngestCommHandler {
           const input = await req.json();
 
           try {
-            this.validateSignature(req.headers.get("x-inngest-signature"), input)
-          } catch(e) {
+            this.validateSignature(
+              req.headers.get("x-inngest-signature"),
+              input
+            );
+          } catch (e) {
             console.warn("Invalid x-inngest-signature", e);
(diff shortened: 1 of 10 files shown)


src/cloudflare.ts
```diff
@@ -89,6 +89,22 @@ class CloudflareCommHandler extends InngestCommHandler {
...
           // Inngest is trying to run a step; confirm signed and run.
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+          const input = await req.json();
+
+          try {
+            this.validateSignature(
+              req.headers.get("x-inngest-signature"),
+              input
+            );
+          } catch (e) {
```
jpwilliams, line 100: For these handlers, do we want to validate the signature in dev too? Ideally, we can trigger functions via the dev server which may not have access to the signing keys.

We could provide an env var like `INNGEST_BYPASS_SIGNATURE=1` that will be picked up here to avoid validating, but then we'd have to always add that in dev which is a pain.

Perhaps we use the `isProd()` check inside `validateSignature()` and only actually validate in production?

tonyhb: Completely agree;  this should be prod only and we should warn if not necessary.

jpwilliams:
Closing in favour of #86.

PR #42, opened 2022-11-08
Use op hash as ID

tonyhb:
Tools use the op hash to load function state, so we must use the op hash as the op ID so that the executor can use this identifier when storing state.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -62,10 +62,9 @@ export type Op = {
   op: StepOpCode;
 
   /**
-   * An identifier for this operation, used to confirm that the operation was
-   * completed when it is received from Inngest.
+   * The unhashed step name for this operation.
    */
-  id: string;
+  name: string;
 
   /**
    * Any additional data required for this operation to send to Inngest. This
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -87,16 +86,18 @@ export type Op = {
...
 };
 
 /**
  * A helper type to represent a stack of operations that will accumulate
- * throughout a step function's run.
+ * throughout a step function's run.  This stack contains an object of
+ * op hashes to data.
  */
-export type OpStack = Record<string, Op>;
+export type OpStack = Record<string, any>;
```
jpwilliams, line 100: Why did this change to `any`?

jpwilliams (approved):
Lovely test changes! <3

src/types.ts
tonyhb: The state store only records the data within an Op, instead of the Op itself.  The rest of the data returned in the Op is used to construct an edge for our queue (eg. sleep in 10 minutes).

We don't need to return the full Op information to the SDK — only the data.  In the future, we can add Op information to the query params (?step=Step&op=$data) to indicate which parallel step to run.  This feels natural, as this is how we configure which step to run already :)

jpwilliams: @tonyhb Aha of course. Cool and big agree with putting the op to run in the query string! Awesome!

PR #44, opened 2022-11-10
Remove ms from date strings, set MS to 0 when comparing dates

tonyhb:


djfarrelly (approved):
I think this could have worked just with removing the millisecond period, but setting the milliseconds ensures that the math that we do is only to the nearest second, which is the fidelity that we provide, so I think it's fine to keep that in there!

PR #46, opened 2022-11-16
When manually triggering a cron, send expected body and headers

jpwilliams:
The manual cron trigger in the SDK UI needs to partially match the expected input as it would be from Inngest.

There's the option to handle blank requests within the lib, though it's probably healthier to keep ensuring incoming requests are good.


djfarrelly (commented):
Good fix!

PR #50, opened 2022-12-01
Add ability for `inngest/remix` import to handle SDK UI

jpwilliams:
## Summary

Remix specifies [Resource Routes](https://remix.run/docs/en/v1/guides/resource-routes), which can be used as generic API routes.

This handler used to work by exporting a single `action` handler, but since the SDK UI was added, we now also need a `loader` handler.

This PR ensures that both are exported.

```ts
import { serve } from "inngest/remix";
import fns from "~/inngest";

export const { loader, action } = serve("My Remix App", fns);
```


tonyhb (approved):

djfarrelly (commented):
LGTM

PR #56, opened 2022-12-15
fix: `require` not defined on landing page

callingmedic911:
**Problem**:
When trying to send event via landing page (For example: http://localhost:8911/inngest), it throws error that `require` is not defined.
image

**Fix:**
Avoid dynamic `require` import. We can use ES6 import but we are sharing Inngest.ts for the npm module (commonjs) and landing page, so thought this might be the safest way to go about it.

Let me know if you think of any changes. Also, feel free to discard this PR if not needed.

cc: @KrisCoulson

### DIFF
--- a/landing/src/components/ExpandableEventSender.tsx
+++ b/landing/src/components/ExpandableEventSender.tsx
@@ -49,7 +49,12 @@ const ExpandableEventSenderUI = ({
       return;
     }
 
-    const inngest = new Inngest({ name: value.appName, inngestBaseUrl: value.devServerURL, eventKey: "dev-server" });
+    const inngest = new Inngest({
+      name: value.appName,
+      inngestBaseUrl: value.devServerURL,,
+      eventKey: "dev-server",
+      fetch: fetch.bind(window),
+    });
     await inngest.send(JSON.parse(data));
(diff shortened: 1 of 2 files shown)


jpwilliams:
@callingmedic911 Thank you for this! 🙌

Frustratingly, `cross-fetch` doesn't have a good time in the Cloudflare Workers environment (see lquixada/cross-fetch#138), hence having to perform the import only if another alternative hasn't been given.

Truthfully we want this event-sending UI to disappear here and push you to the dev server UI (http://localhost:8228 by default) - this input requires the dev server to be up anyway, so we might as well move users to the better experience!

In the meantime, I'll shift this to pass the global fetch implementation to the client, and we can look at a better experience with #59.

jpwilliams (approved):

jpwilliams:
Integration tests run at https://github.com/inngest/inngest-js/actions/runs/3713793457.

Will need to create some fine-grained access tokens for public example repo cloning and either hard-code or use `pull_request_target` triggers for the actions to allow secret access.

jpwilliams:
Released in [v0.8.9](https://github.com/inngest/inngest-js/releases/tag/v0.8.9) 🙂

PR #61, opened 2022-12-22
Add `"inngest/edge"` serve handler for use in v8 edge runtimes

jpwilliams:


tonyhb (approved):
Yeee!  This is great :)

PR #66, opened 2023-01-06
INN-998 Add local Zod schemas when instantiating clients

jpwilliams:
## Summary - INN-998

Same-app types should be easily specified within the codebase, without having to wait for data to flow to production and then generating types.

Also, nobody likes code generation, so if we can avoid requiring that, then everyone's happier.

This PR introduces the ability to utilise [colinhacks/zod](https://github.com/colinhacks/zod) to specify typing.

```ts
const inngest = new Inngest({
  name: "My Client",
  schemas: {
    "app/user.created": {
      data: z.object({
        id: z.number(),
      }),
    },
  },
});
```

If you want to specify types using just TS types instead, a function is provided to let you do that.

```ts
const inngest= new Inngest({
  name: "My Client",
  schemas: schemaFromTypes<{
    "app/user.created": {
      data: {
        id: number;
      };
    };
  }>(),
});
```

I'd love to provide this as a top-level generic, so we can do `new Inngest<TYPES>()` as we can now, but this locks us in a nasty place. We use a lot of inference across the client, but TS still doesn't support inferring only _some_ generics - see microsoft/TypeScript#26242; if we enabled this, then we'd no longer be able to infer functionality or tooling from options choices smartly in the future.

An alternative we could explore could be to create a `Schemas` class, where the API would allow us to build and merge our types from varying sources, e.g.:

```ts
const inngest = new Inngest({
  name: "My Client",
  schemas: new Schemas()
    .fromGenerated<GeneratedEvents>()
    .fromTypes<{
      "app/user.created": {
        data: { id: number };
      };
    }>()
    .fromZod({
      "app/user.deleted": {
        data: z.object({ id: z.number() }),
      },
    }),
});
```

This feels a touch messier to get started but provides a lot more power in the long run, as we could merge generated, TS, and Zod typing as one, choosing overwriting order. In addition, discoverability is great here, allowing users to see the `fromZod()` function during creation to understand they can use Zod types without having to read comments or docs.

## Related

- Requires inngest/inngest-js#45
- Closes inngest/inngest-js#139

### DIFF
--- a/.changeset/gentle-oranges-grow.md
+++ b/.changeset/gentle-oranges-grow.md
@@ -1,5 +1,0 @@
----
-"inngest": patch
----
-
-INN-1253 Show actionable error when steps are nested
--- a/.changeset/gentle-oranges-grow.md
+++ b/.changeset/gentle-oranges-grow.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+INN-1253 Show actionable error when steps are nested
--- a/.changeset/gentle-swans-wonder.md
+++ b/.changeset/gentle-swans-wonder.md
@@ -1,5 +1,0 @@
----
-"inngest": minor
----
-
-Add `onFailure` handler to `createFunction` options, allowing you to specify a new function to run when the initial handler fails
--- a/.changeset/gentle-swans-wonder.md
+++ b/.changeset/gentle-swans-wonder.md
@@ -1,0 +1,5 @@
+---
+"inngest": minor
+---
+
+Add `onFailure` handler to `createFunction` options, allowing you to specify a new function to run when the initial handler fails
--- a/.changeset/heavy-oranges-breathe.md
+++ b/.changeset/heavy-oranges-breathe.md
@@ -1,5 +1,0 @@
----
-"inngest": patch
----
-
-Refactor `InngestCommHandler` to better detect env and reduce duplication (INN-997)
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,56 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
+    "ghcr.io/enricosecondulfo/devcontainer-features/volta:1": {}
+  },
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "deno.enablePaths": ["./deno_compat"],
+        "typescript.tsdk": "node_modules/typescript/lib",
+        "eslint.validate": [
+          "javascript",
+          "javascriptreact",
+          "typescript",
+          "typescriptreact",
+          "graphql"
+        ],
+        "editor.rulers": [80],
+        "typescript.format.enable": false,
+        "eslint.format.enable": true,
+        "editor.codeActionsOnSave": {
+          "source.organizeImports": true,
+          "source.fixAll": true
+        },
+        "search.exclude": {
+          "**/dist": true,
+          "**/etc": true
+        },
+        "[javascript]": {
+          "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        },
+        "[javascriptreact]": {
+          "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        },
+        "[typescript]": {
+          "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        },
+        "[typescriptreact]": {
+          "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+        },
+        "[yaml]": {
+          "editor.defaultFormatter": "redhat.vscode-yaml"
+        },
+        "files.eol": "\n"
+      },
+      "extensions": [
+        "dbaeumer.vscode-eslint",
+        "kumar-harsh.graphql-for-vscode",
+        "wix.vscode-import-cost",
+        "esbenp.prettier-vscode",
+        "OfHumanBondage.react-proptypes-intellisense",
+        "redhat.vscode-yaml"
+      ]
+    }
+  }
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,7 @@ module.exports = {
     tsconfigRootDir: __dirname,
     project: ["./tsconfig.json"],
   },
-  plugins: ["@typescript-eslint", "@inngest"],
+  plugins: ["@typescript-eslint", "@inngest", "import"],
   root: true,
   ignorePatterns: ["dist/", "*.d.ts", "*.js", "deno_compat/"],
   rules: {
@@ -21,6 +21,12 @@ module.exports = {
       "warn",
       { varsIgnorePattern: "^_", argsIgnorePattern: "^_" },
     ],
+    "@typescript-eslint/consistent-type-imports": [
+      "error",
+      { fixStyle: "inline-type-imports" },
+    ],
+    "import/consistent-type-specifier-style": ["error", "prefer-inline"],
+    "import/no-duplicates": ["error", { "prefer-inline": true }],
   },
   overrides: [
     {
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,16 +17,39 @@ env:
 
 jobs:
   test:
-    name: Test
+    name: Runtime
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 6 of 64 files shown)


djfarrelly:
Hey @jpwilliams - excited about first class Zod support. Here are my thoughts on this topic and the potential UX:

- What's the benefit of using chainable methods w/ `new Schemas` constructor? Which is a better UX over passing them as options/args to the Schemas constructor? Or alternatively, make `schemas` an array where you can pass any number of schemas that we combine:
```ts
const MyZodEvent = z.object({
  name: z.literal("my.zod.event"),
  data: z.object({
    z.string(),
  })
});

type APIEvents = API_ItemCreated | API_ItemDeleted;

const new Inngest({
  schemas: [ 
    MyZodEvent
    new Schema<EventA | EventB>(), 
    new Schema<APIEvents>(), 
    new Schema<GeneratedEvents>()
    new Schema()
  ]
})
```
* I'm not sure the simplest UX here, added a `Schemas` or `Schema` constructor feels like another concept that we're creating, although it could give us flexibility to accept things like `json schema` in the future as well. 
* I'm wondering about cases where some folks may separate schemas or types into objects or unions in different files and use the `Inngest` client to combine different groups of events. Mostly thinking about flexibility here.
* If we use a `Schema` constructor - is there any other future potential benefit to that object? Could we use that object to help with data governance/pushing types to Inngest Cloud? 
* If we lean into schemas as an array/object, could we create a `@inngest/schemas` package that has all of our integrations types as Zod types? e.g. 
```ts
import { stripeSchemas } from "@inngest/schemas";
const new Inngest({ schemas: [ new Schemas<MyCustomEventA>(), stripeSchemas ] });
```

- For TypeScript types (probably our default recommended), I think it's a nicer UX to pass all types as a union instead of a type with keys that are duplicates of the event type's `name`. e.g. `Constructor<{ name: "event.b", data: { ok: boolean } } | { name: "event.b", data: { id: string } }>`

I'm not set on the best UX, but thinking through some of the above might allow us to think about something that gives us future flexibility.

jpwilliams:
> What's the benefit of using chainable methods w/ `new Schemas` constructor? Which is a better UX over passing them as options/args to the Schemas constructor? Or alternatively, make `schemas` an array where you can pass any number of schemas that we combine

@djfarrelly The upside of a class with chainable methods is that it's easier to both discover and support new methods of defining schemas.

Suppose the `Generic` in `new Schema<Generic>()` can be a wide array of formats (Zod, keyed events, non-keyed events (in a union), JSON Schema, etc.). In that case, it's much harder for the user to understand how they can define their schema, and much harder to validate on our side. Type errors (e.g. for defining an event with no name) are naturally more verbose as they'll explain how the given type doesn't fit all of the cases we support.

With the chainable methods, every method is discoverable via autocomplete, is responsible only for its own simpler roll-up of the given type to an expected standard type, and errors will be much easier to read as they are solely focused on a single format.

---

> I'm not sure the simplest UX here, added a `Schemas` or `Schema` constructor feels like another concept that we're creating, although it could give us flexibility to accept things like `json schema` in the future as well.

@djfarrelly 💯 Like you, I'm not a fan of introducing more concepts where it feels like a user could just define an object, though I think the benefits are worth it. Already with just supporting Zod, TS types, and generated schemas, the `schemas` field has to accept way too many permutations and it's unclear for users what they can provide without reading docs.

Hopefully with this, we make it clear that users can define a schemas with a `new Schemas()` instance and then each chained method can have prettier comments around how to use each format correctly.

---

> I'm wondering about cases where some folks may separate schemas or types into objects or unions in different files and use the `Inngest` client to combine different groups of events. Mostly thinking about flexibility here.

@djfarrelly Agreed - it's likely some users will want their schemas held away from the client to not muddy that file. Chainable methods will support this pattern well, allowing you to build up your schemas from multiple imported sources.

---

> If we use a `Schema` constructor - is there any other future potential benefit to that object? Could we use that object to help with data governance/pushing types to Inngest Cloud?

@djfarrelly Yes! One common improvement I can see is being able to set schema-specific options up front. For example, some users might like the ability to build an untyped function before specifying their event. For this, we could provide something like `.allowUntyped()` so that Inngest will do its best to infer data from event names given, but it can also just be a string and type the event as `any`.

```ts
const schemas = new Schemas()
  .fromTypes<{
    "app/user.created": {
      data: { id: number };
    };
  }>()
  .allowUntyped();
```

Could also be an option passed to `new Schemas(opts)` or whatever we see fit. Generally though, yes for sure. Zod schemas in general also allow us to enforce data shapes on the client in the future which we could - as you say - push from these objects.

---

> If we lean into schemas as an array/object, could we create a `@inngest/schemas` package that has all of our integrations types as Zod types? e.g.
> 
> ```ts
> import { stripeSchemas } from "@inngest/schemas";
> const new Inngest({ schemas: [ new Schemas<MyCustomEventA>(), stripeSchemas ] });
> ```

@djfarrelly Absolutely! This would be a great way to grab many known types without forcing them on new users.

We could even bundle this into the SDK itself, so there's not a separate package; seeing as they're just types they'll never be shipped so resulting package size isn't an issue, though general package download size for the developer might bloat a bit.

---

> For TypeScript types (probably our default recommended), I think it's a nicer UX to pass all types as a union instead of a type with keys that are duplicates of the event type's `name`. e.g. `Constructor<{ name: "event.b", data: { ok: boolean } } | { name: "event.b", data: { id: string } }>`

@djfarrelly Yep, we can support this pattern! My only reservation is that you can accidentally provide events with the same name which might change some of the internal typing.

When a user does this and then declares `inngest.createFunction("event.b", ...)`, should the library provide a union of the data as the possible event? e.g. in the case of your example above, `event.data` for me would have the type:

```ts
{ ok: boolean } | { id: string }
```

I guess in these cases we'd also want to enforce a `v`ersion that the user could then use to ensure TypeScript knows which data type is appropriate, e.g. they could do:

```ts
if (event.v === "2") {
  // data is typed correctly for this particular event
} else {
  // only one other option, so typed correctly again
}
```

DaleLJefferson:
This would be really good, the automatic schema generation makes a good demo but I really wants things static and defined in code. 

I'm currently using zod for this which works today.

```ts
const data = z.object({
    a: z.string()
})

type Events = {
    "event1": {
        name: "event1";
        data:  z.infer<typeof data>;
    };
};
```

Then I parse the event into Zod in the function, having this built in sounds great.

PS having to write out the event name twice is frustrating.

jpwilliams:
In addition to the above requirements, supporting wildcards is a big win for many systems, e.g. listening to `pipeline/*`, `pipeline/sync.*` or `pipeline/*.errored`.

Given the stacking set of schemas using `new Schemas()`, we might be able to create smart union types from the input events, splitting on our recommended separators of `/` and `.`.

For example, the following stacked schema:
```ts
const inngest = new Inngest({
  name: "My Client",
  schemas: new Schemas()
    .fromTypes<{
      "app/user.created": {
        data: { id: number };
      };
    }>()
    .fromZod({
      "app/user.deleted": {
        data: z.object({ id: z.number() }),
      },
      "app/post.created": {
        data: z.object({ id: z.number() }),
      },
    }),
});
```
Would produce the following types for listening:
```ts
type Events = {
  "app/user.created": AppUserCreated;
  "app/user.deleted": AppUserDeleted;
  "app/post.created": AppPostCreated;
  "app/*": AppUserCreated | AppUserDeleted | AppPostCreated;
  "app/user.*": AppUserCreated | AppUserDeleted;
  "app/post.*": AppPostCreated;
  "app/*.created": AppUserCreated | AppPostCreated;
  "app/*.deleted": AppUserDeleted;
};
```
We might need a `**` for multiple items, too. For example, if I have `app/foo.errored` and `app/foo.bar.errored`, I might want to listen for `app/**.foo` which would capture both, whereas `app/*.errored` would only catch the former. 🤔

jpwilliams:
Will also consider [ArkType](https://github.com/arktypeio/arktype) for a schema candidate. See arktypeio/arktype#687.

tonyhb (approved):
This looks good!  Excited about this change — it opens the door for a lot!

mmachatschek:
@tonyhb @jpwilliams whats the best way to test this out? do you plan to release rc/alpha versions of the sdk so we can get our hands dirty testing out this?

mmachatschek:
@jpwilliams thanks for the bump. this comes right when I needed it 👍 will keep you posted if anything comes up while testing it out

mmachatschek:
@jpwilliams this is working great so far 👍 no issues. currently using the `fromUnion` method.

PR #68, opened 2023-01-11
Allow `sleepUntil` to parse strings

tonyhb:
`tools.run` always JSON-serializes output, which means that if a user returns a Date() then passes this into `tools.sleepUntil`, parsing previously fails.

This brings up another issue:  `tools.run` typing auto-genned is incorrect;  the resulting type is always serialized instead of being an instance of whatever object you return.

For example, a `Set` is always an array, a `Date` is always a string, a `Map` or `WeakMap`, is an object, classes are... tricky :)

### DIFF
--- a/src/components/InngestStepTools.test.ts
+++ b/src/components/InngestStepTools.test.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import ms from "ms";
 import { StepOpCode } from "../types";
 import { createStepTools, StepFlowInterrupt } from "./InngestStepTools";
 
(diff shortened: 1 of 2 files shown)


jpwilliams (approved):

jpwilliams:
Addressed the other issue in #70.

We can look at using extended JSON libraries too.

src/components/InngestStepTools.ts
```diff
@@ -369,16 +369,18 @@ export const createStepTools = <
         /**
          * The date to wait until before continuing.
          */
-        time: Date
+        time: Date | string
       ) => void
     >((time) => {
+      const date: Date = typeof time === "string" ? new Date(time) : time;
```
jpwilliams, line 375: I think `new Date("some-bad-string")` will not throw and instead return a crappy "Invalid Date" object.

You can check for this with `date.toString() === "Invalid Date"` and I guess we can throw here?

src/components/InngestStepTools.test.ts
```diff
@@ -173,13 +174,21 @@ describe("sleepUntil", () => {
...
     expect(() => sleepUntil(future)).toThrow(StepFlowInterrupt);
     await expect(state.nextOp).resolves.toMatchObject({
       op: StepOpCode.Sleep,
     });
   });
 
+  test("parses ISO strings", async () => {
+    const next = new Date().valueOf() + ms("6d");
+
+    expect(() => sleepUntil(new Date(next))).toThrow(StepFlowInterrupt);
```
jpwilliams, line 186: I think we'd need to do `new Date().toISOString()` to test this there, as currently this is passing a date?

jpwilliams (changes requested):

jpwilliams (approved):

PR #71, opened 2023-01-13
Add DigitalOcean handler

tonyhb:
This still needs tests, but allows users to deploy Inngest functions to DigitalOcean.

Example usage:

```typescript
const { serve } = require("inngest/digitalocean");

const fn = createStepFunction("User signup flow", "user/created", function({ event, tools }) {
  tools.run("Send a welcome email", () => {
    // Your email logic...
  });

  tools.sleep("24h");

  tools.run("Send reminder", () => {
   // Another email...
  });

  return {
    ok: true,
    event,
  };
});

// Call serve with your app name (for grouping functions), a list of
// functions which you want to execute, and the serveHost and Path.
// The URL to the function is needed for Inngest to register your
// functions and run them;  without this Inngest doesn't know where
// your functions are hosted.
const run = serve(
  "App name:  Signup flows",
  [fn],
  {
    // This is your digitalocean host
    serveHost: "https://faas-sfo3-your-url.doserverless.co"
    // And your DO path.
    servePath: "/api/v1/web/fn-your-uuid/inngest"
    // Add your signing key to your function env.
    signingKey: process.env["INNGEST_SIGNING_KEY"];
  },
);

async function main(args) {
  return await run(args);
}

// IMPORTANT: Makes the function available as a module in the project.
// This is required for any functions that require external dependencies.
module.exports.main = main;
```

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -750,7 +750,7 @@ type Handler = (...args: any[]) => {
  * The response from the Inngest SDK before it is transformed in to a
  * framework-compatible response by an {@link InngestCommHandler} instance.
  */
-interface ActionResponse {
+export interface ActionResponse {
   /**
    * The HTTP status code to return.
    */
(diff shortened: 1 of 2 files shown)


tonyhb:
Note:  I'm not sure how DO handles transpilation and ES6 so in tests I've used `require` instead of import.  This all works as you'd expect :)

PR #75, opened 2023-01-14
Add `tools.sendEvent()` to step function tooling

jpwilliams:
## Summary

Adds a new tool, `sendEvent()`, used to send events from a step function reliably. Without this, `inngest.send()` will repeatedly send events during step function executions, sending many more events than anticipated.

## Todo

- [ ] Executor changes to send event upon receipt with an idempotency ID of the run ID and hash of the step.

## Related

- #45


tonyhb:
This is a great idea!  I'm stealing your feedback now - should we also let users do `inngest.send()` and handle this under the hood?

Hear me out... what if we make function state accessible to the inngest client (moving away from stateless funcs -> inngest.createFunction), and then `inngest.send` auto-detected function state and ran through the same logic as `tools.send` 🤔 

I'm not sure if this is possible but it'd be good to try and remove that footgun? 🤔 

Part of me thinks that this year we might need our own linting rules to help with this stuff too haha.

jpwilliams:
> This is a great idea! I'm stealing your feedback now - should we also let users do `inngest.send()` and handle this under the hood?
> 
> Is it possible to make function state accessible to the inngest client (moving away from stateless funcs -> inngest.createFunction), and then `inngest.send` auto-detected function state and ran through the same logic as `tools.send`?
> 
> I'm not sure if this is possible but it'd be good to try and remove that footgun? 🤔

@tonyhb A great idea! Yeah, I think that - as we lean on serverless - if a function is being executed then we can safely assume that the attached client should detect this state and silently use `tools.send`.

The only danger comes if we're running serverful and the same client might be running concurrent executions; in this case we lose the ability the understand which execution we should use the tool from.

To solve that, we'd need the client to be able to understand the current context across async functions. Node’s [async context tracking (async_hooks)](https://nodejs.org/api/async_context.html) can provide some value here, though implementation differs between environments; [Deno has a polyfilled version that serves no purpose](https://deno.land/std@0.170.0/node/async_hooks.ts), and v8 worker environments (like Cloudflare) [only offer some of the features](https://github.com/cloudflare/workerd/pull/208). It's not something we can support everywhere, but <ins>**this is also not a current concern**</ins>.

---

> Part of me thinks that this year we might need our own linting rules to help with this stuff too haha.

Yes! 🙌 It can help folks a tonne with the small amount of gotchas with the tooling, e.g. affecting out-of-scope variables in `tools.run`.

Definitely want to hack on this (and an extension to bundle it) when I have time!

tonyhb:
I'm a huge fan of this and we should get this in with a lil rebase!

jpwilliams:
Closing in favour of #100.

PR #76, opened 2023-01-14
Add `Server-Timing` headers to HTTP responses

jpwilliams:
## Summary

Approximately measures performance during function execution and returns a `Server-Timing` header to the executor, e.g.:
```
Server-Timing: handler;dur=0, memoising;dur=1, running-step;dur=164
```
We can add to and tweak what we want to see here, but it should provide a sane method of approximately measuring some more SDK performance points.

## Related

- #45


src/helpers/promises.ts
```diff
@@ -37,3 +37,106 @@ export const resolveAfterPending = (): Promise<void> => {
...
 };
+
+/**
+ * A class to manage timing functions and arbitrary periods of time before
+ * generating a `Server-Timing` header for use in HTTP responses.
+ *
+ * This is a very simple implementation that does not support nested timings or
+ * fractions of a millisecond.
+ */
+export class ServerTiming {
```
djfarrelly, line 48: Could be nice as it's own file, eventually with it's own tests - feels like an standalone functionality that's not just a promise helper 🤔

src/components/InngestCommHandler.ts
```diff
@@ -412,18 +413,20 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
    */
   public createHandler(): (...args: Parameters<H>) => Promise<TransformedRes> {
     return async (...args: Parameters<H>) => {
+      const timer = new ServerTiming();
```
djfarrelly, line 416: Should this be in the `createHandler` or created within `runStep`? From what I understand, this is created when `serve()` is called, which is fine in serverless, but may have conflicts in serverful envs like Express. I may be wrong with my assumptions here.

src/helpers/promises.ts
```diff
@@ -37,3 +37,106 @@ export const resolveAfterPending = (): Promise<void> => {
...
+    string,
+    { description: string; start?: number; end?: number }
+  > = {};
+
+  /**
+   * Start a timing. Returns a function that, when called, will stop the timing
+   * and add it to the header.
+   */
+  public start(name: string, description?: string): () => void {
+    if (this.#timings[name]) {
```
djfarrelly, line 59: Is this possible that this could happen in a serverful env? (see other comment)

djfarrelly (commented):
Hey @jpwilliams - just one key question around serverful environments and multiple active timers. Let me know if I misunderstood something!

src/components/InngestCommHandler.ts
jpwilliams: It's okay that it's here, as it's inside the function returned by `createHandler()`, therefore this is only called when a request comes in, rather than when the request handler is instantiated.

src/helpers/promises.ts
jpwilliams: @djfarrelly We should be okay - each request that comes in creates its own timer (as opposed to a timer for each handler created), so `this.#timings` here will be unique.

src/helpers/promises.ts
jpwilliams: @djfarrelly Great shout!

jpwilliams:
> Hey @jpwilliams - just one key question around serverful environments and multiple active timers. Let me know if I misunderstood something!

@djfarrelly Thanks! Situation with multiple timers should be all good. Will shift to a different file. :)

tonyhb (approved):
This is great!

PR #81, opened 2023-01-16
Add an exported non-retriable error to halt SDK execution

jpwilliams:
## Summary

Adds an exported `NonRetriableError` that can be used to throw during function runs (or inside step tooling such as `tools.run()`) to cease all further execution of a function.

```ts
inngest.createStepFunction("Example", "app/user.created", () => {
  try {
    // Do something
  } catch (cause) {
    throw new NonRetriableError("Some descriptive message", { cause });
  }
});
```

Very similar implementation to #73, but submitting another PR to target the current SDK version as #73 has #45 as its base.

## Related

- #73

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -15,6 +15,7 @@ import type {
 import { version } from "../version";
 import type { Inngest } from "./Inngest";
 import type { InngestFunction } from "./InngestFunction";
+import { NonRetriableError } from "./NonRetriableError";
 
 /**
  * A handler for serving Inngest functions. This type should be used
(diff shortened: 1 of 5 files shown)


src/components/InngestCommHandler.ts
```diff
@@ -434,9 +435,9 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
 
         const stepRes = await this.runStep(runRes.fnId, "step", runRes.data);
 
-        if (stepRes.status === 500) {
+        if (stepRes.status === 500 || stepRes.status === 400) {
```
djfarrelly, line 438: Should we support `4xx` e.g. `404` may be popular ? Maybe make this `stepRes.status >= 400 && stepRes.status <= 500`?

djfarrelly (commented):
Looks pretty good @jpwilliams  - just one question on 400 vs. 4xx!

src/components/InngestCommHandler.ts
jpwilliams: This is abstracted from the user atm; they can't purposefully send any codes, so only `400` is possible in the `4xx` range atm, and any further info can be communicated via the error data passed.

tonyhb (approved):

PR #84, opened 2023-01-17
Add retry overrides to SDK options

tonyhb:
This allows users to customize the number of retries in each step.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -412,6 +412,11 @@ export interface FunctionOptions {
      */
     period: TimeStr;
   };
+
+  /**
+   * Specifies the maximum number of retries for all steps across this function.
+   */
+  retries?: number;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -412,6 +412,11 @@ export interface FunctionOptions {
      */
     period: TimeStr;
   };
+
+  /**
+   * Specifies the maximum number of retries for all steps across this function.
+   */
+  retries?: number;
```
jpwilliams, line 419: I can see someone passing in `Infinity` here to try to get it to never stop retrying.

Is a function with unlimited retries possible atm?

tonyhb: Its not, it caps at 20

jpwilliams: Added some verbose validation to this for `0` to `20`. Should warn most users early, and then the executor can complain if it gets deployed with a number outside the range.

jpwilliams (approved):

PR #86, opened 2023-01-18
Refactor handlers to add signatures

tonyhb:
This takes #35 and rebases the code to work with refactored handlers.


src/components/InngestCommHandler.ts
```diff
@@ -748,8 +753,63 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
     return this.showLandingPage ?? strBoolean(strEnvVar) ?? true;
   }
 
-  protected validateSignature(): boolean {
-    return true;
+  protected validateSignature(sig: string | undefined, body: Record<string, any>) {
+    if (!this.signingKey) {
+      console.warn(
```
djfarrelly, line 758: Reminder for me to add this to #72 when that's reviewed and merged 😁

src/components/InngestCommHandler.ts
```diff
@@ -748,8 +753,63 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
...
+    }
+
+    if (!sig) {
+      console.warn("No x-inngest-signature provided");
+      return;
+    }
+
+    // sig should have a format of `t=${unixTS}&s=${signature}` to parse.
+    //
+    // We cant use URLSearchParams here because this needs to run directly in V8
```
djfarrelly, line 775: `URLSearchParams` should be in all the major envs including v8-like envs like Deno + Cloudflare Workers (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility) and I believe it's a global in Node since v10. This solution works, but if we want to use a built-in parser we should be able to.

src/components/InngestCommHandler.ts
```diff
@@ -748,8 +753,63 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
     return this.showLandingPage ?? strBoolean(strEnvVar) ?? true;
   }
 
-  protected validateSignature(): boolean {
-    return true;
+  protected validateSignature(sig: string | undefined, body: Record<string, any>) {
+    if (!this.signingKey) {
+      console.warn(
+        "No signing key provided to validate signature.  Find your dev keys at https://app.inngest.com/test/secrets"
```
djfarrelly, line 759: Unsure about this warning being `/test`, you don't need this in development, so it can be skipped completely. I haven't used signing keys locally and it may be a noisy warning for folks getting started every time a function is called 🤔

djfarrelly (commented):
The code looks good to me, I had a question on the DX flow for local when there is no key set. It's not required for local dev, so the warnings may be a big noisy and jarring for teams.

There also has been a bit of confusion from devs wondering "Does test mode == my local dev?" and I wonder if the warning + direction might feed that confusion since the signing key is not required until the SDK is being served on the public internet.

src/components/InngestCommHandler.ts
tonyhb: IIRC it had issues with cloudflare workers, so people had workarounds: https://community.cloudflare.com/t/parse-url-query-strings-with-cloudflare-workers/90286.

src/components/InngestCommHandler.ts
tonyhb: Fixed by checking for !prod and !signingKey then returning early.  Good point!

src/components/InngestCommHandler.ts
djfarrelly: Odd, looked like from 2020, I would have hoped they fixed that by now! 😬

jpwilliams:
Pushed some changes to this:

- Added the signature verification to the handler tests, so they're tested against all of the example repos with their respective dev servers / environments too
- Refactored lots of the logic in to a `RequestSignature` class to encapsulate it a touch more (we could also move the unit tests to directly test this instead)
- A couple of fixes for detecting production across handler actions
- Try using `URLSearchParams` for parsing (see the [inngest/sdk-example-cloudflare-workers test](https://github.com/inngest/inngest-js/actions/runs/3958812287/jobs/6780827713) that tests the `"inngest/cloudflare"` handler with this)

Feel free to push this around some more!

jpwilliams (approved):

PR #91, opened 2023-01-23
Add proper error marshalling

tonyhb:
In some circumstances errors returned from the SDK were opaque (we'd return '{}').  This change ensures that we uniformly marshal errors so that users can properly debug any issues.

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -2,6 +2,7 @@ import { sha256 } from "hash.js";
 import { z } from "zod";
 import { envKeys, queryKeys } from "../helpers/consts";
 import { devServerAvailable, devServerUrl } from "../helpers/devserver";
+import { marshalError, stringifyError } from "../helpers/error";
 import { strBoolean } from "../helpers/scalar";
 import type { MaybePromise } from "../helpers/types";
 import { landing } from "../landing";
@@ -438,7 +439,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
         if (stepRes.status === 500 || stepRes.status === 400) {
           return {
             status: stepRes.status,
-            body: stepRes.error || "",
+            body: stringifyError(stepRes.error),
             headers: {
               ...headers,
               "Content-Type": "application/json",
@@ -521,7 +522,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
     } catch (err) {
       return {
         status: 500,
-        body: JSON.stringify(err),
+        body: stringifyError(err),
         headers: {
           ...headers,
           "Content-Type": "application/json",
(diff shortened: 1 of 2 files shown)


src/components/InngestCommHandler.ts
```diff
@@ -438,7 +439,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
         if (stepRes.status === 500 || stepRes.status === 400) {
           return {
             status: stepRes.status,
-            body: stepRes.error || "",
```
tonyhb, line 441: This is the offending line: `JSON.marshal(stepRes.error)` will always return `{}` due to how errors are transformed into JSON.

src/helpers/error.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (err instanceof Error) {
+    return {
+      message: err.message,
+      stack: err.stack,
+      name: err.name,
+    }
+  }
+
+  return `Unknown error: ${JSON.stringify(err)}`
```
djfarrelly: What about returning this string as the message parameter so it's more consistent in return type?

```suggestion
  return {
    message: `Unknown error: ${JSON.stringify(err)}`
  }
```

src/helpers/error.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (err instanceof Error) {
+    return {
+      message: err.message,
+      stack: err.stack,
+      name: err.name,
+    };
+  }
+
+  return `Unknown error: ${JSON.stringify(err)}`;
```
djfarrelly, line 25: What about returning this string as the message parameter so it's more consistent in return type?

```suggestion
  return {
    message: `Unknown error: ${JSON.stringify(err)}`
  }
```

tonyhb: Sounds good to me.  Was a quick code shuffle so I left what was in there :D

src/types.ts
```diff
@@ -451,7 +451,11 @@ export interface FunctionOptions {
 export type StepRunResponse =
   | {
       status: 500;
-      error?: string;
+      error?: Record<string, any> | string;
```
djfarrelly, line 454: I don't believe any step is returning a string error anymore with this code change - they should all be objects w/ the new marshal function. Perhaps we could drop the `string` here and in the `400`

src/components/InngestCommHandler.ts
```diff
@@ -438,7 +439,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
...
+            body: JSON.stringify(marshalError(stepRes.error)),
```
djfarrelly, line 442: We already marshal in `runStep`, I don't think it's needed twice 🤔

tonyhb: We get type issues otherwise.

djfarrelly (commented):
I think we should use the marshal function on line ~525 where the `handleAction`'s try/catch handles any general error that was triggered somewhere.

I left a comment on where i think we're double-marshalling on line 442 too.

src/components/InngestCommHandler.ts
djfarrelly: I would guess that if we double-marshaled like this, it would create all errors that looked like `Unknown error: ${JSON.stringify(err)}` Perhaps this is related to the union of the Record and string types in the other comment?

djfarrelly: Yeah - I was able to validate that in local testing: 
image

tonyhb: Yeah, I think this is why the previous stuff returned a string.  I wasn't super into a deep dive into refactoring code I don't own here to be honest — just to unblock this particular issue we're seeing 😬

src/components/InngestCommHandler.ts
```diff
@@ -521,7 +522,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
     } catch (err) {
       return {
         status: 500,
-        body: JSON.stringify(err),
+        body: stringifyError(err),
```
djfarrelly, line 525: 👍 this is key! If any internal error was thrown here this would have been `{}`

PR #97, opened 2023-02-10
Fix bad client path generation when transforming Deno code for testing

jpwilliams:


jpwilliams:
Closing as this will change with shifting examples housing.

PR #101, opened 2023-02-14
fix: Wait a while for nested promises to resolve

jpwilliams:
## Summary

Resolves an issue where nested promises could result in missing steps when running a step function.

- Adds a `memoizing-ticks` to the `Server-Timing` header so that we can track the performance of this over time (when benchmarking, this resolves in ~0.1ms)
- Adds an integration test that was previously breaking, a sequential reduce

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -654,7 +654,7 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
         timer
       );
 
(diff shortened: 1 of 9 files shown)


tonyhb (approved):
yer a wizard, 'arry

edit:  in actual speak, this is a great PR.  aware of the limitations and downsides to having to loop, and this is as close to a "fix" as possible given the way that promises can be resolved under the hood (from you teaching me, here).

PR #105, opened 2023-02-16
Add cancellation to function definition types

tonyhb:
This adds type support for cancellation signals, which are officially supported.

### DIFF
--- a/.changeset/slimy-drinks-beg.md
+++ b/.changeset/slimy-drinks-beg.md
@@ -1,0 +1,5 @@
+---
+"inngest": minor
+---
+
+Add `cancelOn` option when creating a function, allowing you cancel execution of a function based on incoming events.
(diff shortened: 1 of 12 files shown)


jpwilliams:
Will adjust this to add:

- [ ] `event` autocomplete based on `Inngest` client
- [ ] `timeout` to match the timeout in `step.waitForEvent()`
- [ ] `match` field autocomplete to match the field in `step.waitForEvent()`

jpwilliams:
Rebasing on #133 to reap some of the typing benefits from that change.

jpwilliams (approved):
A lot of the changes for `onFailure` and cancellation became intertwined; they both required similar refactors of typing.

Pushing this through - it has some internal changes for `onFailure` (#133) in there, too, and we can uncomment a couple of pieces to finish off that functionality.

PR #113, opened 2023-03-01
Ensure `event|cron` union is strictly adhered to when declaring a func

jpwilliams:
## Summary

#110 was shipped in `v1.3.0` but reverted in 4c1ee8d388de9f0837b11db74a74134b8a9f7e66 due to causing an event typing regression.

The fix is reintroduced here to resolve an issue where `event` and `cron` could be passed in the same object when running `inngest.createFunction()`. This includes some tests that came with the revert in 4c1ee8d388de9f0837b11db74a74134b8a9f7e66 to ensure it's working correctly.

### DIFF
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -86,3 +86,24 @@ export type ObjectPaths<T extends Record<string, any>> = Path<T>;
 export type KeysNotOfType<T, U> = {
   [P in keyof T]: T[P] extends U ? never : P;
 }[keyof T];
+
+/**
+ * Returns all keys from objects in the union `T`.
+ */
+type UnionKeys<T> = T extends T ? keyof T : never;
+
+/**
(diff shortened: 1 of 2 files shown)


src/helpers/types.ts
```diff
@@ -86,3 +86,24 @@ export type ObjectPaths<T extends Record<string, any>> = Path<T>;
 export type KeysNotOfType<T, U> = {
   [P in keyof T]: T[P] extends U ? never : P;
 }[keyof T];
+
+/**
+ * Returns all keys from objects in the union `T`.
+ */
+type UnionKeys<T> = T extends T ? keyof T : never;
```
IGassmann, line 93: **question (non-blocking)**: can this be simplified to the following? Is there a case where `T` would not extend `T`?

```suggestion
type UnionKeys<T> = keyof T;
```

PR #131, opened 2023-03-02
Add concurrency limits to function options

tonyhb:
This PR adds concurrency limits to function options, allowing users to specify the total number of concurrent steps that can run for the function.


jpwilliams:
This PR is being held while we test internally.

The functionality can be accessed via the `inngest@concurrency` distribution tag on npm.

jpwilliams (approved):

Issue #43, opened 2022-11-09
Add an `onError` handler to `run` which is called when a step permanently fails


tonyhb:
It would be good to allow users to specify cleanup code for _individual steps_ of a function, eg:

```
tools.run(
  "Something fragile",
  () => {
    // This always fails
    throw new Error("noooo!");
  },
  {
    onFail: (error: Error) => {
      // This runs after the step has been
    }
  },
});
```

jpwilliams:
#45 should address this with `.catch()`, `try`/`catch` blocks, etc.

jpwilliams:
Added in #133.

PR #133, opened 2023-03-06
Add ability to provide a function to run when another function fails

jpwilliams:
## Summary

This is the SDK side of inngest/inngest#370. From that PR:

> This [PR] outlines the ability to create functions called automatically any time functions permanently fail. Failure functions can be their own step functions and should be called with the latest error and the incoming event. This allows users to run any logic necessary for eg. cleanup, alerting, logging, etc.

For users, this ensures `inngest.createFunction()` provides some syntactical sugar over simply creating a new function that listens for the correct `inngest/function.failed` event.

```ts
inngest.createFunction(
  { name: "Send welcome email", fns: { ...emailer } },
  { event: "app/user.created" },
  async ({ event, fns: { sendEmail } }) => {
    await sendEmail({ type: "welcome", to: event.data.email });
  },
  async ({ event }) => {
    await sendEmail({
      type: "alert",
      to: "engineering@example.com",
      subject: `Welcome email failed to send to ${event.data.email}`,
    });
  }
);
```

Future per-step error handling via `.catch()`, `try`/`catch`, etc., could be preferable for some functions; the example above is short enough to be cleaner with a `try`/`catch` block. For larger functions, however, this ensures you don't have to wrap your entire function in a `try`/`catch` block or otherwise reorganise your code to be able to catch errors.

## Related

- inngest/inngest#370
- #43

## Todo

- [ ] Example and integration tests
- [ ] Documentation

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -17,6 +17,26 @@ export type EventData<Event> = Event extends never
     };
 
 /**
+ * The payload for an internal Inngest event that is sent when a function fails.
+ *
+ * @public
+ */
+export type FailureEventPayload<P extends EventPayload> = {
+  name: "inngest/function.failed";
+  data: {
+    function_id: string;
+    run_id: string;
+    error: {
+      message: string;
+      stack?: string;
+      cause?: string;
+      status?: number;
+    };
+    event: P;
+  };
+};
+
+/**
  * Arguments for a multi-step function, extending the single-step args and
  * including step function tooling.
  *
@@ -25,8 +45,9 @@ export type EventData<Event> = Event extends never
 export type HandlerArgs<
   Events extends Record<string, EventPayload>,
   Event extends keyof Events,
-  Opts extends FunctionOptions
-> = EventData<Events[Event]> & {
+  Opts extends FunctionOptions,
+  Payload = Events[Event]
+> = EventData<Payload> & {
   /**
    * @deprecated Use `step` instead.
    */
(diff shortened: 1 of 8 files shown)


djfarrelly (commented):
While I do understand the logic of following a `.then(onResolve, onReject)` pattern here vs. `onFailure`, I think this code does not read well and it's not obvious that the second anonymous async callback is an error handler. It's for this reason that I think the 

```ts
inngest.createFunction(
  { name: "Send welcome email", fns: { ...emailer } },
  { event: "app/user.created" },
  async ({ event, fns: { sendEmail } }) => {
    // success
  },
  async ({ error }) => {
    // error basically just maps to `event.data.error`
  }
);
```

This may make it more obvious. that it's an error handler, even if the user can still access the raw `event` in the error handler callback too.

Still thinking on this, but wanted to share thoughts early!

jpwilliams:
@djfarrelly Agreed. Purely with the `error` property being added, this makes it loads easier to parse.

src/types.ts
```diff
@@ -37,6 +44,18 @@ export type FailureEventPayload<P extends EventPayload> = {
...
+export type FailureEventArgs<P extends EventPayload = EventPayload> = {
+  /**
+   * The event data present in the payload.
+   */
+  event: FailureEventPayload<P>;
+
+  /**
+   * The final error that caused this function to exhaust all retries.
+   */
+  err: FailureEventPayload<P>["data"]["error"];
```
djfarrelly, line 56: Why `err` and not `error`?

jpwilliams: @djfarrelly Errors are most commonly assigned as `err` in JS which I see. Do you think it's more reasonable to use the full version?

djfarrelly: I think `error` is a bit more explicit and clear. `err` feels like a convention as does `catch(e)`, but we write our `event` and `step` arguments in full - shortening `error` to `err` feels like a weird divergence. Unfortunately, with an options arg, one cannot rename this as they want, unless we change the error handler to be `onFailure(error, { event })`.

jpwilliams: @djfarrelly Great points. Cool, let's stick with `error` then - you're right that it's more consistent.

PR #145, opened 2023-03-21
Add `"inngest/lambda"` serve handler for AWS Lambda environments

jpwilliams:
## Summary

Adds a new supported serve handler, `"inngest/lambda"` for use with AWS Lambda.

```ts
import { serve } from "inngest/lambda";

const inngest = new Inngest({ name: "My Lambda App" });

const fn = inngest.createFunction(
  { name: "Hello World" },
  { event: "test/hello.world" },
  async ({ event }) => {
    return "Hello World";
  }
);

export const handler = serve(inngest, [fn]);
```

Lambda can be triggered in many different ways from many different places. The handler currently tries to support API Gateway v1/v2 (which includes Lambda Function URLs), and we can add support for more triggers in the future (SNS, S3, etc.) as needed.

### DIFF
--- a/src/lambda.ts
+++ b/src/lambda.ts
@@ -61,9 +61,15 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
       const path = eventIsV2 ? event.requestContext.http.path : event.path;
 
       const env = allProcessEnv();
-      const proto = event.headers["x-forwarded-proto"] || "https";
 
-      const url = new URL(path, `${proto}://${event.headers.host || ""}`);
+      let url: URL;
+
+      try {
+        const proto = event.headers["x-forwarded-proto"] || "https";
+        url = new URL(path, `${proto}://${event.headers.host || ""}`);
+      } catch (err) {
+        throw new Error("Could not parse URL from `event.headers.host`");
+      }
 
       const isProduction =
         env.CONTEXT === "production" ||
--- a/src/lambda.ts
+++ b/src/lambda.ts
@@ -1,0 +1,134 @@
+import type {
+  APIGatewayEvent,
+  APIGatewayProxyEventV2,
+  APIGatewayProxyResult,
+  Context,
+} from "aws-lambda";
+import {
+  InngestCommHandler,
+  ServeHandler,
+} from "./components/InngestCommHandler";
+import { headerKeys, queryKeys } from "./helpers/consts";
+import { allProcessEnv } from "./helpers/env";
+
+/**
+ * With AWS Lambda, serve and register any declared functions with Inngest,
+ * making them available to be triggered by events.
+ *
+ * @example
+ *
+ * ```ts
+ * import { Inngest } from "inngest";
+ * import { serve } from "inngest/lambda";
+ *
+ * const inngest = new Inngest({ name: "My Lambda App" });
+ *
+ * const fn = inngest.createFunction(
+ *   { name: "Hello World" },
+ *   { event: "test/hello.world" },
+ *   async ({ event }) => {
+ *     return "Hello World";
+ *   }
+ * );
+ *
+ * export const handler = serve(inngest, [fn]);
+ * ```
+ *
+ * @public
+ */
+export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
+  const handler = new InngestCommHandler(
+    "aws-lambda",
+    nameOrInngest,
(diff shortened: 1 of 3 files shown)


src/lambda.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      const method = eventIsV2
+        ? event.requestContext.http.method
+        : event.httpMethod;
+      const path = eventIsV2 ? event.requestContext.http.path : event.path;
+
+      const env = allProcessEnv();
+      const proto = event.headers["x-forwarded-proto"] || "https";
+
+      const url = new URL(path, `${proto}://${event.headers.host || ""}`);
```
djfarrelly, line 66: It seems that this URL statement will fail if `host` is falsy, i.e. `new URL("/pathname", "https://")` - I'm guessing this is to defend against typescript saying `host` could be undefined. I wouldn't expect it to be undefined, but I'm wondering in the case that it could be `undefined`, this would throw a `failed to constructor 'URL'` error rather than something more useful for the user like `Lambda event host header undefined` 🤔

src/lambda.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      const isProduction =
+        env.CONTEXT === "production" ||
+        env.ENVIRONMENT === "production" ||
+        env.NODE_ENV?.startsWith("prod") ||
+        false;
+
+      return {
+        register: () => {
+          if (method === "PUT") {
```
djfarrelly, line 76: I'm curious, why does every handler need to have these conditional statements within each handler function? Could we instead pass the `method` as a string to the return statement like `register` and `run`? it feels like it's duplicative in these handlers.  i.e.

```ts
// for this example
return {
  method,
  register : () => {...
}
// for other handlers that use "req"
return {
  method: req.method,
  register : () => {...
}
```

Theoretically, we could do this for a few key standard fields like `method`, `env`, `isProduction`, `url` too

djfarrelly, line 76: I'm curious, why does every handler need to have these conditional statements within each handler function? Could we instead pass the `method` as a string to the return statement like `register` and `run`? it feels like it's duplicative in these handlers.  i.e.

```ts
// for this example
return {
  method,
  register : () => {...
}
// for other handlers that use "req"
return {
  method: req.method,
  register : () => {...
}
```

tonyhb:
Great stuff.  This early refactor you did has saved so much time and hassle with implementing new serve handlers!

src/lambda.ts
jpwilliams: @djfarrelly Yeah, a tiny refactor that's been on the list is to generically provide `env`, `isProduction`, and `url` alongside the methods as they're required for every call; just haven't done that refactor in this PR.

We don't explicitly provide `method` as it's technically an implementation detail - e.g. it may be that a particular platform converts every request to `GET`, or maybe it never exposes the method at all. Instead, we use the `register`/`run`/`view` functions to decide what to do as it gives us a few benefits over just returning a string:

- It's easier to see what cases need to be supported, rather than having to know the internals and therefore which method to pass if you need to create one on your own.

- We can use any detail (not just HTTP method) to decide what action we should take, so it's super flexible for different platforms.

- If we add a new way to interact with the SDK, let's say `tunnel`, we could add it as [a new `HandlerAction`](https://github.com/inngest/inngest-js/blob/main/src/components/InngestCommHandler.ts#L1039-L1080) and TypeScript will enforce that all handlers now account for that action.

- As with the above point, when writing a new serve handler, the use of this pattern forces the developer to account for every single action required by the SDK, so even someone writing a new handler for their custom framework is guided through the process rather than having to understand how Inngest works.

src/lambda.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      const env = allProcessEnv();
+      const proto = event.headers["x-forwarded-proto"] || "https";
+
+      const url = new URL(path, `${proto}://${event.headers.host || ""}`);
+
+      const isProduction =
+        env.CONTEXT === "production" ||
+        env.ENVIRONMENT === "production" ||
+        env.NODE_ENV?.startsWith("prod") ||
```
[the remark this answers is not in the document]
goodoldneon: 🎉

src/lambda.ts
jpwilliams: Fixed in b28dd08794c8933b73cd83d3e47025836eea2f1d.

PR #147, opened 2023-03-21
Refactor `InngestCommHandler` to better detect env and reduce duplication

### LINKED ISSUE #146: Refactor sporadic envs checks for production across serve handlers
## Summary

Most serve handlers provide their own logic for checking if the handler is in production mode or not, though we almost always use environment variables for this.

We should refactor these handlers to primarily use the existing [`isProd()`](https://github.com/inngest/inngest-js/blob/main/src/helpers/env.ts#L34-L42) call.

## Related
- https://github.com/inngest/inngest-js/pull/145#discussion_r1143524140

jpwilliams:
## Summary

Most serve handlers provide their own logic for checking if the handler is in production mode or not, though we almost always use environment variables for this. This PR:

- Refactors an existing [`isProd()`](https://github.com/inngest/inngest-js/blob/main/src/helpers/env.ts#L34-L42) call to better check multiple env vars
- Refactors `InngestCommHandler` to use `isProd()` and fetch env internally unless an override is provided in a serve handler
- Reduce duplication in serve handlers by unifying passing of `url`, `env`, and `isProduction`

## Related

- https://github.com/inngest/inngest-js/pull/145#discussion_r1143524140
- Closes #146

### DIFF
--- a/src/helpers/env.ts
+++ b/src/helpers/env.ts
@@ -31,14 +31,44 @@ export const devServerHost = (): string | undefined => {
   return values.find((a) => !!a);
 };
 
-export const isProd = (): boolean => {
-  const values = [
-    processEnv("NODE_ENV"),
-    processEnv("VERCEL_ENV"),
-    processEnv("CONTEXT"),
-  ];
+const prodCheckFns = {
+  equals: (actual, expected) => actual === expected,
+  "starts with": (actual, expected) =>
+    expected ? actual?.startsWith(expected) ?? false : false,
+  "is truthy": (actual) => Boolean(actual),
+} satisfies Record<
+  string,
+  (actual: string | undefined, expected: string | undefined) => boolean
+>;
 
-  return values.includes("production");
+const prodChecks: [
+  key: string,
+  customCheck: keyof typeof prodCheckFns,
+  value?: string
+][] = [
+  ["CF_PAGES", "equals", "1"],
+  ["CONTEXT", "starts with", "prod"],
+  ["ENVIRONMENT", "starts with", "prod"],
+  ["NODE_ENV", "starts with", "prod"],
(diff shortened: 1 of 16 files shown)


src/helpers/env.ts
```diff
@@ -31,14 +31,44 @@ export const devServerHost = (): string | undefined => {
...
+    expected ? actual?.startsWith(expected) ?? false : false,
+  "is truthy": (actual) => Boolean(actual),
+} satisfies Record<
+  string,
+  (actual: string | undefined, expected: string | undefined) => boolean
+>;
+
+const prodChecks: [
+  key: string,
+  customCheck: keyof typeof prodCheckFns,
```
goodoldneon, line 46: I love TypeScript so much. So much static protection 😍

src/test/helpers.ts
```diff
@@ -106,7 +106,7 @@ export const testFramework = (
    * Create a helper function for running tests against the given serve handler.
    */
   const run = async (
-    handlerOpts: Parameters<typeof handler["serve"]>,
+    handlerOpts: Parameters<(typeof handler)["serve"]>,
```
goodoldneon, line 109: New Prettier version, I'm assuming

goodoldneon: Could pin `devDependencies`, but now I'm bikeshedding so ignore me

jpwilliams: @goodoldneon Oh, no, as in I explicitly had to do `yarn upgrade prettier` or Prettier can't even parse the file. 😅 Very intentional upgrade. :)

goodoldneon (approved):

tonyhb (approved):
+1 on what Aaron said.  This code is really nice, easy to read, clean, and makes writing handlers much better.

Issue #150, opened 2023-03-24
Type errors in inngest types and in dependencies


mahdavipanah:
**Describe the bug**
Typescript (v4.4.3) gives type errors. The errors are both related to `inngest/types.d.ts` and Inngest's **type-fest** dependency module.

**To Reproduce**

1. Create an instance of `Inngest`:
```typescript
import { Inngest } from "inngest";

const client = new Inngest({ name: "My app" });
```
2. Run `tsc`.

**Expected behavior**
I expected the code to be compiled without any errors.

**Code snippets / Logs / Screenshots**
**Temporarily fix**:
Setting typescript's `skipLibCheck` compiler option to true will solve the problem temporarily as it skips the type checking on imported packages. `tsconfig.json`:
```json.
  ...
  "compilerOptions": {
    "skipLibCheck": true,
  },
  ...
```

**Error logs:**
```sh
node_modules/inngest/node_modules/type-fest/source/camel-case.d.ts:25:32 - error TS1005: '?' expected.

25  infer FirstWord extends string,
                                  ~

node_modules/inngest/node_modules/type-fest/source/camel-case.d.ts:26:42 - error TS1005: '?' expected.

26  ...infer RemainingWords extends string[],
                                            ~

node_modules/inngest/node_modules/type-fest/source/join.d.ts:32:56 - error TS1005: '?' expected.

32    ...infer Rest extends ReadonlyArray<string | number>,
                                                          ~

node_modules/inngest/types.d.ts:137:45 - error TS1005: '>' expected.

137     tools: ReturnType<typeof createStepTools<TEvents, TTrigger>>[0];
                                                ~

node_modules/inngest/types.d.ts:137:64 - error TS1005: '(' expected.

137     tools: ReturnType<typeof createStepTools<TEvents, TTrigger>>[0];
                                                                   ~

node_modules/inngest/types.d.ts:174:1 - error TS1128: Declaration or statement expected.

174 };
    ~


Found 6 errors.

error Command failed with exit code 2.
```

**System info:**

- OS: Mac
- npm package Version 1.5.1
- Framework Express
- Platform Netlify

jpwilliams:
Thank you for the report!

We must look into additional integration tests for different TS versions to root these problems out earlier.

[...]

PR #151, opened 2023-03-24
INN-1009 Add CI tests for different TS versions to ensure back compat

jpwilliams:
## Summary INN-1009

Providing a TS library means supporting various versions of TypeScript. Above just watching public API changes, this means:

- Ensuring we have an explicit earliest supported TS version
- Ensuring we have an explicit earliest supported Node version
- Ensuring that earlier TS versions maintain the same level of inference across all tooling

TypeScript introduce breaking changes with minor increments, which seems to be the norm among tooling with a large surface area (see [Follow SemVer · Issue #2888 · jashkenas/backbone (github.com)](https://github.com/jashkenas/backbone/issues/2888#issuecomment-29076249)). This also means that TS changes faster than it appears to and that it's harder for users to upgrade TS versions, even though semver might suggest otherwise.

Most TS configs will also type-check any installed libraries every time code is built, which results in errors on TS `<4.7.0` as the earlier compiler can't even parse the file due to unexpected characters (see #150).

## Actions

There are some fantastic resources covering this problem (see https://www.semver-ts.org/), but even tooling such as [downlevel-dts](https://github.com/sandersn/downlevel-dts) takes a long time to support the latest TS releases (it currently still doesn't support some 4.7 features) _and_ some of the "down-leveling" just falls back to wider types such as `any` or `unknown`, which could be insufficient depending on where it is.

The easiest step to take is to add some CI testing for various TypeScript versions and ensure that tests pass and compile. We have a lot of type-only unit tests to assert inference functionality, so testing that against earlier TS versions should do the trick.

With this, we can pretty easily support >=4.7.0, though going any further would require some chonky rewrites. The main culprit being [instantiation expressions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#instantiation-expressions) released in [TypeScript 4.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html). This PR is mostly about adding a framework for asserting and committing to a specific TS version support, then support for earlier versions can be added in further PRs.

Time-wise, it means we can comfortably support TS versions within the last 10 months, and specify that fact in our `peerDependencies`.

- ✅ **TypeScript 5.0** - March 2023 (<1mo)
- ✅ **TypeScript 4.9** - November 2022 (~4mo)
- ✅ **TypeScript 4.8** - August 2022 (~7mo)
- ✅ **TypeScript 4.7** - May 2022 (~10mo)
- ❌ **TypeScript 4.6** - Feb 2022 (~1y1mo)
- ❌ **TypeScript 4.5** - November 2021 (~1y4mo)
- ❌ **TypeScript 4.4** - August 2021 (~1y7mo)
- ❌ **TypeScript 4.3** - May 2021 (~1y10mo)
- ❌ **TypeScript 4.2** - Feb 2021 (~2y1m)
- ❌ **TypeScript 4.1** - November 2020 (~2y4m)
- ❌ **TypeScript 4.0** - August 2020 (~2y7mo)

To test types, we'll do a couple of actions:

- [x] Add a new `test:types` script that will test that `dist/` and `src/**/*.test.ts` compile correctly.
This will ensure that users with the default of `skipLibCheck: false` can properly build the code, and compiling test files ensures that all type tests are also working.
- [x] Add CI matrix to test multiple TS versions against the above files

## Related

- #150

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,5 +99,8 @@
   },
   "peerDependencies": {
     "typescript": ">=4.7.2"
+  },
+  "engines": {
+    "node": ">=14"
   }
 }
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "build": "yarn run clean && tsc --project tsconfig.build.json",
     "test": "node --expose-gc --max-old-space-size=4096 ./node_modules/.bin/jest --silent --logHeapUsage --maxWorkers=8 --coverage --ci --verbose",
     "test:examples": "node --expose-gc --max-old-space-size=4096 ./node_modules/.bin/jest --logHeapUsage --maxWorkers=8 --testMatch \"**/examples/**/*.test.ts\" --ci --verbose",
+    "test:types": "tsc --noEmit --project tsconfig.types.json",
     "clean": "rm -rf ./dist",
     "lint": "eslint .",
     "postversion": "yarn run build && yarn run build:copy",
--- a/package.json
+++ b/package.json
@@ -95,5 +95,8 @@
   "volta": {
     "node": "18.12.1",
     "yarn": "1.22.19"
+  },
+  "peerDependencies": {
+    "typescript": ">=4.7.2"
   }
 }
--- a/.changeset/popular-dogs-sniff.md
+++ b/.changeset/popular-dogs-sniff.md
@@ -1,0 +1,7 @@
+---
+"inngest": patch
+---
+
+INN-1009 Show warnings when using the package with TS versions `<4.7.2` and Node versions `<14`
+
+This includes tests to assert we appropriately support these versions now and in the future.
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -95,5 +96,8 @@
   "volta": {
     "node": "18.12.1",
     "yarn": "1.22.19"
+  },
+  "peerDependencies": {
+    "typescript": ">=4.7.2"
   }
 }
```
IGassmann, line 103: **suggestion:** This would warn the user, if he installs the package with a Node.js version lower than 14.

See https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines

```suggestion
  },
  "engines": {
    "node": ">=14"
  }
}
```

IGassmann (approved):
Amazing work!

PR #157, opened 2023-03-29
INN-1054 Ensure `ServeHandler` always returns `any`

jpwilliams:
## Summary INN-1054

Serve handlers used to all return any in the past, but have changed to unknown in #135.

This is a mistake. We need any so that the output can be appropriately slotted into any requirement the user has.

`any` feels dangerous here, but is appropriate: we can infer the types for the returned function appropriately, e.g. `(req: NextRequest, res: NextResponse) => NextResponse`, but this is also a misstep. To do this, we must install a package for the `NextRequest` and `NextResponse` types, but the package we use could be different to the one the user uses. This can create a drift in types and cause a collision where there shouldn't be one.

The fix is just to revert all handlers to appropriately return `any`.

## Related

- #135
- INN-1054

### DIFF
--- a/.changeset/late-kings-admire.md
+++ b/.changeset/late-kings-admire.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+INN-1054 Ensure serve handlers return `any` instead of `unknown` so that they don't needlessly conflict with user types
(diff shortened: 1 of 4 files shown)


jpwilliams:
GitHub Actions [was down](https://www.githubstatus.com/incidents/z3c6q056q332).

Test suite [ran manually](https://github.com/inngest/inngest-js/actions/runs/4564083172).

PR #158, opened 2023-03-29
INN-1029 Add `env` option to `Inngest` client and find branch names from env vars

jpwilliams:
## Summary INN-1029

Adds the ability for a user to explicitly define an Inngest environment to send events to.

The workspace the event should go to is defined by the event key being used, but this can be used to direct events to another parallel workspace if allowed.

This is a relatively busy PR as it encapsulates a few changes to testing, too. Key files are:

- [src/helpers/env.ts](https://github.com/inngest/inngest-js/pull/158/files#diff-6419f12dc66fc6ed8ae4c92c68c4da1b5a4837494267428c9390a01986a8db1e)
Adding new `getEnvironmentName()` and `inngestHeaders()` exports to grab generated headers from the environment.
- [src/components/Inngest.ts](https://github.com/inngest/inngest-js/pull/158/files#diff-485bedbbee0f0a94697b7e5fa853e50e4e51d38c90f1b6df30da246314ded58b)
Adding an `env` option and building headers using `inngestHeaders()`.
- [src/components/InngestCommHandler.ts](https://github.com/inngest/inngest-js/pull/158/files#diff-5087958ec6f8b087e8827c51631a00210094a1d599c750e7ccade1b12ffc59c5)
Building headers to respond with during registration using `inngestHeaders()` and providing the `x-inngest-framework` header too.

## Related

- INN-1030

### DIFF
--- a/src/helpers/consts.ts
+++ b/src/helpers/consts.ts
@@ -31,12 +31,22 @@ export enum envKeys {
   VercelBranch = "VERCEL_GIT_COMMIT_REF",
 
   /**
+   * Expected to be `"1"` if defined.
+   */
+  IsVercel = "VERCEL",
+
+  /**
    * The branch name of the current deployment. May only be accessible at build
    * time, but included here just in case.
    *
    * {@link https://developers.cloudflare.com/pages/platform/build-configuration/#environment-variables}
    */
   CloudflarePagesBranch = "CF_PAGES_BRANCH",
+
+  /**
+   * Expected to be `"1"` if defined.
+   */
+  IsCloudflarePages = "CF_PAGES",
 
   /**
    * The branch name of the deployment from Git to Netlify, if available.
@@ -46,6 +56,11 @@ export enum envKeys {
   NetlifyBranch = "BRANCH",
 
   /**
+   * Expected to be `"true"` if defined.
+   */
+  IsNetlify = "NETLIFY",
+
+  /**
    * The Git branch for a service or deploy.
    *
    * {@link https://render.com/docs/environment-variables#all-services}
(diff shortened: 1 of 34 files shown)


src/helpers/consts.ts
```diff
@@ -19,6 +19,45 @@ export enum envKeys {
...
+   * {@link https://docs.netlify.com/configure-builds/environment-variables/#git-metadata}
+   */
+  NetlifyBranch = "BRANCH",
+
+  /**
+   * The Git branch for a service or deploy.
+   *
+   * {@link https://render.com/docs/environment-variables#all-services}
+   */
+  RenderBranch = "RENDER_GIT_BRANCH",
```
djfarrelly, line 53: Nice work on all of these! 🙌

src/helpers/env.ts
```diff
@@ -74,6 +74,31 @@ export const isProd = (
...
 };
 
+/**
+ * getBranchName returns the suspected branch name for this environment by
+ * searching through a set of common environment variables.
+ *
+ * This could be used to determine if we're on a branch deploy or not, though it
+ * should be noted that we don't know if this is the default branch or not.
+ */
+export const getBranchName = (
```
djfarrelly, line 84: Hey @jpwilliams - I see this is declared, but it's not used in the code - should we be using this in the `Inngest.ts`?

goodoldneon: Also, should this function be called `getEnvironmentName`? I know the provider-specific env vars are for branches, but that's kind of an implementation detail to our users. Also, the first variable we check is `INNGEST_ENV` so it's a little unintuitive to have it considered in a "branch" function

jpwilliams: Hi chaps - yeah, this is from the end of last week where the implementation was a little fuzzy and I had removed this functionality. See 5049a6412a8d662a5f8f65ff992e59803fa0fafd.

Re-adding this now with the new spec.

PR #163, opened 2023-03-31
INN-1087 Add edge streaming support to all serve handlers

jpwilliams:
## Summary INN-1087

Adds the ability for the SDK to stream responses back to Inngest, enabling longer timeouts in edge runtimes such as Deno, Cloudflare, or Next.js.

- **Add edge streaming support**
  > We're currently very strict about streaming support, only allowing it for `"inngest/next"`, ensuring users must opt in using `allowEdgeStreaming: true` and only allowing Next.js projects on Vercel.
  >
  > Following this being tested in production for a while, streaming support for other supported platforms is trivial to add, just writing code to turn it on.
- **Add support for Next.js Edge Functions**
  > The `"inngest/next"` handler now supports both Serverless Functions and Edge Functions; users can switch between both using Next.js conventions without having to change any of our code or imports.
- **Add support for Next.js 13+ [Route Handlers](https://beta.nextjs.org/docs/routing/route-handlers)**
  > Ended up wanting this as part of testing; if folks are using edge/streaming then it feels more likely that they're happy to use a beta version of Next.js itself too, so we might as well support this.
  > 
  > Route handlers use the same `"inngest/next"` serve handler, but users must export differently:
  > ```ts
  > export default serve(...);
  > // to
  > export const { GET, POST, PUT } = serve(...);
  > ```

## Related

- inngest/inngest#429

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -26,6 +26,7 @@ import {
   RegisterOptions,
   RegisterRequest,
   StepRunResponse,
+  SupportedFrameworkName,
 } from "../types";
 import { version } from "../version";
 import { Inngest } from "./Inngest";
@@ -560,6 +561,7 @@ export class InngestCommHandler<
         this.allowEdgeStreaming &&
         this.streamTransformRes &&
         platformSupportsStreaming(
+          this.frameworkName as SupportedFrameworkName,
           actions.env as Record<string, string | undefined>
         )
       ) {
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -490,9 +490,6 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
         ])
       ) as typeof rawActions;
 
-      /**
-       * TODO Allow trailing header ONLY IF the incoming request allows it
-       */
       const getHeaders = () => ({
         ...inngestHeaders({
           env: actions.env as Record<string, string | undefined>,
@@ -511,11 +508,13 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
         const runRes = await actions.run();
         if (runRes) {
           const { stream, finalize } = await this.createStream();
+
           /**
-           * Is it okay if this throws?
+           * Errors are handled by `handleAction` here to ensure that an
+           * appropriate response is always given.
            */
           void actionRes.then(finalize);
-          res = { status: 200, body: stream, headers: getHeaders() };
+          res = { status: 201, body: stream, headers: getHeaders() };
         }
       }
 
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -1103,7 +1103,9 @@ type Handler = (...args: any[]) => {
  * The response from the Inngest SDK before it is transformed in to a
  * framework-compatible response by an {@link InngestCommHandler} instance.
  */
-export interface ActionResponse {
+export interface ActionResponse<
+  TBody extends string | ReadableStream = string
+> {
   /**
    * The HTTP status code to return.
    */
@@ -1117,7 +1119,7 @@ export interface ActionResponse {
   /**
    * A stringified body to return.
    */
-  body: string | ReadableStream;
+  body: TBody;
 }
 
 /**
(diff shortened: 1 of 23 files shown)


src/types.ts
```diff
@@ -493,6 +493,8 @@ export interface RegisterOptions {
    * Default level: "info"
    */
   logLevel?: LogLevel;
+
+  allowEdgeStreaming?: boolean;
```
jpwilliams, line 497: Needs documenting.

src/lambda.ts
```diff
@@ -117,7 +117,7 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
         body,
         statusCode: status,
         headers,
-      });
+      } as APIGatewayProxyResult);
```
jpwilliams, line 120: This is hiding the fact that `body` can be a `ReadableStream`. 🤮

I don't like that these response transformers now have to _maybe_ deal with streaming. Perhaps a better solution would be to provide a second response transformer if streaming is supported, in which case we can also easily ascertain internally whether a handler supports it or not.

This also likely means we can safely make edge streaming opt-out instead of opt-in, as handlers themselves can declare support rather than having to assume it'll work.

src/components/InngestCommHandler.ts
```diff
@@ -524,7 +587,6 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
                 )
             ),
             headers: {
-              ...getHeaders(),
```
jpwilliams, line 527: > Why were the headers removed? - @goodoldneon 

There was a lot of duplicated code here where headers were set for almost every response.

This is now handled in a single place higher up in `createHandler()` (see [streamed headers](https://github.com/inngest/inngest-js/pull/163/files#diff-5087958ec6f8b087e8827c51631a00210094a1d599c750e7ccade1b12ffc59c5R518) and [regular headers](https://github.com/inngest/inngest-js/pull/163/files#diff-5087958ec6f8b087e8827c51631a00210094a1d599c750e7ccade1b12ffc59c5R527)), then merged with any _extras_ that are passed here, usually just `Content-Type`.

src/components/InngestCommHandler.ts
```diff
@@ -461,13 +471,66 @@ export class InngestCommHandler<H extends Handler, TransformedRes> {
...
+      if (this.allowEdgeStreaming && isEdgeRuntime()) {
+        const runRes = await actions.run();
+        if (runRes) {
+          const { stream, finalize } = await createStream();
+
+          /**
+           * Errors are handled by `handleAction` here to ensure that an
+           * appropriate response is always given.
+           */
+          void actionRes.then(finalize);
```
jpwilliams, line 517: We know that `actionRes` is an `ActionResponse` here, so we should probably also run `getHeaders()` and merge in any trailing headers with the response. 🤔

tonyhb: Do we ever need to run `transformRes` on streaming responses?  This is called on L523 for non-streaming responses so I'm double checking :)

src/next.ts
```diff
@@ -1,70 +1,117 @@
...
 export const name = "nextjs";
 
+/**
+ * Assert whether the given request is a Next.js API request or an `NextRequest`
+ * from an edge function.
+ */
+const isEdgeRequest = (
+  req: NextApiRequest | NextRequest
+): req is NextRequest => {
+  return typeof req?.headers?.get === "function";
```
tonyhb, line 20: This is in `src/next` which is almost always deployed to vercel, but should this be platform-specific instead of framework specific?

jpwilliams: @tonyhb Have made these changes - see https://github.com/inngest/inngest-js/pull/163/files#diff-5087958ec6f8b087e8827c51631a00210094a1d599c750e7ccade1b12ffc59c5R536-R542

tonyhb (approved):
This. Is. AMAZING!

ludalex:
Apologies for the reply here, if there is a better place please do tell.

Is there a way to enable streaming only for specific functions? 
I need to support long-timeout tasks (2-3m) on Next.js/Vercel, but I have Prisma as a Inngest middleware which prevents me to set the Inngest endpoit as Edge Function.

PR #172, opened 2023-04-18
Filter undefined functions prior to serving

tonyhb:
People can mess up imports without typescript, causing errors.


jpwilliams (approved):
Ace! Did we see this from a user?

PR #180, opened 2023-04-26
INN-1270 Create a handler to enforce more actionable errors

jpwilliams:
## Summary INN-1270

Many errors within the SDK are handled in code, but not logged for users to see, meaning it can be unclear for users how failures should be remedied. We can fix this by providing a number of more in-depth, user-facing errors across the SDK.

This PR creates a small helper we can use across the app to take in and error and enforce that we:

- Say what happened
- Provide reassurance if we can
- Say why it happened
- Say how to fix it now
- Give them a back-up plan

It replaces some existing warnings, and adds TODOs for all other areas we must address.

## Example

image

### DIFF
--- a/.changeset/unlucky-buttons-develop.md
+++ b/.changeset/unlucky-buttons-develop.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+INN-1270 Create an internal handler to enforce more actionable user-facing errors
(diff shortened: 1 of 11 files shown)


tonyhb (commented):
These errors are REALLY nice.  This is great!

PR #184, opened 2023-05-09
Ensure that we pass env from client into headers

tonyhb:
Without this, adding `env` to our client is not respected during registration.  This is necessary for eg. Netlify.

### DIFF
--- a/src/helpers/env.ts
+++ b/src/helpers/env.ts
@@ -3,6 +3,7 @@
 // along with prefixes, meaning we have to explicitly use the full `process.env.FOO`
 // string in order to read variables.
 
+import { type Inngest } from "../components/Inngest";
 import { SupportedFrameworkName } from "../types";
 import { version } from "../version";
 import { envKeys, headerKeys } from "./consts";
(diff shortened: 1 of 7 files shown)


src/helpers/env.ts
```diff
@@ -171,7 +191,10 @@ export const inngestHeaders = (opts?: {
     headers[headerKeys.Framework] = opts.framework;
   }
 
-  const env = opts?.env || allProcessEnv();
+  const env = {
+    ...allProcessEnv(),
+    ...opts?.env,
+  };
```
jpwilliams, line 197: This was a good shout, @tonyhb!

src/helpers/env.ts
```diff
@@ -183,7 +206,11 @@ export const inngestHeaders = (opts?: {
     headers[headerKeys.Platform] = platform;
   }
 
-  return headers;
+  return {
+    ...headers,
+    ...opts?.client?.["headers"],
+    ...opts?.extras,
+  };
```
jpwilliams, line 213: Making sure we do this in one place now.

jpwilliams (approved):

PR #187, opened 2023-05-11
INN-1347 INN-1349 Fix deadlock when an async function finds a step

jpwilliams:
## Summary INN-1347 INN-1349

Currently, if we detect that a user has a regular async function and is using no step tooling, we eventually await the function. In cases where the function declares a step after this decision has been made, the step never resolves, therefore the function never resolves.

We can fix this by adding an extra check at this point, either per-tick or step functions synchronously throwing an error.

This includes the fix for for INN-1349, making `step.sendEvent()` an inline action if run within a non-step function.

## Related

- #188

### DIFF
--- a/src/components/InngestStepTools.ts
+++ b/src/components/InngestStepTools.ts
@@ -1,6 +1,7 @@
 import canonicalize from "canonicalize";
 import { sha1 } from "hash.js";
 import { type Jsonify } from "type-fest";
+import { prettyError } from "../helpers/errors";
 import { timeStr } from "../helpers/strings";
 import {
   type ObjectPaths,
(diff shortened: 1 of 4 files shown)


src/components/InngestStepTools.ts
```diff
@@ -88,6 +89,20 @@ export const createStepTools = <
...
+     *
+     * Created step tooling can decide how to manually handle this on a
+     * case-by-case basis.
+     *
+     * In the future, we can provide a way for a user to override this if they
+     * wish to and understand the danger of side-effects.
+     *
+     * Defaults to `false`.
+     */
+    nonStepFnDetected: boolean;
```
darwin67, line 105: Can this be computed instead of making it a flag?

darwin67 (approved):
Seems okay to me but best to have 2nd pairs of eyes to make sure.

General feedback is to avoid `flags` and instead compute the results as much as possible.
Makes it easier to test branch changes or even state status that way.
The permutation and mental load when need to read flags can be hard to handle sometimes.

src/components/InngestStepTools.ts
jpwilliams: @darwin67 We'd still need to know a little more about the state than we currently do in order to calculate this on-the-fly, at least that we've done memoizing state.

That in itself would require a flag or us marking stack items as we pass them, so might not be worth it here.

Agreed, though - there are a probably a few other places this approach would help clear up logic!

PR #188, opened 2023-05-11
INN-1348 Throw an actionable error when we detect mixed async logic

jpwilliams:
## Related

- #187 
- INN-1348

### DIFF
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -1,5 +1,9 @@
 import { internalEvents, queryKeys } from "../helpers/consts";
-import { deserializeError, serializeError } from "../helpers/errors";
+import {
+  deserializeError,
+  functionStoppedRunningErr,
+  serializeError,
+} from "../helpers/errors";
 import { resolveAfterPending, resolveNextTick } from "../helpers/promises";
 import { type ServerTiming } from "../helpers/ServerTiming";
 import { slugify, timeStr } from "../helpers/strings";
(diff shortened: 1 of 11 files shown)


src/components/InngestFunction.ts
```diff
@@ -485,6 +499,30 @@ export class InngestFunction<
...
+         * promises to return no further actions, so we should only throw if
+         * this state is reached and there are no other pending steps.
+         */
+        const hasOpsPending = Object.values(state.allFoundOps).some((op) => {
+          return op.fulfilled === false;
+        });
+
+        if (!hasOpsPending) {
+          throw new Error(
+            functionStoppedRunningErr(ErrCode.ASYNC_DETECTED_AFTER_MEMOIZATION)
```
jpwilliams, line 523: Should throwing any of these errors disable retries? e.g. return `4xx`. I think yes.

darwin67: I agree retries should not attempt since there's no point until user update their code

darwin67 (approved):

darwin67:
Is there a follow up task to communicate this change to users when releasing?
Best people are aware and know what changed in the new release so people's code don't blow up all of a sudden, and we don't suddenly get a spike of support requests.

tonyhb (approved):

tonyhb:
> Is there a follow up task to communicate this change to users when releasing? Best people are aware and know what changed in the new release so people's code don't blow up all of a sudden, and we don't suddenly get a spike of support requests.

I dont think we need this, as existing code is already broken and wouldn't have been functioning.  Now, users will at least start seeing errors.

When we enable this, via an unsafe flag, we should probably let users know.

PR #189, opened 2023-05-12
INN-1339: Add interface accepting logger implementations for use

darwin67:
## Summary

Allow registering an arbitrary logger during the initialization of Inngest.
If not passed in, it'll default to a logger that can only log to the console.
```js
const inngest = new Inngest({ name: "<>", logger: logger })
```

This will enable a logger within all Inngest functions when declaring user functions.
```js
async ({ event, step, logger }) => {
  logger.info("hello")
}
```

This is the first step and is simply a passthrough right now. Follow up PRs will update this to a Proxy and treat it as an `Op`.

The defined API interfaces are based on the top most used loggers in JS.
js-loggers

## Tests
Manual functional tests locally using Winston and a Datadog transport for logger delivery.
logger-function-test

Automated tests will be added in follow up PRs, when adding things like Proxy implementation.

### DIFF
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -85,7 +85,7 @@ export class Inngest<
 
   private readonly fetch: FetchT;
 
-  public readonly logger: Logger;
+  private readonly logger: Logger;
 
   /**
    * A client used to interact with the Inngest API by sending or reacting to
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -25,6 +25,7 @@ import {
   type TriggerOptions,
 } from "../types";
 import { InngestFunction } from "./InngestFunction";
+import { type Logger, DefaultLogger } from "../middleware/logger";
 
 /**
  * Capturing the global type of fetch so that we can reliably access it below.
@@ -84,6 +85,8 @@ export class Inngest<
 
   private readonly fetch: FetchT;
 
+  public readonly logger: Logger;
+
   /**
    * A client used to interact with the Inngest API by sending or reacting to
    * events.
@@ -113,6 +116,7 @@ export class Inngest<
     inngestBaseUrl = "https://inn.gs/",
     fetch,
     env,
+    logger,
   }: ClientOptions) {
     if (!name) {
       // TODO PrettyError
@@ -143,6 +147,9 @@ export class Inngest<
     });
 
     this.fetch = getFetch(fetch);
+
+    // TODO: Wrap this in a proxy for non default loggers
+    this.logger = logger ? logger : new DefaultLogger();
   }
 
   /**
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -17,6 +17,7 @@ import { strBoolean } from "../helpers/scalar";
 import { createStream } from "../helpers/stream";
 import { stringifyUnknown } from "../helpers/strings";
 import { type MaybePromise } from "../helpers/types";
+import { type ILogger } from "../middleware/logger";
 import { landing } from "../landing";
 import {
   type FunctionConfig,
@@ -252,6 +253,17 @@ export class InngestCommHandler<
    */
   protected readonly logLevel: LogLevel;
 
+  /**
+   * An optional logger provided by the user to be used within all functions
+   *
+   * Expects the following interfaces to be defined
+   * - .info()
+   * - .warn()
+   * - .error()
+   * - .debug()
+   */
+  protected readonly logger?: ILogger;
+
   protected readonly streaming: RegisterOptions["streaming"];
 
   /**
@@ -307,6 +319,7 @@ export class InngestCommHandler<
       fetch,
       landingPage,
       logLevel = "info",
+      logger,
       signingKey,
       serveHost,
       servePath,
@@ -460,6 +473,7 @@ export class InngestCommHandler<
     this.serveHost = serveHost;
     this.servePath = servePath;
     this.logLevel = logLevel;
+    this.logger = logger;
     this.streaming = streaming ?? false;
 
     this.fetch = getFetch(
@@ -836,7 +850,7 @@ export class InngestCommHandler<
           }) ?? [];
 
       const ret = await fn.fn["runFn"](
-        { event, runId: ctx?.run_id },
+        { event, runId: ctx?.run_id, logger: this.logger },
         opStack,
         /**
          * TODO The executor is sending `"step"` as the step ID when it is not
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -17,7 +17,6 @@ import { strBoolean } from "../helpers/scalar";
 import { createStream } from "../helpers/stream";
 import { stringifyUnknown } from "../helpers/strings";
 import { type MaybePromise } from "../helpers/types";
-import { type Logger } from "../middleware/logger";
 import { landing } from "../landing";
 import {
   type FunctionConfig,
@@ -253,17 +252,6 @@ export class InngestCommHandler<
    */
   protected readonly logLevel: LogLevel;
 
-  /**
-   * An optional logger provided by the user to be used within all functions
-   *
-   * Expects the following interfaces to be defined
-   * - .info()
-   * - .warn()
-   * - .error()
-   * - .debug()
-   */
-  protected readonly logger?: Logger;
-
   protected readonly streaming: RegisterOptions["streaming"];
 
   /**
@@ -319,7 +307,6 @@ export class InngestCommHandler<
       fetch,
       landingPage,
       logLevel = "info",
-      logger,
       signingKey,
       serveHost,
       servePath,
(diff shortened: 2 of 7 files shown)


src/middleware/logger.ts
```diff
@@ -0,0 +1,11 @@
+import { type LogArg } from "../helpers/types";
+
+// Based on https://datatracker.ietf.org/doc/html/rfc5424#autoid-11
+// it's pretty reasonable to expect a logger to have the following interfaces
+// available.
+export interface ILogger {
```
goodoldneon, line 18: Nit: the name `ILogger` is Hungarian Notationy. Can we use `Logger` instead?

src/components/InngestFunction.ts
```diff
@@ -251,7 +252,7 @@ export class InngestFunction<
      * add tools.
      */
     const fnArg = {
-      ...(data as { event: EventPayload }),
+      ...(data as { event: EventPayload; logger: ILogger }),
```
goodoldneon, line 255:
```suggestion
      ...(data as { event: EventPayload; logger?: ILogger }),
```

Is `logger` optional?

src/middleware/logger.ts
darwin67: Sure, didn't know that's a thing.

src/components/InngestFunction.ts
darwin67: yup, nice catch.

goodoldneon: It'll always be there when they specify `logger` upon instantiation, right? If so, is there a way to make this optionality dependent on whether they specified `logger`?

darwin67: > It'll always be there when they specify logger upon instantiation, right?
Correct.

> If so, is there a way to make this optionality dependent on whether they specified logger?
That's some TS black magic I'm not aware of.
I can take a look. Hopefully the code doesn't result in something scary.

darwin67: This is the only article I found that seems to explain forcefully unwrapping optional arguments.
https://medium.com/terria/typescript-transforming-optional-properties-to-required-properties-that-may-be-undefined-7482cb4e1585

Definitely something I don't want to put my feet into, so here's a proposal.

If doing `logger?.info` seems too annoying, I can implement a dummy logger and make the type required instead of optional.
The dummy logger can just do `console.log` or something of the kind.

goodoldneon: Could use a generic. However, I needed to add a `// @ts-expect-error` because TS isn't smart enough to realize this is safe:

```ts
type Logger = {};

class InngestFunction<T extends Logger | undefined = undefined> {
  logger: T;

  constructor({ logger = undefined }: { logger?: T }) {
    // @ts-expect-error TS is unhappy because it's unable to recognize that
    // logger is only undefined when the constructor param is unspecified. In
    // other words, TS is trying to protect us against the impossible scenario
    // where we assign undefined to this.logger when the generic T is not
    // undefined.
    this.logger = logger;
  }
}

// a: {}
const a = new InngestFunction({ logger: {} }).logger;

// b: undefined
const b = new InngestFunction({ logger: undefined }).logger;

// c: undefined
const c = new InngestFunction({}).logger;
```

jpwilliams: > If doing logger?.info seems too annoying or non ergonomic, I can implement a dummy logger and make the type required instead of optional.
The dummy logger can just do console.log or something of the kind.

@darwin67 Yeah, I think it's reasonable to have a default `logger` that proxies `console` if nothing has been specified by the user.

Helps us avoid `logger?.*` both for the user and the fun type inference above we'd have to do to definitively say it either does or doesn't exist within the function body.

jpwilliams (changes requested):
Awesome!

One change this will need is that we should specify the logger when creating the _client_ rather than when _serving_, e.g.
```ts
const inngest = new Inngest({
  name: "My App",
  logger: myLogger,
});

export default serve(inngest, [...fns]);
```
This should just be shifting the `logger` option from `RegisterOptions` to `ClientOptions` in the same file, then some `logger` properties in different spots.

The reason for this is that only the client is used to create functions; if we want to do any fancy inference for logging in the future, we'd need it specified there in order to do so.

Once that's there, this is looking great to silently merge!

src/components/InngestFunction.ts
goodoldneon: Default/dummy logger sounds best 👍

src/components/InngestCommHandler.ts
```diff
@@ -17,6 +17,7 @@ import { strBoolean } from "../helpers/scalar";
 import { createStream } from "../helpers/stream";
 import { stringifyUnknown } from "../helpers/strings";
 import { type MaybePromise } from "../helpers/types";
+import { type Logger } from "../middleware/logger";
```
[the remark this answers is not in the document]
goodoldneon: You're manually reordering imports to make ESLint happy? Have you used the Prettier plugin https://github.com/IanVS/prettier-plugin-sort-imports?

src/middleware/logger.ts
```diff
@@ -0,0 +1,23 @@
...
+ * All kinds of arguments can come through
+ *
+ * Examples seen are
+ * - string
+ * - object / hash
+ * - values used for string interpolation, basically anything
+ *
+ * See https://linear.app/inngest/issue/INN-1342/flush-logs-on-function-exitreturns for more details
+ */
+type LogArg = object;
```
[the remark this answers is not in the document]
goodoldneon: Would `unknown` work? Every time an engineer writes an `any` an angel gets its wings clipped

src/components/InngestCommHandler.ts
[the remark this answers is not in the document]
jpwilliams: > You're manually reordering imports to make ESLint happy?

@goodoldneon Oh heck no - all done automatically on save, but because of some creative use of `tsconfig.json` files for testing against different TS versions, sometimes automatically-generated imports were being added as `inngest/my/import` instead of relative imports.

Think it's sorted now, but just checking. :D

src/middleware/logger.ts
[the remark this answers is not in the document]
jpwilliams: @goodoldneon 💯

src/components/InngestCommHandler.ts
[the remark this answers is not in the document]
darwin67: I add the line manually but something that's running in `yarn dev` keep warning me to change it to a `type` and I just didn't bother to fight it.

src/middleware/logger.ts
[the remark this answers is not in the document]
darwin67: I'd prefer to not use `any` either.
I end up with `object` because I got a recommendation with maybe the linter when running `yarn dev`.
I can also just do `string | number | object` too and that should cover pretty much everything.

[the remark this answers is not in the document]
goodoldneon: We don't care what type the arguments are, right? I'd just use `unknown`. The `unknown` type is basically the same as `any` but it forces you to do type assertions before doing anything with it.

TypeScript's `unknown` is a lot like Go's `any` (a.k.a. `interface {}`)

darwin67:
rebased against master

src/components/Inngest.ts
```diff
@@ -143,6 +147,9 @@ export class Inngest<
     });
 
     this.fetch = getFetch(fetch);
+
+    // TODO: Wrap this in a proxy for non default loggers
+    this.logger = logger ? logger : new DefaultLogger();
```
jpwilliams, line 152: Nit
```suggestion
    this.logger = logger ?? new DefaultLogger();
```

src/components/Inngest.ts
```diff
@@ -84,6 +85,8 @@ export class Inngest<
 
   private readonly fetch: FetchT;
 
+  public readonly logger: Logger;
```
jpwilliams, line 88: Would be great to make this `private`; the fewer options the user sees when typing `inngest.`, the better.
```suggestion
  private readonly logger: Logger;
```
This may change some other code to require `this.#client["logger"]` instead of `this.#client.logger`.

src/types.ts
```diff
@@ -391,6 +397,26 @@ export interface ClientOptions {
...
+   * The provider logger is expected to implement the following API interfaces
+   * - .info()
+   * - .warn()
+   * - .debug()
+   * - .error()
+   * which most loggers already do.
+   *
+   * Defaults to a dummy logger that just log things to the console if nothing is provided.
+   *
+   * TODO: Wrap the logger with a Proxy to make it operate like a step.
```
jpwilliams, line 415: Can kill this line now 🥳

```suggestion
```

src/components/Inngest.ts
jpwilliams: Ah - I see the next PR adjusts this anyway.

src/types.ts
jpwilliams: Ah - I see the next PR adjusts this anyway.

jpwilliams (approved):

PR #194, opened 2023-05-12
INN-1342: Add ProxyLogger to help handle buffering and flushing of logs (experimental)

darwin67:
## Summary

Built on #189.
Resolves INN-1343 as well.

Wrap the passed in `Logger`, and at function returns, wait a little to allow provided logger to flush.

This change is not expected to be widely used yet, and more as an experimental feature that we can start using ourselves to see if it works as expected.

### Notes
Even with the current implementation, there could be a minor impact on function execution.
Something to address in the future is to make the 1s wait only run on serverless environments.

When ran as part of an application server, the whole `flush` method is very likely to be irrelevant completely.
This is all easier to address once we have some kind of middleware interface that can be plugged in.

The approach of enable/disable logger is a temporary solution until we provide a middleware solution.

### DIFF
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -320,18 +320,28 @@ export class InngestFunction<
 
       // eslint-disable-next-line @typescript-eslint/no-misused-promises, no-async-promise-executor
       const userFnPromise = new Promise(async (resolve, reject) => {
+        if (opStack.length == 0 && !requestedRunStep) {
+          logger.enable();
+        }
+
         try {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           resolve(await userFnToRun(fnArg as Context<any, any, any>));
         } catch (err) {
           reject(err);
+          logger.error(err);
         }
+        logger.disable();
       });
 
       let pos = -1;
 
       do {
         if (pos >= 0) {
+          if (!requestedRunStep && pos == opStack.length - 1) {
+            logger.enable();
+          }
+
           state.tickOps = {};
           const incomingOp = opStack[pos] as IncomingOp;
           state.currentOp = state.allFoundOps[incomingOp.id];
@@ -348,7 +358,9 @@ export class InngestFunction<
              * undefined state.
              */
             throw new NonRetriableError(
-              functionStoppedRunningErr(ErrCode.ASYNC_DETECTED_DURING_MEMOIZATION)
+              functionStoppedRunningErr(
+                ErrCode.ASYNC_DETECTED_DURING_MEMOIZATION
+              )
             );
           }
 
@@ -356,7 +368,6 @@ export class InngestFunction<
 
           if (typeof incomingOp.data !== "undefined") {
             state.currentOp.resolve(incomingOp.data);
-            logger.reset();
           } else {
             state.currentOp.reject(incomingOp.error);
           }
@@ -364,6 +375,7 @@ export class InngestFunction<
 
         await timer.wrap("memoizing-ticks", resolveAfterPending);
 
+        logger.disable();
         state.reset();
         pos++;
       } while (pos < opStack.length);
@@ -394,12 +406,14 @@ export class InngestFunction<
 
         const runningStepStop = timer.start("running-step");
         state.executingStep = true;
+        logger.enable();
 
         const result = await new Promise((resolve) => {
           return resolve(userFnToRun());
         })
           .finally(() => {
             state.executingStep = false;
+            runningStepStop();
           })
           .then((data) => {
             return {
@@ -431,9 +445,6 @@ export class InngestFunction<
                 error: err,
               };
             }
-          })
-          .finally(() => {
-            runningStepStop();
           });
 
         return [
@@ -471,9 +482,11 @@ export class InngestFunction<
            * complete, so we should only do this if we're sure that all
            * registered ops have been resolved.
            */
-          const allOpsFulfilled = Object.values(state.allFoundOps).every((op) => {
-            return op.fulfilled;
-          });
+          const allOpsFulfilled = Object.values(state.allFoundOps).every(
+            (op) => {
+              return op.fulfilled;
+            }
+          );
 
           if (allOpsFulfilled) {
             return ["complete", fnRet.data];
@@ -532,7 +545,9 @@ export class InngestFunction<
 
           if (!hasOpsPending) {
             throw new NonRetriableError(
-              functionStoppedRunningErr(ErrCode.ASYNC_DETECTED_AFTER_MEMOIZATION)
+              functionStoppedRunningErr(
+                ErrCode.ASYNC_DETECTED_AFTER_MEMOIZATION
+              )
             );
           }
         }
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -282,8 +282,6 @@ export class InngestFunction<
        *
        * We only use the onFailure handler if
        */
-
-      await logger.flush();
       if (!this.#onFailureFn) {
         // TODO PrettyError
         throw new Error(
--- a/src/components/InngestFunction.ts
+++ b/src/components/InngestFunction.ts
@@ -8,6 +8,7 @@ import {
 import { resolveAfterPending, resolveNextTick } from "../helpers/promises";
 import { type ServerTiming } from "../helpers/ServerTiming";
 import { slugify, timeStr } from "../helpers/strings";
+import { type Logger, ProxyLogger } from "../middleware/logger";
 import {
   StepOpCode,
   type Context,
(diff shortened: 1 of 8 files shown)


jpwilliams (commented):
> Due to the actual deliveries of logs being all run in a Promise, the log delivery will likely be out of order.

I think this is okay now, as we expect the underlying logger to be synchronous? Can likely remove the use of `async` for calling `this._logger[method](...args)`.

> Even with the current implementation, there could be a minor impact on function execution.
Something to address in the future is to make the 1s wait only run on serverless environments.

Yeah, the 1 second wait is grim. I know we discussed this - even with middleware, flushing logs is not a standardized API across these different logging libraries, so it'd still be guesswork for us to know what code to run to try and flush a particular library.

Seeing as we're adding first-class support for this, I wonder if it's reasonable to give users a way to define a "flush" function of their own alongside defining their logger. e.g.
```ts
new Inngest({
  logger: {
    client: myWinstonLogger,
    flush: () => { ... },
  },
});
```
This would also give us space to add other log-specific options in the future if we needed them. 🤔

> When ran as part of an application server, the whole flush method is very likely to be irrelevant completely.

Yeah, agreed. Potentially even more reason for us to not abstract the flushing away and give users control over it.

---

I do have a worry about this method and the ordering of logs. With the default console logger and flushing, I think this example would log `foo, 1, 2` instead of the expected `1, foo, 2`?
```ts
async ({ logger }) => {
  logger.info(1);
  logFoo();
  logger.info(2);
}
```
It might be appropriate to start tracking the `phase` in our local `state` and have the logger use that to understand whether it should log or not instead of using a buffer. We can chat about this!

src/middleware/logger.ts
```diff
@@ -39,3 +39,94 @@ export class DefaultLogger implements Logger {
...
+ *
+ * The expected usage of this class is,
+ * 1. store the log attempt with its arguments and level
+ * 2. if function hits a step that has been memorized, clear the _buffer
+ * 3. on function return, flush all buffers
+ *
+ * And it should be invisible to the user as much as possible.
+ */
+export class ProxyLogger implements Logger {
+  private readonly _logger: Logger;
```
goodoldneon, line 70: Should we use real private fields instead of the semantic `_` prefix?

```suggestion
  readonly #logger: Logger;
```

src/middleware/logger.ts
```diff
@@ -39,3 +39,94 @@ export class DefaultLogger implements Logger {
...
+    if (this.bufSize() === 0) return;
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    const deliveries = this._buffer.map(async (log) => {
+      const args = log.args;
+      const method = log["level"] as keyof Logger;
+      return this._logger[method](...args);
+    });
+    // NOTE: timestamp will likely not be linear as expected.
+    await Promise.all(deliveries);
```
goodoldneon, line 107: `Promise.all` will immediately reject on the first rejected promise in `deliveries`. Should we use `Promise.allSettled` instead?

src/middleware/logger.ts
```diff
@@ -39,3 +39,94 @@ export class DefaultLogger implements Logger {
...
+    //
+    // TODO:
+    // This should only happen when using a serverless environment because it's very
+    // costly from the compute perspective.
+    // server runtimes should just let the logger do their thing since most of them
+    // should have already figured what to do in those environments, be it threading or
+    // something else.
+    if (this._logger.constructor.name !== DefaultLogger.name) {
+      await new Promise((resolve) => {
+        setTimeout(() => resolve(null), 1000);
```
goodoldneon, line 121: I'm confused by this. What are we waiting 1 second for? Haven't we already awaited the deliveries?

src/middleware/logger.ts
darwin67: Learned something new again! :D
Will update

src/middleware/logger.ts
darwin67: Because based on how the user setup their logger, the implementation details within the provided logger could have timeouts or sleep, when they actually handle the delivery.
Just calling `_logger.info` doesn't actually mean the provided logger will attempt delivery right away.

Some examples,

*Pino*
https://github.com/pinojs/pino/blob/master/docs/transports.md#asynchronous-startup
https://github.com/pinojs/pino/blob/171d8056a2a66c42e4bd2e13b5936b6a652af830/lib/transport.js#L53

*Winston*
https://github.com/winstonjs/winston/blob/f8320c1e3bb7cc3544b1385da92cc625e2b3b0ac/lib/winston/transports/http.js#L199-L203

It's not consistent, but 1s is a guess that could balance enough time to flush vs hit to throughput.

darwin67:
@jpwilliams 

> Seeing as we're adding first-class support for this, I wonder if it's reasonable to give users a way to define a "flush" function of their own alongside defining their logger. e.g.

That sounds reasonable, but I won't include it in this change. I'm not planning to announce this change to users right away, and instead want to start dogfooding ourselves first to see how well the 1s works now, and if we will miss delivery or not.

> It might be appropriate to start tracking the phase in our local state and have the logger use that to understand whether it should log or not instead of using a buffer. We can chat about this!

I'd like to understand what problem you're trying to solve. Debugging in local development?
In prod, logging is likely happening at multiple places at the same time, so logs getting all mixed up is not uncommon when you're looking at a logging system provider, and the order of the logs usually doesn't matter anymore because of the chaos.

I can see that it's mentally uncomfortable when logs don't show up in the expected order, but I guess do you ever log something in prod in a way that's supposed to be looked at sequentially?

src/components/InngestFunction.ts
```diff
@@ -343,6 +348,7 @@ export class InngestFunction<
            * Whichever the case, this is bad and we can't continue in this
            * undefined state.
            */
+          await logger.flush();
```
jpwilliams, line 351: It looks like every call to `logger.flush()` happens just before the function ends, either from a `throw` or a `return`.

To keep the code paths low (and to make sure we don't miss any in the future), it might be possible to wrap everything in a `try`/`finally` and run `await logger.flush()` in the `finally` block.

```ts
try {
  // everything
  // starting at userFnPromise
  // ending at the end of the fn
} finally {
  await logger.flush();
}
```

Feels a bit grim indenting so many lines, but I'm happy to do so for now to make sure all the code paths are guaranteed to be captured and it's a good signal we need to refactor `runFn` a touch when we can anyway!

src/middleware/logger.ts
```diff
@@ -39,3 +39,94 @@ export class DefaultLogger implements Logger {
...
+    if (this.bufSize() === 0) return;
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    const deliveries = this.#buffer.map(async (log) => {
+      const args = log.args;
+      const level = log["level"];
+      return this.#logger[level](...args);
+    });
+    // NOTE: timestamp will likely not be linear as expected.
+    await Promise.allSettled(deliveries);
```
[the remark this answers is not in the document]
darwin67: those `info` should always be sync.

src/components/InngestFunction.ts
```diff
@@ -279,6 +283,7 @@ export class InngestFunction<
        * We only use the onFailure handler if
        */
 
+      await logger.flush();
```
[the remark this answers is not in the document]
darwin67: good catch. doesn't seem like it's needed since tests passed without it.

src/components/InngestFunction.ts
darwin67: sure, I'll wrap the `try` after logger got instantiated then

src/middleware/logger.ts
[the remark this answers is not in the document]
darwin67: This is no longer relevant.

src/components/InngestFunction.ts
```diff
@@ -251,288 +252,310 @@ export class InngestFunction<
...
-      }
-    });
+        try {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          resolve(await userFnToRun(fnArg as Context<any, any, any>));
+        } catch (err) {
+          reject(err);
+          logger.error(err);
+        }
+        logger.disable();
```
jpwilliams, line 334: Did you find these were still needed?

jpwilliams (commented):
Couple of nits and then we should be good to ship! Awesome! 🥳

src/components/InngestFunction.ts
darwin67: I found no differences when it comes to output, but decided to keep it for people reading it later.
It's going to be confusing regardless but only seeing `logger.enable` and it never gets disabled seems like it'll pretty hard to comprehend.

basically when something was opened, let's just close it just in case.

jpwilliams (approved):

PR #195, opened 2023-05-13
Return event IDs when sending events

jpwilliams:
## Summary

When events are sent to Inngest, return the event IDs from the response.

### DIFF
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -1,3 +1,4 @@
+import { z } from "zod";
 import { envKeys } from "../helpers/consts";
 import { devServerAvailable, devServerUrl } from "../helpers/devserver";
 import {
@@ -148,7 +149,7 @@ export class Inngest<
   /**
    * Given a response from Inngest, relay the error to the caller.
    */
-  async #getResponseError(response: globalThis.Response): Promise<Error> {
+  #getResponseError(response: globalThis.Response, rawBody: unknown): Error {
     let errorMessage = "Unknown error";
     switch (response.status) {
       case 401:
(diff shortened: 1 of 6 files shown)


src/components/Inngest.ts
```diff
@@ -339,11 +342,22 @@ export class Inngest<
...
+    try {
+      rawBody = await response.json();
 
-    throw await this.#getResponseError(response);
+      const body = await z
+        .object({
+          ids: z.array(z.string()),
+          status: z.number().min(200).max(299),
+        })
+        .parseAsync(rawBody);
```
djfarrelly, line 355: If we decide at some point to return more data from the response and the shape of the response changes, would this parsing fail? Or will it only parse the data expected and ignore the rest?

djfarrelly (approved):
Hey @jpwilliams - this looks great! The only question that I have on this one is whether we want the return value to be just the event IDs, or is there a possibility in the future to add more data to the response in which case we may want to return an object? e.g.

```ts
const { ids } = await inngest.send([...])
```

I don't currently have any other ideas, but before shipping, I might challenge us to ask the team to think if there are any other ideas in which we might want to pass them through to the client.

djfarrelly (commented):
On second review after reviewing the middleware docs @jpwilliams, I'm kind of leaning towards returning the entire response here instead of just the string. We may return other metadata and just returning a string array limits us in the future instead of just returning an object.

djfarrelly (changes requested):
After further thought - we definitely should return an object for future extensibility. I'll make that change on this branch after rebasing from main.

jpwilliams:
Closing for #256.

PR #200, opened 2023-05-20
Add platform deploy checks

tonyhb:
This extends isProd, checking if we're deployed to a known platform.  We use this when sending events or checking signing keys, ensuring that even if you deploy to eg. Vercel with NODE_ENV=development we still use production mode.


goodoldneon:
`isProd` is specifically checking whether it's a production environment but `isPlatformDeploy` is checking whether it's deployed to a provider _at all_. That doesn't seem safe to swap.

Also, `isPlatformDeploy` only works for 4 specific providers but `isProd` can work for any provider (since it uses `NODE_ENV`)

goodoldneon:
Oh wait, I'm dumb and didn't see you concatted

tonyhb:
I think we're debating about names here;  we're specifically checking whether to use the dev server, so if we change `isPlatformDeploy` to `skipDevServer` then it might sound safer.  This is a necessary addition to prevent local checks on platforms.

src/helpers/env.ts
```diff
@@ -62,6 +62,38 @@ const prodChecks: [
...
+  key: string,
+  customCheck: keyof typeof prodCheckFns,
+  value?: string
+][] = prodChecks.concat([
+  // Extend prod checks, then check if we're deployed to a platform.
+  [envKeys.IsRender, "is truthy"],
+  [envKeys.IsVercel, "is truthy"],
+  [envKeys.IsNetlify, "is truthy"],
+  [envKeys.RailwayBranch, "is truthy"],
+]);
```
jpwilliams, line 79: Fun to see this monstrosity reused. 😅

src/components/InngestCommHandler.ts
```diff
@@ -636,7 +636,7 @@ export class InngestCommHandler<
         },
       });
 
-    this._isProd = actions.isProduction ?? isProd(env);
+    this._isProd = actions.isProduction ?? isPlatformDeploy(env);
```
jpwilliams, line 639: As discussed elsewhere (cc @goodoldneon), keeping the `isProd()` check here would be ace!
```suggestion
    this._isProd = actions.isProduction ?? isProd(env) || isPlatformDeploy(env);
```

src/components/Inngest.ts
```diff
@@ -328,7 +328,7 @@ export class Inngest<
     // dev server.
     let url = this.inngestApiUrl.href;
 
-    if (!isProd()) {
+    if (!isPlatformDeploy()) {
```
jpwilliams, line 331: As discussed elsewhere (cc @goodoldneon), keeping `isProd()` here (or adding a new `useDevServer()` function) would be good.

src/components/InngestCommHandler.ts
```diff
@@ -637,6 +645,8 @@ export class InngestCommHandler<
       });
 
     this._isProd = actions.isProduction ?? isProd(env);
+    // If we're in production always skip.
+    this._skipDevServer = actions.isProduction ?? skipDevServer(env);
```
jpwilliams, line 649: Should this be checking the above `this._isProd` instead so that we get the value of `isProd(env)`?

```suggestion
    this._skipDevServer = this._isProd ?? skipDevServer(env);
```

src/helpers/env.ts
```diff
@@ -35,7 +35,7 @@ export const devServerHost = (): string | undefined => {
   return values.find((a) => !!a);
 };
 
-const prodCheckFns = (<
+const checkFns = (<
```
jpwilliams, line 38: Good call.

src/helpers/env.ts
```diff
@@ -62,6 +62,57 @@ const prodChecks: [
...
+    return checkFns[checkKey](stringifyUnknown(env[key]), expected);
+  });
+};
+
+
+/**
+ * Returns `true` if we believe the current environment is a platform, eg>
+ * Vercel, Render, Railway, Netlify, etc.
+ */
+export const isPlatformDeploy = (
```
jpwilliams, line 104: Can this entire function be removed now? Doesn't look like it's being used anymore.

jpwilliams (commented):
This is much nicer to read - `skipDevServer()` was a good call.

Just a couple of nits and you're good to merge.

src/helpers/env.ts
```diff
@@ -62,6 +62,57 @@ const prodChecks: [
...
+][] = [
+  // Extend prod checks, then check if we're deployed to a platform.
+  [envKeys.IsVercel, "is truthy"],
+  [envKeys.IsNetlify, "is truthy"],
+  [envKeys.IsRender, "is truthy"],
+  [envKeys.RailwayBranch, "is truthy"],
+  [envKeys.IsCloudflarePages, "is truthy"],
+];
+
+const skipDevServerCHecks = prodChecks.concat(platformDeployChecks);
```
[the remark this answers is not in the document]
tonyhb: Thanks!

src/helpers/env.ts
tonyhb: It can be, though given we have the checks for it I figured we should expose them separately?  Dont mind either way!

jpwilliams (approved):

PR #204, opened 2023-05-30
INN-1418: Provide additional metadata with child loggers if applicable

darwin67:
## Summary
Some of the commonly used loggers provide a `child` implementation, which is a syntactic sugar to simplify adding additional metadata for logging.

If the user provided logger has such implementation, we will automatically add certain runtime metadata when a user uses the default logger middlware.

### DIFF
--- a/src/components/Inngest.ts
+++ b/src/components/Inngest.ts
@@ -498,13 +498,14 @@ const builtInMiddleware = (<T extends MiddlewareStack>(m: T): T => m)([
             functionName: arg.fn.name,
           };
 
+          /* eslint-disable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/ban-ts-comment */
           // create a child logger if the provided logger has child logger implementation
-          /* eslint-disable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment */
           const providedLogger: Logger =
             "child" in client["logger"]
+              // @ts-ignore-TS2571
               ? client["logger"].child(metadata)
               : client["logger"];
-          /* eslint-enable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment */
+          /* eslint-enable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/ban-ts-comment */
           const logger = new ProxyLogger(providedLogger);
 
           return {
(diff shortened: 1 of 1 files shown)


src/components/Inngest.ts
```diff
@@ -490,8 +490,22 @@ const builtInMiddleware = (<T extends MiddlewareStack>(m: T): T => m)([
...
+            functionName: arg.fn.name,
+          };
+
+          // create a child logger if the provided logger has child logger implementation
+          /* eslint-disable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment */
+          const providedLogger: Logger =
+            "child" in client["logger"]
+              ? client["logger"].child(metadata)
+              : client["logger"];
+          /* eslint-enable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment */
```
jpwilliams, line 507: Is it reasonable to wrap this in a `try`/`catch` just in case we meet a library that dislikes this signature?
```suggestion
          let providedLogger: Logger = client["logger"];

          try {
            // create a child logger if the provided logger has child logger implementation
            if ("child" in providedLogger) {
              providedLogger = (
                providedLogger.child as (
                  metadata: Record<string, unknown>
                ) => Logger
              )(metadata);
            }
          } catch (err) {
            // no-op
          }

```

jpwilliams (commented):
Awesome! I think we can bubble-wrap this for some extra safety and we're good to go! 🔥

We could also do some runtime checks to make sure the resulting logger also has our contracted methods (`.info()`, etc.), but that might be a bit excessive for the moment.

src/components/Inngest.ts
darwin67: sure, we can add it, but would like it to be something more readable instead of mixing both type declaration with func execution.

src/components/Inngest.ts
```diff
@@ -490,8 +490,31 @@ const builtInMiddleware = (<T extends MiddlewareStack>(m: T): T => m)([
...
+
+          let providedLogger: Logger = client["logger"];
+          /* eslint-disable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/ban-ts-comment */
+          // create a child logger if the provided logger has child logger implementation
+          try {
+            if ("child" in client["logger"]) {
+              type ChildLoggerFn = (
+                metadata: Record<string, unknown>
+              ) => Logger;
+              // @ts-ignore-2339: ignore child method is not defined error since it's already checked above
```
jpwilliams, line 509: Nit - I think you can remove all of these `@ts-ignore` and `eslint-disable` comments if you check `if ("child" in providedLogger)` instead of `client["logger"]`.

jpwilliams (approved):

PR #207, opened 2023-05-31
Handle circular JSON errors while stringifying across the SDK

jpwilliams:
## Summary

Use [json-stringify-safe](https://www.npmjs.com/package/json-stringify-safe) instead of `JSON.stringify` to serialize JSON to ensure that circular errors are correctly handled.


goodoldneon:
`json-stringify-safe` is only 27 lines. Should we copy-paste it instead of adding a dependency? Their license permits that as long as we credit the authors in the file

goodoldneon:
No harm! I just get feel good vibes with fewer dependencies 😎

PR #216, opened 2023-06-06
INN-1403: Read events for batch handling

darwin67:
Requires https://github.com/inngest/monorepo/pull/1396 to be released first.

## Summary

Changing the parsing of data from SDK to use `events` instead of `event`.
Over time, as SDK version gets updated, we can stop sending `event` data over to the SDK completely.

For backwards compatibility, `event` should always be the first item of `events`.
And if it's a single item batch, `[event] === events` should always be true.

### Side note
Sneaking in some nix-shell configurations for replicating local development setups.

### Releases
Can be released whenever with the next bug fix or minor update. Don't need to be an independent release of its own.


darwin67:
@jpwilliams 
I don't see any conventions for making API calls to Inngest right now so I added a new API class as a handler and attached it to the client.

I dump an ignore here now to verify that it's functioning but the type signature here is something I'm not clear how to handle with `cross-fetch`.
https://github.com/inngest/inngest-js/pull/216/files#diff-5087958ec6f8b087e8827c51631a00210094a1d599c750e7ccade1b12ffc59c5R842-R845

I can modify the API response interface but it doesn't solve the fundamental issue of response objects being completely different.

Also, another thing, I'm not sure what we want to do if the calls here failed. Abort early?

src/api/api.ts
```diff
@@ -0,0 +1,64 @@
...
+  ): Promise<Result<StepsResponse, ErrorResponse>> {
+    const url = `${this.baseUrl}/v0/runs/${runId}/actions`;
+
+    return fetch(url, {
+      headers: { Authorization: `Bearer ${this.eventKey}` },
+    }).then(async (resp) => {
+      const data: unknown = await resp.json();
+
+      if (resp.ok) {
+        return Ok(StepsSchema.parse(data));
```
tonyhb, line 40: LOL i love this

src/types.ts
```diff
@@ -919,6 +939,10 @@ export interface FunctionConfig {
     }
   >;
   idempotency?: string;
+  batchEvents?: {
+    maxSize: number;
+    timeout: string;
```
tonyhb, line 944: 🤔  should we make a `SecondTimeStr` type here?

darwin67: chatted in person and decided this can be a follow up.
backend will verify on registration anyways.

src/types.ts
```diff
@@ -988,3 +1012,68 @@ export interface StepOpts {
...
+export const Err = <E>(error?: E): Result<never, E> => {
+  return { ok: false, error };
+};
+
+/**
+ * Format of data send from the executor to the SDK
+ */
+export const fnDataSchema = z.object({
+  event: z.object({}).passthrough(),
+  events: z.array(z.object({}).passthrough()).default([]),
```
jpwilliams, line 1039: Should the default of `events` be an empty array? Do we not always expect it to be populated?

jpwilliams: Noting - as a follow-up, it'd be great to not send the event twice in the payload, i.e. `event` and `events[0]`. I think we talked about this before and discussed that the executor can look at the SDK version / feature set and decide what payload shape to send. Doesn't have to happen now 🙂

src/api/api.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private get hashedKey(): string {
+    return hashSigningKey(this.signingKey);
+  }
+
+  async getRunSteps(
+    runId: string
+  ): Promise<Result<StepsResponse, ErrorResponse>> {
+    const url = `${this.baseUrl}/v0/runs/${runId}/actions`;
```
jpwilliams, line 37: Nit: Can we make this a URL object instead of a string? Just helps with silly errors like double slashes.
```suggestion
    const url = new URL(`/v0/runs/${runId}/actions`, this.baseUrl);
```

src/api/api.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    return fetch(url, {
+      headers: { Authorization: `Bearer ${this.hashedKey}` },
+    }).then(async (resp) => {
+      const data: unknown = await resp.json();
```
jpwilliams, line 42: If `fetch()`, `resp.json()`, or `ErrorSchema.parse(data)` throws here, the function will throw instead of returning an `ErrorResponse`. Is this okay?

If we want to always return an `ErrorResponse`, it might be reasonable here (and in `getRunBatch()`) to wrap the entire call in a `try`/`catch`.

Alternatively you could just add a `.catch()` to the end of the `fetch()` call, but you won't have access to the `data` to parse it.

src/api/api.ts
```diff
@@ -0,0 +1,69 @@
...
+      } else {
+        return Err(ErrorSchema.parse(data));
+      }
+    });
+  }
+
+  async getRunBatch(
+    runId: string
+  ): Promise<Result<BatchResponse, ErrorResponse>> {
+    const url = `${this.baseUrl}/v0/runs/${runId}/batch`;
```
jpwilliams, line 55: Nit: Can we make this a URL object instead of a string? Just helps with silly errors like double slashes.
```suggestion
    const url = new URL(`/v0/runs/${runId}/batch`, this.baseUrl);
```

src/api/api.ts
```diff
@@ -0,0 +1,69 @@
...
+  BatchSchema,
+  type BatchResponse,
+} from "./schema";
+
+interface InngestAPIConstructorOpts {
+  baseUrl?: string;
+  signingKey: string;
+}
+
+export class InngestAPI {
```
jpwilliams, line 18: The biggest of all the nits: Can this (and other related types) be `InngestApi` for consistency?

src/components/Inngest.ts
```diff
@@ -163,6 +166,13 @@ export class Inngest<TOpts extends ClientOptions = ClientOptions> {
       inngestEnv: env,
     });
 
+    const signingKey = processEnv(envKeys.SigningKey) || "";
```
jpwilliams, line 169: This is tricky. Some environments don't give you access to environment variables upon instantiation and instead pass them _alongside a request_, e.g. Cloudflare.

For this reason, `InngestCommHandler` attempts to pull these [as late as possible](https://github.com/inngest/inngest-js/blob/main/src/components/InngestCommHandler.ts#L646).

To support these types of environments, it might be reasonable for `upsertSigningKeyFromEnv` to _also_ update `this.client["inngestApi"]` (maybe via a `this.client["inngestApi"].setSigningKey()` method) with the new key as well as setting it for itself.

src/components/Inngest.ts
```diff
@@ -89,6 +90,8 @@ export class Inngest<TOpts extends ClientOptions = ClientOptions> {
    */
   public readonly inngestBaseUrl: URL;
 
+  public readonly inngestAPI: InngestAPI;
```
jpwilliams, line 93: In the spirit of hiding this from users when they type `inngest.` in their editor, can we make this `private` and refer to it using brackets elsewhere? e.g. `inngest["inngestApi"]` instead of `inngest.inngestApi`

src/helpers/functions.ts
```diff
@@ -58,3 +61,54 @@ export const waterfall = <TFns extends ((arg?: any) => any)[]>(
...
+      ]);
+
+      if (evtResp.ok) {
+        result.events = evtResp.value;
+      } else {
+        return Err(
+          prettyError({
+            whatHappened: "failed to retrieve list of events",
+            consequences: "function execution can't continue",
+            stack: true,
```
jpwilliams, line 86: For the `prettyError()` calls in this function, it'd be great to be able to give users some more information about how to proceed. In this case, function execution can't continue, but what's my course of action? Was it my fault? Have my events been lost?

There are a few extra properties we can use to further describe the problem and how to fix it:

```ts
/**
 * If applicable, provide a full sentence to reassure the user about certain
 * details, for example if an error occurred whilst uploading a file, but we
 * can assure the user that uploading succeeded and something internal failed.
 */
reassurance?: string;

/**
 * Tell the user why the error happened if we can. This should be a full
 * sentence or paragraph that explains the error in more detail, for example
 * to explain that a file failed to upload because it was too large and that
 * the maximum size is 10MB.
 */
why?: string;

/**
 * If we can, tell the user what they can do to fix the error now. This should
 * be a full sentence or paragraph that explains what the user can do to fix
 * the error, for example to tell them to try uploading a smaller file or
 * upgrade to a paid plan.
 */
toFixNow?: string | string[];

/**
 * If applicable, tell the user what to do if the error persists, they want
 * more information, or the fix we've given them doesn't work.
 *
 * This should be a full sentence or paragraph, and will likely refer users
 * to contact us for support, join our Discord, or read documentation.
 */
otherwise?: string;

/**
 * If applicable, provide a code that the user can use to reference the error
 * when contacting support.
 */
code?: ErrCode;
```

A new `code` for each unique error would be great for allowing us to later provide docs pages and shortlinks for each of these.

It may be that we can't specify all of these properties or be all that helpful for a user right at this second; filling out these fields is pretty difficult, thinking about what the user should actually do in this situation. The _least_ we can do is provide reassurance and ask them to reach out on Discord.

src/helpers/functions.ts
```diff
@@ -58,3 +61,54 @@ export const waterfall = <TFns extends ((arg?: any) => any)[]>(
...
+  data: unknown,
+  api: InngestAPI
+): Promise<Result<FnData, ParseErr>> => {
+  try {
+    const result = fnDataSchema.parse(data);
+
+    if (result.use_api) {
+      const [evtResp, stepResp] = await Promise.all([
+        api.getRunBatch(result.ctx?.run_id as string),
+        api.getRunSteps(result.ctx?.run_id as string),
```
jpwilliams, line 76: Is it worth checking `result.ctx.run_id` at the start of the function and returning an error if we somehow have an unidentifiable run?

src/helpers/functions.test.ts
```diff
@@ -0,0 +1,64 @@
+import { type EventPayload } from "../types";
+import { parseFnData } from "./functions";
+import { InngestAPI } from "../api/api";
```
jpwilliams, line 3: In test files, these imports need to be prefixed with `@local`. This is a cheeky trick used so we can test the built types against multiple consuming versions of TypeScript.

If you don't do this, type tests will silently always use the library's version of TS (currently 5), meaning we're not actually testing for 4.7+ support.

This is a major footgun that needs a custom ESLint rule adding to `rules/`.

```suggestion
import { InngestAPI } from "@local/api/api";
import { parseFnData } from "@local/helpers/functions";
import { type EventPayload } from "@local/types";
```

jpwilliams (changes requested):
`Inngest.test.ts` has some type tests that assert some of the input into an Inngest function is correctly typed.

It'd be awesome to also include checking this new `events` key for each of those tests to make sure it's exactly what we expect it to be.

src/types.ts
```diff
@@ -646,6 +648,24 @@ export interface FunctionOptions<
...
 
+  /**
+   * batchEvents specifies the batch configuration on when this function
+   * should be invoked when one of the requirements are fulfilled.
+   */
+  batchEvents?: {
+    /**
+     * The maximum number of events to be consumed in one batch
+     */
+    maxSize: number;
```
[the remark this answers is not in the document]
darwin67: I get what you're saying but it's more often referred as `batch size` instead of `batch limit`. I'd prioritized natural language more than consistency for this one.

src/types.ts
```diff
@@ -646,6 +648,24 @@ export interface FunctionOptions<
...
+
+    /**
+     * How long to wait before invoking the function with a list of events.
+     * If timeout is reached, the function will be invoked with a batch
+     * even if it's not filled up to `maxSize`.
+     */
+    timeout: string;
```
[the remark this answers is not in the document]
darwin67: currently 1s - 60s is the expected value.
could be a follow up if it gets a little complicated to deal with.

src/test/helpers.ts
```diff
@@ -706,7 +710,7 @@ export const testFramework = (
                 method: "POST",
                 headers: {
                   [headerKeys.Signature]:
-                    "t=1674082860&s=88b6453463050d1846743cbba0925bae7c1cf807f9c74bbd41b3d5cfc9c70d11",
+                    "t=1687306735&s=70312c7815f611a4aa0b6f985910a85a6c232c845838d7f49f1d05fd8b2b0779",
```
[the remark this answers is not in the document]
darwin67: I'll share this with you separately

src/api/api.ts
darwin67: I'll do a catch at the end. if it fails, data won't be available anyways.

src/helpers/functions.ts
darwin67: there's really nothing the user can do since none of these errors will be within a user's control. it's best for us to be notified automatically somehow when these errors happened so we can look into it on our end.

src/helpers/consts.ts
```diff
@@ -21,6 +21,7 @@ export enum envKeys {
   DevServerUrl = "INNGEST_DEVSERVER_URL",
   Environment = "INNGEST_ENV",
   BranchName = "BRANCH_NAME",
+  InngestAPIBaseUrl = "INNGEST_INNGEST_API_BASE_URL",
```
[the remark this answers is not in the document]
darwin67: I was thinking there might be a separate `API_BASED_URL` but probably not. will change.

djfarrelly (commented):
Hey @darwin67 - I checked out the PR and there are some great refactors and I like the additions! I left some comments mostly around style. I do not see any functional issues with this code - that all looks solid. Long term we'll have to pay attention to SDK versions and the Inngest API versions which will be subject to change. I think with this change in the code, we may want to prioritize getting some of these endpoints to `v1` for long term support.

---

I see some Rust-style changes here and I don't mind them, but at a high level, I'm thinking that we will need to work on increasing consistency about how errors are handled in this repo. We have pretty errors (strings), custom error classes, serialized errors, some `console.*` of errors and some returning of errors back to the executor - _and now_ the Rust-style `Result` & `Err`. I would say that these notes are well outside of the scope of this PR, but it's something we should make a plan for as a team and improve over time. (Personal opinion: I kind of like the Rust-style result).

src/helpers/functions.ts
```diff
@@ -58,3 +61,71 @@ export const waterfall = <TFns extends ((arg?: any) => any)[]>(
...
+          })
+        );
+      }
+    }
+
+    return Ok(result);
+  } catch (err) {
+    // print it out for now.
+    // move to something like protobuf so we don't have to deal with this
+    console.error(err);
```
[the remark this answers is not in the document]
darwin67: I thought about it but zod errors are weirdly structured and it's cryptic enough that it's not very useful.

Besides, the data's structure and exchange happening here is between the SDK and API, meaning the user can't do anything even if there's a parsing error.

djfarrelly (approved):

tonyhb (approved):

src/types.ts
```diff
@@ -646,6 +653,27 @@ export interface FunctionOptions<
...
+    maxSize: number;
+
+    /**
+     * How long to wait before invoking the function with a list of events.
+     * If timeout is reached, the function will be invoked with a batch
+     * even if it's not filled up to `maxSize`.
+     *
+     * Expects 1s to 60s.
+     */
+    timeout: TimeStrBatch;
```
jpwilliams, line 674: We'll standardize this with Temporal durations in the future.

src/components/InngestCommHandler.ts
```diff
@@ -645,6 +637,7 @@ export class InngestCommHandler<
       if (runRes) {
         this.upsertSigningKeyFromEnv(env);
         this.validateSignature(runRes.signature, runRes.data);
+        this.client["inngestApi"].setSigningKey(this.signingKey);
```
jpwilliams, line 640: Nice.

jpwilliams (approved):
Nice!

PR #218, opened 2023-06-07
Fix output hooks not running when main function body throws

jpwilliams:
## Summary

Output hooks were not running if an asynchronous, non-step function's body threw. We expect to handle all errors, so the output hook should be appropriately reporting the problem here.

This also ensures that the built-in logger middleware logs any error from a function's execution, as is expected.

### DIFF
--- a/.changeset/flat-fans-nail.md
+++ b/.changeset/flat-fans-nail.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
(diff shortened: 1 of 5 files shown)


tonyhb (approved):
good catch!

jpwilliams:
Blocked by #221 to ensure integration tests are fine, then will merge.

PR #220, opened 2023-06-09
Fix creating a new comparison date when hashing step sleep IDs

jpwilliams:
## Summary

Fixes a bug where we use two different `Date` objects to represent "now," breaking `step.sleep()` hashing in _very rare_ circumstances.

Adds some tests that try to give the problem an opportunity to appear; it's hard to cause the problem to appear due to it being temporally reliant.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -927,7 +927,7 @@ export interface FunctionConfig {
   cancel?: {
     event: string;
     if?: string;
-    timeout?: TimeStr;
+    timeout?: string;
   }[];
 }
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -927,7 +927,7 @@ export interface FunctionConfig {
   cancel?: {
     event: string;
     if?: string;
-    timeout?: TimeStr;
+    timeout?: string;
```
jpwilliams, line 930: This is internal config; not the config exposed to the user.

tonyhb (approved):
LGTM, though def would feel a bit more at ease if this had a unit test or two :D

jpwilliams:
Blocked by https://github.com/inngest/inngest-js/pull/221 to ensure integration tests are fine, then will merge.

PR #231, opened 2023-06-12
Add more info to 405s and ensure rejected GET is a 403

jpwilliams:
## Summary

Add more information when responding with a failure from the SDK.

- Rejected `GET` now returns `403 Forbidden` instead of `405 Method Not Allowed`
- Rejected `GET` now returns `isProd`, `skipDevServer`, and `showLandingPage` values
- No actions being found remains a `405 Method Not Allowed` response
- No actions being found now returns `isProd` and `skipDevServer` values

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -689,8 +689,13 @@ export class InngestCommHandler<
 
         if (this._isProd || !showLandingPage) {
           return {
-            status: 405,
-            body: "",
+            status: 403,
+            body: JSON.stringify({
+              message: "Landing page requested but is disabled",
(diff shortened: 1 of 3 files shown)


src/components/InngestCommHandler.ts
```diff
@@ -756,7 +761,11 @@ export class InngestCommHandler<
 
     return {
       status: 405,
-      body: "",
+      body: JSON.stringify({
+        message: "No action found; request was likely not POST, PUT, or GET",
```
djfarrelly, line 765: What does this error mean? Aren't these http methods the only way to invoke the endpoint? If it's not one of these, should we log more diagnostic data?

PR #240, opened 2023-06-13
Harden error serialization

jpwilliams:
## Summary

There are some edge cases around error serialization that causes some non-useful errors to bubble up to users.

Here we aim to cover more of these edge cases to return first-party errors when possible.

### DIFF
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -1,27 +1,45 @@
 import chalk from "chalk";
+import stringify from "json-stringify-safe";
 import {
   deserializeError as cjsDeserializeError,
   serializeError as cjsSerializeError,
+  errorConstructors,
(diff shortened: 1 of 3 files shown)


src/helpers/errors.ts
```diff
@@ -32,29 +50,100 @@ export interface SerializedError {
...
+      };
+    } catch (err) {
+      // If this failed, then stringifying the object also failed, so we'll just
+      // return a completely generic error.
+      // Failing to stringify the object is very unlikely.
+      return {
+        name: "Could not serialize source error",
+        message: "Serializing the source error failed.",
+        stack: "",
+        [SERIALIZED_KEY]: SERIALIZED_VALUE,
```
tonyhb, line 103: Should we add the original subject arg here?

PR #250, opened 2023-06-23
Include 'modified' in register response

goodoldneon:
# Changes

Include `modified` in register response.

# Purpose

The API will be able to forward the `modified` value to the UI during registration. This will give the UI enough information to know whether a deploy was created or deduped.

# Testing

I manually tested this with an API feature branch and the `modified` value in the response is `true` if a deploy was created and `false` if the deploy was deduped.

### DIFF
--- a/src/components/InngestCommHandler.ts
+++ b/src/components/InngestCommHandler.ts
@@ -104,6 +104,7 @@ type FetchT = typeof fetch;
 const registerResSchema = z.object({
   status: z.number().default(200),
   skipped: z.boolean().optional().default(false),
+  modified: z.boolean().optional().default(false),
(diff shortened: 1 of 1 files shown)


jpwilliams:
@goodoldneon How does `modified` differ from `skipped`?

It looks like if `skipped: false` then we assume that the functions were changed, so a new deployment was created. Does this mean that `skipped: false == modified: true`?

goodoldneon:
@jpwilliams, API currently returns `modified` instead of `skipped`. But maybe we should change things on the API side to match what the Dev Server returns?

jpwilliams (approved):

PR #252, opened 2023-06-30
Add support for Koa

AnandChowdhary:
Hey team, we use [Koa](https://koajs.com) and have made a `serve` handler based on the Express handler.

I've added `koa` and `@types/koa` as dev dependencies. I don't know how the tests work (seems like you'll have to create a `@local/koa`?) so I could not add them, could someone from the team please take over? You would also have to add Koa to `SupportedFrameworkName`.


jpwilliams:
Thanks loads for this, @AnandChowdhary!

Superseding this with #262 as there are some changes to the repo structure coming that make it easier to handle in a new PR. 🙂 Includes tests and an example repo.

PR #253, opened 2023-06-30
Always add the `ts` field when sending events.

tonyhb:
This is auto-filled by the event server when missing, and fixes any caching issues people may see if their server configs cache outgoing requests (uh... server actions :D)

### DIFF
--- a/.changeset/five-bats-begin.md
+++ b/.changeset/five-bats-begin.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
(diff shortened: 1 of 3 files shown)


djfarrelly (approved):
This is probably a good practice anyway because it's better to have the timestamp at the origin rather than our server. 

This looks like a safe change to me to roll out as a patch. We could also backport this to 1.x after this is merged. I added the backport github label which should create a PR after this is merged.

PR #256, opened 2023-06-30
Return event IDs when sending events

djfarrelly:
## Summary

We should return the response data from the Event API, including `ids` when events are sent. This returns an object for future extensibility beyond `ids`.

Applies the changes in #195, but in a post-2.0, post-Middleware setup.

- [x] Add return data to `inngest.send()`
- [ ] Add return data to `steps.sendEvent()`

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -142,11 +142,7 @@ jobs:
       # Install dependencies in the example repo
       # Don't use "npm ci", "--immutable" etc., as example repos won't be
       # shipped with lock files.
-      - name: Install example dependencies
-        run: npm install
-        working-directory: examples/${{ matrix.example }}
-
-      - name: Add local SDK to example
+      - name: Add local SDK to example with dependencies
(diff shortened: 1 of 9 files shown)


djfarrelly:
Hey @jpwilliams - now that you're back online, could you help move this one forward? I'm happy to riff on this tomorrow morning if you want to chat about it!

src/components/InngestMiddleware.ts
```diff
@@ -456,17 +458,26 @@ type MiddlewareSendEventArgs = Readonly<{
...
+) => MaybePromise<{
   payloads?: EventPayload[];
 } | void>;
 
+/**
+ * Arguments for the SendEventOutput hook
+ *
+ * @internal
+ */
+type MiddlewareSendEventOutputArgs = { result: Readonly<SendEventBaseOutput> };
```
jpwilliams, line 472: This needs to retain `payloads` (`MiddlewareSendEventArgs`).

PR #261, opened 2023-07-12
Move to pnpm workspaces

jpwilliams:
# Summary

Shifts the project structure around to use pnpm workspaces so that we can maintain and publish multiple packages from a single repo.

Also makes `examples/` more prominent and shifts integration testing to use these local example apps and functions instead of remote `inngest/sdk-example-*` repositories; we can move repos there and manage them in this central location.


tonyhb (approved):

goodoldneon (approved):

jpwilliams:
Waiting on inngest/inngest#554, then will merge.

IGassmann (approved):
TLDR; x) but really excited about this!

PR #266, opened 2023-07-14
Add support for streaming to `inngest/remix`

jpwilliams:
# Summary

Adds streaming support to the `inngest/remix` handler.

### DIFF
--- a/src/remix.ts
+++ b/src/remix.ts
@@ -1,11 +1,38 @@
 import {
   InngestCommHandler,
+  type ActionResponse,
   type ServeHandler,
 } from "./components/InngestCommHandler";
 import { headerKeys, queryKeys } from "./helpers/consts";
 import { type SupportedFrameworkName } from "./types";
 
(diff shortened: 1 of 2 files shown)


src/remix.ts
```diff
@@ -66,27 +93,8 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts): unknown => {
...
-        Res = Response;
-      }
-
-      return new Res(body, {
-        status,
-        headers,
-      });
-    }
+    createNewResponse,
+    createNewResponse
```
darwin67, line 97: why twice?

jpwilliams: Two separate definitions that happen to be the same in this case: https://github.com/inngest/inngest-js/blob/9aa99f4a26220a672dbec78ab89fc23404fc29ba/src/components/InngestCommHandler.ts#L364-L407

darwin67 (approved):

PR #272, opened 2023-07-26
Assert types instead of casting in Express

goodoldneon:
# Description

- In the Express `serve` function, assert types instead of casting.
- When parsing function data, include Zod validation errors in the pretty error.

### DIFF
--- a/src/express.ts
+++ b/src/express.ts
@@ -30,10 +30,10 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
         run: () => {
           if (req.method === "POST") {
             return {
-              fnId: req.query[queryKeys.FnId] as string,
-              stepId: req.query[queryKeys.StepId] as string,
-              data: req.body as Record<string, unknown>,
-              signature: req.headers[headerKeys.Signature] as string,
+              fnId: getFromQuery(req, queryKeys.FnId),
+              stepId: getFromQuery(req, queryKeys.StepId),
+              data: getBody(req),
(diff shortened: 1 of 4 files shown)


src/express.ts
```diff
@@ -67,3 +67,45 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
 
   return handler.createHandler();
 };
+
+function getBody(req: Request): Record<string, unknown> {
+  const body: unknown = req.body;
+
+  if (body === undefined) {
+    throw new Error("missing request body");
```
jpwilliams, line 75: Could this be a `throw new Error(prettyError(...))` which directs users to https://www.inngest.com/docs/sdk/serve#framework-express so they can fix the issue for themselves?

src/express.ts
```diff
@@ -67,3 +67,45 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
...
+  }
+
+  if (typeof body !== "object" || body === null || Array.isArray(body)) {
+    throw new Error("body is not an object");
```
jpwilliams, line 79: Could this be a `throw new Error(prettyError(...))` which directs users to https://www.inngest.com/docs/sdk/serve#framework-express so they can fix the issue for themselves?

jpwilliams (commented):
This should be great for catching this mis-step! Would love to shift a couple of the errors to something more actionable for the user and we can ship ASAP!

jpwilliams (approved):

PR #276, opened 2023-07-30
Add batch-related mutual exclusivity to createFunction

goodoldneon:
# Description

Add type errors to `createFunction` when combining batching with unsupported features. We don't support rate limiting or cancellation when batching is enabled ([docs](https://www.inngest.com/docs/guides/batching)).


jpwilliams (approved):
Awesome! I added some tests.

PR #277, opened 2023-08-01
Fix: Use fetch helper for environment and testing support

djfarrelly:
## Description

This re-uses the `getFetch` helper to mirror the way that the rest of the SDK selects what `fetch` to use or use `cross-fetch`. This also will allow folks passing their own `fetch` implementation of their choosing to use that same implementation with the `InngestApi` class for testing purposes.

This helps with two things:
- Allows the user to pass their own fetch implementation or mock during testing.
- Better support fetch in different environments like edge

[Original report from community](https://discord.com/channels/842170679536517141/1135910511616208988/1135910511616208988)

### DIFF
--- a/.changeset/brave-frogs-beg.md
+++ b/.changeset/brave-frogs-beg.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix `cross-fetch` import issue in testing environemtnst. API package also uses custom `fetch` passed via arguments.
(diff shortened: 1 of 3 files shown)


djfarrelly:
User confirmed new version fixed their issue with tests

darwin67 (approved):

jpwilliams (approved):

PR #288, opened 2023-08-11
Vercel/node serve handler

cohlar:
## Summary
Fixes https://github.com/inngest/inngest/issues/553

FYI - I have tested the Express `serve` handler on Vercel node serverless functions, and [it failed to deploy](https://discord.com/channels/842170679536517141/1136943054071279646). I do not know what server framework Vercel uses under the hood, but the `Request` and `Response` are wrapped differently from Express.

## Checklist

- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR
- [ ] Added unit/integration tests
- [ ] Added changesets if applicable

## Related

- INN-

PR #290, opened 2023-08-11
Support Vercel's Express in `"inngest/express"` serve handler

jpwilliams:
## Summary

Adds support for Vercel's Node Express server, which has differing reported types from the `express` package.

## Checklist

## Related

- Supersedes #288
- Closes inngest/inngest#553


packages/inngest/src/express.ts
```diff
@@ -20,10 +22,20 @@ export const serve: ServeHandler = (nameOrInngest, fns, opts) => {
...
+    ) => {
+      const hostname = req.headers["host"];
+
+      const protocol = hostname?.includes("://")
+        ? ""
+        : `${req.protocol || "https"}://`;
+
+      const url = new URL(
+        req.originalUrl || req.url || "",
+        `${protocol}${hostname || ""}`
```
djfarrelly, line 37: I think if the hostname is null this will throw an error. Is there a possibility that `req.headers["host"]` is null or is the `|| ""` just defining against typescript types?

```ts
// this throws an error
new URL("/pathname", "https://") 
```

If it's a TS thing I wonder if it may be nicer to conditionally check for hostname and if it's blank, we look for the `Inngest` client's `serveHost`?

djfarrelly (commented):
Hey @jpwilliams - this change looks good, just have a question on how we're handling if the hostname is null. I know this was already there before this PR, but wanted to raise it while we were making this improvement!

packages/inngest/src/express.ts
jpwilliams: Good catch!

tonyhb (approved):

PR #293, opened 2023-08-14
Reveal `key` in `concurrency` options

jpwilliams:
## Summary

Reveals the `key` option when setting `concurrency`.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A
- [ ] ~~Added unit/integration tests~~ N/A

### DIFF
--- a/.changeset/famous-experts-explain.md
+++ b/.changeset/famous-experts-explain.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Add `key` to `concurrency` types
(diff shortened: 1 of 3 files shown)


tonyhb (approved):

cohlar:
May I ask how this new `key` works? I can't find anything about it in the documentation or in the source code comments.

1. If I define a function with `concurrency: { limit: 1, key: 'foo' },`, then I assume two functions with key `foo` cannot run concurrently while two functions with different keys can run concurrently - did I get that part right?
2. Since we don't have access to the `event` object, how should the `key` string be defined? - e.g. is it similar to the `match` option in `waitForEvents()` (`key: 'data.userId'`)?
3. Can we define a `key` based on two different properties from the `event` object? e.g. ``` key: "user='data.userId'&region='data.region'" ```

PR #298, opened 2023-08-17
Add ability to provide Zod schemas using objects and literals

jpwilliams:
## Summary

Adds the ability to provide Zod schemas using `z.object()` instead of requiring a record format.

```ts
// Previously we supported this
new EventSchemas().fromZod({
  "test.event": {
    data: z.object({ a: z.string() }),
    user: z.object({ b: z.number() }),
  },
});

// Now we ALSO support this
new EventSchemas().fromZod([
  z.object({
    name: z.literal("test.event"),
    data: z.object({ a: z.string() }),
    user: z.object({ b: z.number() }),
  }),
]);
```

This should help users that wish to declare their events piece-by-piece instead of in a single object.

```ts
const firstEvent = z.object({
  name: z.literal("app/user.created"),
  data: z.object({ id: z.string() }),
});

const secondEvent = z.object({
  name: z.literal("shop/product.deleted"),
  data: z.object({ id: z.string() }),
});

new EventSchemas().fromZod([firstEvent, secondEvent]);
```

This _does_ encourage folks to want to share the event schemas manually and use `z.infer` to understand payload shape, but we should still encourage use of `GetEvents` where possible.

```ts
const inngest = new Inngest({
  name: "My App",
  schemas: new EventSchemas().fromZod([firstEvent, secondEvent]),
});

type Events = GetEvents<typeof inngest>;
type ShopProductDeletedEvent = Events["shop/product.deleted"];
```

There's a new type exported, `LiteralZodEventSchema`, that can be used to ensure your Zod schema matches what's expected without having to use `new EventSchemas()`. Not sure why you'd immediately need this, but it's there in case folks want to feel extra safe and can provide some autocomplete when defining events outside.

image

We also have a cheeky error for empty names, to ensure folks don't cause a problem for themselves.

image

## Checklist

## Related

- inngest/website#479

### DIFF
--- a/.changeset/warm-lemons-tan.md
+++ b/.changeset/warm-lemons-tan.md
@@ -1,0 +1,49 @@
+---
+"inngest": minor
+---
+
+Add the ability to provide Zod schemas using `z.object()` instead of requiring a record format
+
+```ts
+// Previously we supported this
+new EventSchemas().fromZod({
+  "test.event": {
+    data: z.object({ a: z.string() }),
+    user: z.object({ b: z.number() }),
+  },
+});
+
+// Now we ALSO support this
+new EventSchemas().fromZod([
+  z.object({
+    name: z.literal("test.event"),
+    data: z.object({ a: z.string() }),
+    user: z.object({ b: z.number() }),
+  }),
+]);
+```
+
+This should help if you wish to declare your events piece-by-piece instead of in a single object.
+
+```ts
+const firstEvent = z.object({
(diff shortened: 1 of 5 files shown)


joelhooks:
should this work with some format of "fromSchema" or similar so valibot or others could be used versus naming zod explicitly? https://valibot.dev/

I don't know if they produce similar output, mostly just curious.

goodoldneon (approved):

PR #367, opened 2023-10-19
Fix client allowing extraneous properties upon creation

jpwilliams:
## Summary

Incorrect properties would be ignored and produce no type errors when creating a client via `new Inngest()`. For example, the following would show no errors, even though `baseURL` should be `baseUrl`:

```ts
new Inngest({
  id: "local-dev",
  name: "Local Dev",
  schemas: new EventSchemas().fromUnion<EventUnion>(),

  // Wrong but not error
  baseURL: "http://localhost:9999",
});
```

This PR ensures we're looking for the exact object instead.

Currently, we also ensure we allow excess properties if the options passed are a variable instead of an object literal, as this may indicate that the user wants to do more with the object.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A Improvement


goodoldneon (approved):

jpwilliams:
Addressed at runtime instead.

PR #375, opened 2023-10-25
Fix `stringify()` failing to serialize `BigInt`; remove instead

jpwilliams:
## Summary

[`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)s are not natively parsed by JSON. Attempting to return a `BigInt` from a step or function will throw an error. This error was hard to see, which is fixed in #374.

Our types reflect that `BigInt` is unparseable and will type it as `never`, though these were obscured from the user a touch, which is fixed in #371.

For this reason, this PR ensures `BigInt`s are removed during parsing instead of throwing. We can investigate adding `BigInt` support across our SDKs past this fix; the `n` suffix seems an unwritten standard.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A

## Related

- #371
- #374

### DIFF
--- a/.changeset/clever-dolphins-return.md
+++ b/.changeset/clever-dolphins-return.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix failing to parse `BigInt` during step/function result serialization; it is now correctly typed and returned as `never`
--- a/packages/inngest/src/components/InngestStepTools.test.ts
+++ b/packages/inngest/src/components/InngestStepTools.test.ts
@@ -195,28 +195,52 @@ describe("run", () => {
       symbol: Symbol("foo"),
       map: new Map(),
       set: new Set(),
+      bigint: BigInt(123),
+      typedArray: new Int8Array(2),
+      promise: Promise.resolve(),
+      weakMap: new WeakMap([[{}, "test"]]),
+      weakSet: new WeakSet([{}]),
     };
 
     const output = step.run("step", () => input);
 
-    assertType<
-      Promise<{
+    type Expected = {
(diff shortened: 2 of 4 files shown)


goodoldneon (approved):

jpwilliams:
Add to tests:

- `BigInt` type check
- Runtime tests for each type

https://github.com/inngest/inngest-js/blob/4c71b9a699529772bc787b7085f964aab67e6f78/packages/inngest/src/components/InngestStepTools.test.ts#L179-L221

davewasmer:
I understand the logic here, but thought I would add a less-than-ideal experience I just had with this.

If the type returned by a `step.run()` is `bigint | null`, it narrows to just `null`. I had some admittedly loose code later in my workflow which accepted a type of `{ foo: bigint | null }` when it really should have been `{ foo: bigint }`, so I ran into problems.

However, this "drop BigInts silently" behavior made debugging this much harder.

I eventually realized something unusual was happening with Inngest when my `step.run()` was returning something different than the actual step function returned. Suspecting serialization issues, I ran a manual test of `JSON.stringify({ n: BigInt(1) })`, but I got an error, so I assumed something else was going on (or else I would have seen the same error in my workflow).

Only after extensive testing did I finally search the repo issues here and find this PR.

Having something behave _almost_ like the standard JSON.stringify, but vary in subtle & undocumented ways (the docs make no mention of dropping BigInts) is very surprising. Personally I'd prefer the hard errors, but even a quick docs mention would be useful.

PR #387, opened 2023-11-08
Added framework-nestjs example with dependency injection into inggest function

anasabdullahysfzai:
## Summary
Added framework-nestjs example with dependency injection into inggest function. Existing example doesnt mention about injecting dependency in inngest function thats why we added a logic to do so.

### DIFF
--- a/examples/framework-nestjs/.eslintrc.js
+++ b/examples/framework-nestjs/.eslintrc.js
@@ -1,0 +1,25 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: 'tsconfig.json',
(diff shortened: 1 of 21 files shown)


anasabdullahysfzai:
@jpwilliams can you please review this PR

jpwilliams (approved):
Awesome! Thanks, @anasabdullahysfzai.

I'd love to also get an example going with <kbd>[nest-inngest](https://www.npmjs.com/package/nest-inngest)</kbd> ([thawankeane/nest-inngest](https://github.com/thawankeane/nest-inngest)) later.

I've made it slightly more generic in a few places and we can improve as we go. Thanks so much for the contribution! 🙌

PR #412, opened 2023-11-29
Add Google Functions Framework example

djfarrelly:
## Summary

This adds an example from our docs with working code that also supports TypeScript which the [Functions Framework's README](https://github.com/GoogleCloudPlatform/functions-framework-nodejs) doesn't include.

## Checklist

- [x] Added a [docs PR](https://github.com/inngest/website) that references this PR (already in docs!)

## Related
https://github.com/inngest/website/pull/563

### DIFF
--- a/examples/framework-google-functions-framework/README.md
+++ b/examples/framework-google-functions-framework/README.md
@@ -1,0 +1,29 @@
+# Inngest Google Functions Framework Template
+
+This is a [Google Functions Framework](https://github.com/GoogleCloudPlatform/functions-framework-nodejs) project. It is a reference on how to send and receive events with Inngest and Google's Functions Framework for Node.js. Functions Framework is an Express.js compatible API.
+
+## Getting Started
+
+Use [`create-next-app`](https://www.npmjs.com/package/create-next-app) with [npm](https://docs.npmjs.com/cli/init), [Yarn](https://yarnpkg.com/lang/en/docs/cli/create/), or [pnpm](https://pnpm.io) to bootstrap the example:
+
+```bash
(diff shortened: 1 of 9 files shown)


examples/framework-google-functions-framework/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+Enter the directory and run:
+
+```bash
+# Transpile TypeScript to JavaScript
+npm run build
+# Run the service
+npm start
+```
```
djfarrelly, line 28: @jpwilliams - I noticed that other READMEs don't have a command on how to run the service itself. Is that something that we should add to all READMEs or does the create-next-app give more info on that? This has a discrete build step, so I thought it was worth calling this out. What do you think?

jpwilliams: Yeah, that'd be good! `create-next-app` does indeed tell you what to do (`cd` in and `npm run dev`), but it's still a missing step on the readme.

examples/framework-google-functions-framework/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "framework-google-functions-framework",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "tsc",
+    "start": "functions-framework --target=inngest --port=3000",
+    "dev": "nodemon -e ts -x \"npm run build && npm run start\" -i node_modules"
```
jpwilliams, line 9: I added a `dev` script to get it working with tests.

PR #427, opened 2023-12-14
Fix missing `EventPayload.id`

### LINKED ISSUE #423: [BUG] `id` field is missing from the type definition of `EventPayload`
**Describe the bug**
The optional `id` field for idempotent event processing is not defined in the `EventPayload` interface and therefore provides no type checking / editor integration.

see: https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/types.ts#L305

docs: https://www.inngest.com/docs/reference/events/send#inngest-send-event-payload-event-payload-promise

jpwilliams:
## Summary

Fixes #423, where `id` is missing in types when attempting to send an event using `inngest.send()` or `step.sendEvent()`.

## Checklist

## Related

- Fixes #423

### DIFF
--- a/packages/inngest/src/types.ts
+++ b/packages/inngest/src/types.ts
@@ -304,7 +304,10 @@ export type AnyHandler = Handler<any, any, any, any>;
  */
 export interface EventPayload {
   /**
-   * A unique identifier for the event
+   * A unique identifier for the type of event. We recommend using lowercase dot
+   * notation for names, prepending `prefixes/` with a slash for organization.
+   *
+   * e.g. `cloudwatch/alarms/triggered`, `cart/session.created`
    */
   name: string;
 
--- a/packages/inngest/src/types.ts
+++ b/packages/inngest/src/types.ts
@@ -304,11 +304,11 @@ export type AnyHandler = Handler<any, any, any, any>;
  */
 export interface EventPayload {
   /**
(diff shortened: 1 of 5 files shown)


packages/inngest/src/types.ts
```diff
@@ -304,7 +304,19 @@ export type AnyHandler = Handler<any, any, any, any>;
  */
 export interface EventPayload {
   /**
-   * A unique identifier for the event
+   * A unique id used to idempotently ingest a given event payload once (i.e.
+   * deduplication).
+   *
+   * Set this when sending events to ensure that the event is only processed
+   * once; if an event with the same ID is sent again, it will be ignored.
```
tonyhb, line 311:
```suggestion
   * A unique id used to idempotently process a given event payload.
   *
   * Set this when sending events to ensure that the event is only processed
   * once; if an event with the same ID is sent again, it will not invoke functions.
```

tonyhb (approved):

packages/inngest/src/types.ts
tonyhb: The aim here is to make it clear that funciton runs are handled idempotently.  Handling idempotency at the event level is too much work for the events API, which handles tens of thousands of events per second 😬

jpwilliams: (ty - committed this)

PR #385, opened 2023-11-07
Fix a TS error when using Inngest in projects with `composite` setting

MonsterDeveloper:
## Summary

Adds some type exports to `index.ts` that fix a TS error in `composite` projects.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A
- [ ] ~~Added unit/integration tests~~ N/A
- [x] Added changesets if applicable

## Related
Fixes #384

Issue #384, opened 2023-11-07
[BUG] The inferred type cannot be named without a reference to '../node_modules/inngest/components/InngestFunction' when using `composite` 


MonsterDeveloper:
**Describe the bug**
When enabling the `composite` in a project's `tsconfig.json`, the compiler shows an error when using `inngest.createFunction`:

```
The inferred type of 'default' cannot be named without a reference to '../node_modules/inngest/components/InngestFunction'. This is likely not portable. A type annotation is necessary. typescript(2742)
```

**To Reproduce**

1. Create a function with `inngest.createFunction`
2. Enable `composite` in `tsconfig.json`

[...]

PR #437, opened 2023-12-18
Add experimental Bun example

djfarrelly:
## Summary

An example using [Bun](https://bun.sh/) and the existing `edge` serve handler. 

### Notes

* There is a module resolution issue with the default `tsconfig.json` generated via `bun init` which can be "fixed" by changing the `"moduleResolution"` to `"node"`. `The inferred type of 'helloWorld' cannot be named without a reference to '../node_modules/inngest/types'. This is likely not portable. A type annotation is necessary.`

## Checklist

## Related
N/A


examples/framework-bun/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "module": "esnext",
+    "target": "esnext",
+    "moduleResolution": "node",
```
djfarrelly, line 6: By default this is `"bundler"` which makes all of the exports defined in the `./inngest` files have this TS error:
```
The inferred type of 'functions' cannot be named without a reference to '../node_modules/inngest/types'. This is likely not portable. A type annotation is necessary.ts(2742)
```

jpwilliams: I've moved it back to the default of `"bundler"` - this is a good test for issues like #384 (also see #385 / #460).

jpwilliams (approved):

PR #462, opened 2024-01-15
Allow a single-element `concurrency` option in function definitions

jpwilliams:
## Summary

Fixes `concurrency: [ConcurrencyOption]` not being allowed.

## Checklist

- [ ] ~~Added unit/integration tests~~ N/A

### DIFF
--- a/packages/inngest/src/types.ts
+++ b/packages/inngest/src/types.ts
@@ -18,6 +18,7 @@ import { type internalEvents } from "./helpers/consts";
 import {
   type IsStringLiteral,
   type ObjectPaths,
+  type RecursiveTuple,
   type StrictUnion,
 } from "./helpers/types";
 import { type Logger } from "./middleware/logger";
@@ -791,12 +792,13 @@ export interface FunctionOptions<
    * can occur across all runs of the function.  A value of 0 (or undefined) means
    * use the maximum available concurrency.
    *
-   * Specifying just a number means specifying only the concurrency limit.
+   * Specifying just a number means specifying only the concurrency limit. A
+   * maximum of two concurrency options can be specified.
    */
   concurrency?:
     | number
     | ConcurrencyOption
-    | [ConcurrencyOption, ConcurrencyOption];
+    | RecursiveTuple<ConcurrencyOption, 2>;
 
   /**
    * batchEvents specifies the batch configuration on when this function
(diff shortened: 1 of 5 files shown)


packages/inngest/src/types.ts
```diff
@@ -791,12 +792,13 @@ export interface FunctionOptions<
...
    *
-   * Specifying just a number means specifying only the concurrency limit.
+   * Specifying just a number means specifying only the concurrency limit. A
+   * maximum of two concurrency options can be specified.
    */
   concurrency?:
     | number
     | ConcurrencyOption
-    | [ConcurrencyOption, ConcurrencyOption];
+    | RecursiveTuple<ConcurrencyOption, 2>;
```
tonyhb, line 801: I love the type - but how does this look in VSCode?  Is this easy to read?

jpwilliams: @tonyhb - Looks exactly as you'd expect, e.g.
```ts
number | ConcurrencyOption | [ConcurrencyOption] | [ConcurrencyOption, ConcurrencyOption]
```
image

TypeScript will always try to expand types as much as it can, so as long as it has all the information it needs to resolve it, it will.

```ts
type _T0 = RecursiveTuple<string, 2>;
//   ^? type _T0 = [string] | [string, string]
```

That also means that it expands type aliases, too.

```ts
type Foo = boolean;
type _T0 = RecursiveTuple<Foo, 2>;
//   ^? type _T0 = [boolean] | [boolean, boolean]
```

Object types, though, often keep their alias intact.

```ts
type Foo = {
  bar: boolean;
};
type _T0 = RecursiveTuple<Foo, 2>;
//   ^? type _T0 = [Foo] | [Foo, Foo]
```

There are tricks to remove that alias by using a mapped type like [`Simplify<T>`](https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts) in [sindresorhus/type-fest](https://github.com/sindresorhus/type-fest).

There used to be some reliable ways to preserve an alias, too, if you'd prefer that TypeScript _didn't_ expand your type into oblivion. For example, `RecursiveTuple<ConcurrencyOption, 10>` would produce a pretty huge type, so maybe we'd prefer to keep the alias and have `number | ConcurrencyOption | TupleOfMax<ConcurrencyOption, 10>` (naming is hard).

Those methods mostly involved "branding" your type by intersecting it with an object, but later TypeScript versions are good enough to still expand what they can and you end up with some ugly partial expansion that looks worse than either path.

goodoldneon (approved):

PR #484, opened 2024-02-12
Add `timeout` to `step.invoke()` options

jpwilliams:
## Summary

Adds a `timeout` option to `step.invoke()`, allowing users to specify a timeout other than the default when invoking a function as a step.

Uses the the same `timeout` as `step.waitForEvent()` and function config.

```ts
await step.invoke("Example", {
  function: myFn,
  timeout: "1h",
});
```

The Executor already supports this:

https://github.com/inngest/inngest/blob/658e7a6e003b1a3b4fb0f4349c3c7e635e09b9de/pkg/execution/state/driver_response.go#L159-L163

## Checklist

## Related

- inngest/website#676

### DIFF
--- a/packages/inngest/src/components/InngestStepTools.ts
+++ b/packages/inngest/src/components/InngestStepTools.ts
@@ -486,6 +486,9 @@ type InvocationOpts<TFunction extends InvokeTargetFunctionDefinition> =
        * of this time, at which point the retured promise will be rejected
        * instead of resolved with the output of the invoked function.
        *
+       * Note that the invoked function will continue to run even if this step
+       * times out.
+       *
        * The time to wait can be specified using a `number` of milliseconds, an
        * `ms`-compatible time string like `"1 hour"`, `"30 mins"`, or `"2.5d"`,
        * or a `Date` object.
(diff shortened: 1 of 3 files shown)


packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -473,7 +479,21 @@ type InvocationTargetOpts<TFunction extends InvokeTargetFunctionDefinition> = {
...
+       * The step function will wait for the invocation to finish for a maximum
+       * of this time, at which point the retured promise will be rejected
+       * instead of resolved with the output of the invoked function.
+       *
+       * The time to wait can be specified using a `number` of milliseconds, an
+       * `ms`-compatible time string like `"1 hour"`, `"30 mins"`, or `"2.5d"`,
+       * or a `Date` object.
+       *
+       * {@link https://npm.im/ms}
+       */
```
djfarrelly, line 494: We may want to add a disclaimer to this like "The invoked function will not be cancelled after the timeout" - I could see some users thinking that this is tells _the invoked function to timeout_, so adding that could be clarifying. We definitely will need this within our docs when this is released.

djfarrelly (approved):

PR #488, opened 2024-02-14
INN-2754 Add support for `INNGEST_DEV`

jpwilliams:
## Summary

Adds support for `INNGEST_DEV` and a new `isDev` option on the client.

This lightly refactors the current checks based around `isProd` and `skipDevServer()`, which were getting a little difficult to read.

- The SDK now has two "modes:" `"dev"` and `"cloud"`.
- Each mode is either **explicit** or **inferred**. An inferred mode means that the current (`v3.x.x`) version of the SDK can make a decision to attempt to contact the Dev Server. Future versions will remove this and default to `"cloud"` mode.
- Setting the `INNGEST_DEV` environment variable or the `isDev` client option **explicitly** sets the mode to either `"cloud"` or `"dev"`.
- `INNGEST_DEV` accepts some sensible defaults. We'll recommend `1` to explicitly set `"dev"` mode and `0` to explicitly set `"cloud"` mode, though it also accepts `"true"`, `"y"`, `"no"`, etc.
- Explicitly setting either mode also sets the event ingestion and syncing URLs. They continue to be further overwritten by passing `INNGEST_BASE_URL`, `INNGEST_API_BASE_URL`, and `INNGEST_EVENT_API_BASE_URL`.

> [!NOTE]
> To support many runtimes and environments, environment variables are not always accessible a) at all times, and b) on `process.env`. Sometimes environment variables are accessed via different global objects, or sometimes runtime objects that are passed to requests.
>
> For this reason, handling environment variables is more complex and relies on making best guesses during instantiation, then later making another decision when we have access to the environment.

Supersedes both #424 and #425.

## Checklist

- [x] Added a [docs PR](https://github.com/inngest/website) documenting these modes and the new environment variables that references this PR
- [x] Push env-related changes to the OS SDK Spec

## Related

- INN-2754
- Supersedes #424
- Supersedes #425
- inngest/website#679


packages/inngest/src/helpers/env.ts
```diff
@@ -388,3 +396,37 @@ export const getResponse = (): typeof Response => {
...
+  }
+
+  if (typeof value === "string") {
+    const trimmed = value.trim().toLowerCase();
+
+    if (trimmed === "undefined") {
+      return undefined;
+    }
+
+    if (["true", "1", "y", "yes"].includes(trimmed)) {
```
goodoldneon, line 424: Are you ok with dropping support for `"y"` and `"yes"`? I've never seen anyone set a boolean env var to those -- I've only seen `"true"` and `"1"`

goodoldneon:
LGTM, but the combination of the 2 modes + explicit/implicit is a bit of a mind f 😄

goodoldneon (approved):

tonyhb:
One additional thought:  the `INNGEST_DEV` variable in the Go SDK also controls the `INNGEST_BASE_URL`;  if you set `INNGEST_DEV=http://192.168.1.200` then dev mode is enabled and we use the given URL as the dev server.  I think this makes lots of sense for development and would also like to standardize around this in the future.  Not necessarily for this PR.

jpwilliams:
> @tonyhb: One additional thought: the `INNGEST_DEV` variable in the Go SDK also controls the `INNGEST_BASE_URL`; if you set `INNGEST_DEV=http://192.168.1.200` then dev mode is enabled and we use the given URL as the dev server. I think this makes lots of sense for development and would also like to standardize around this in the future. Not necessarily for this PR.

Yeah, I was thinking of this too. There are so many ways to configure it at this point, though; I'm worried about perpetually making this more and more complex.

- Check `serve()` arguments
  - If `serve({ baseUrl })` then `baseUrl`
- Check explicit environment variables
  - If `INNGEST_API_BASE_URL` then use that
  - If `INNGEST_BASE_URL` then use that
- Check `new Inngest()` arguments
  - If `new Inngest({ baseUrl })` then `baseUrl`
  - If `new Inngest({ isDev: true })` then `http://localhost:8288`
- Check explicit dev environment variables
  - If `INNGEST_DEV=1` then `http://localhost:8288`
  - If `INNGEST_DEV=[A_URL]` then `[A_URL]`
- Use `isProduction()` check in serve handler, e.g. `"inngest/next"`
  - If returns `true` then `https://api.inngest.com`
  - If returns `false`, try `http://localhost:8288` but fall back to `https://api.inngest.com`
  - If returns or is `undefined`, perform inferred prod checks, e.g. `NODE_ENV`, `CF_PAGES`
    - If dev mode is inferred from this, try `http://localhost:8288` but fall back to `https://api.inngest.com`
- If nothing else matched, default to `https://api.inngest.com`

tbh the end result is that (mostly) the user only has to care about a single environment variable. I doubt that the ordering of these will be much of a consideration/issue for most.

packages/inngest/src/components/Inngest.ts
```diff
@@ -407,21 +436,9 @@ export class Inngest<TOpts extends ClientOptions = ClientOptions> {
     let url = this.sendEventUrl.href;
 
     /**
-     * INNGEST_BASE_URL is used to set both dev server and prod URLs, so if a
-     * user has set this it means they have already chosen a URL to hit.
+     * If we're in prod mode and have no key, fail now.
```
sylwiavargas, line 439:
```suggestion
     * If in prod mode and key is not present, fail now.
```

packages/inngest/src/components/Inngest.ts
```diff
@@ -432,6 +449,29 @@ export class Inngest<TOpts extends ClientOptions = ClientOptions> {
...
     }
 
+    /**
+     * If we've inferred that we're in dev mode, try to hit the dev server
+     * first to see if it exists. If it does, use it, otherwise fall back to
+     * whatever server we have configured.
+     *
+     * `INNGEST_BASE_URL` is used to set both dev server and prod URLs, so if a
+     * user has set this it means they have already chosen a URL to hit.
+     */
```
sylwiavargas, line 459:
```suggestion
     * If in dev mode, try to hit the dev server
     * first to see if it exists. If it does, use it, otherwise fall back to
     * whatever server we have configured.
     *
     * `INNGEST_BASE_URL` is used to set both dev server and prod URLs, so if a
     * user has set this it means they have already chosen a URL to hit.
     */
```

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1170,8 +1165,10 @@ export class InngestCommHandler<
   }
 
   protected validateSignature(sig: string | undefined, body: unknown) {
-    // Never validate signatures in development.
-    if (!this.isProd) {
+    // Never validate signatures outside of prod. Make sure we check the mode
```
sylwiavargas, line 1168:
```suggestion
    // Never validate signatures outside of production. Make sure to check the mode
```

sylwiavargas (approved):
Thank you! Looks good. I left a few suggestions on language but I'm approving already.

packages/inngest/src/components/Inngest.ts
djfarrelly: +1 to this clearer tone. I think we should generally avoid "we" in the SDK, so on boarding with Sylwia's suggestions!

PR #492, opened 2024-02-19
Fix delayed environment variables not using event key

### LINKED ISSUE #481: [BUG] Cloudflare workers/pages integration doesn't pick up the INNGEST_EVENT_KEY environment variable
**Describe the bug**
/inngest endpoint returns `"hasEventKey":false` even though `INNGEST_EVENT_KEY` is properly set. Same issue in development and production with Cloudflare pages. Same issue in development with `wrangler pages dev` as well as in the actual deployment. INNGEST_

**To Reproduce**

1. Create a Cloudflare Pages deployment with INNGEST_EVENT_KEY and INNGEST_SIGNING_KEY as environment variables.
2. Navigate to the /inngest endpoint
3. Get the following result: `{"message":"Inngest endpoint configured correctly.","hasEventKey":false,"hasSigningKey":true,"functionsFound":1}`

**Expected behavior**
hasEventKey should return true

N/A

- OS: [e.g. Mac, Windows]
- npm package Version 3.13.0
- Framework none
- Platform Cloudflare Pages

**Additional context**
N/A

jpwilliams:
## Summary

Fixes an issue where delayed access to environment variables was incorrectly missing setting the event key.

`Inngest.eventKey` will always be a populated string, falling back to a dummy value of `NO_EVENT_KEY_SET` if it's not specified during instantiation. We do this as API routing for ingestion (`https://inn.gs/e/:event_key`) sensibly relies on the event key being present, even in local dev.

In environments such as Cloudflare Workers where access to environment variables is delayed until the request handler is called, we'll should set the event key again if we only have a dummy and we now have a real one.

This code was erroneously checking that `eventKey` was truthy instead of calling the existing `eventKeySet()` method that considers dummy keys. 

## Checklist

## Related

- Fixes #481

### DIFF
--- a/packages/inngest/src/test/helpers.ts
+++ b/packages/inngest/src/test/helpers.ts
@@ -175,19 +175,25 @@ export const testFramework = (
     const [req, res] = createReqRes(mockReqOpts);
 
     let envToPass = { ...env };
+    let prevProcessEnv = undefined;
 
     /**
      * If we have `process` in this emulated environment, also mutate that to
      * account for common situations.
      */
     if (typeof process !== "undefined" && "env" in process) {
-      process.env = { ...process.env, ...envToPass };
+      prevProcessEnv = process.env;
+      process.env = { ...prevProcessEnv, ...envToPass };
       envToPass = { ...process.env };
     }
 
     const args = opts?.transformReq?.(req, res, envToPass) ?? [req, res];
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const ret = await (serveHandler as (...args: any[]) => any)(...args);
+
+    if (prevProcessEnv) {
+      process.env = prevProcessEnv;
+    }
 
     return (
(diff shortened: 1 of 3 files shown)


packages/inngest/src/test/helpers.ts
```diff
@@ -175,20 +175,26 @@ export const testFramework = (
...
+    let prevProcessEnv = undefined;
 
     /**
      * If we have `process` in this emulated environment, also mutate that to
      * account for common situations.
      */
     if (typeof process !== "undefined" && "env" in process) {
-      process.env = { ...process.env, ...envToPass };
+      prevProcessEnv = process.env;
+      process.env = { ...prevProcessEnv, ...envToPass };
```
goodoldneon, line 186: Why doesn't `...process.env` work? Also, is it OK to mutate `process.env`?

goodoldneon (approved):

PR #493, opened 2024-02-19
Fix mutating default register URL across requests

goodoldneon:
## Summary
Fix bug where we were mutating the default register URL across requests


jpwilliams (approved):
A triumph.

PR #460, opened 2024-01-12
Fix a TS error when using Inngest in projects with `composite` setting

MonsterDeveloper:
## Summary

This PR Adds some type exports to `index.ts` that fix a TS error in composite projects.

After changes to the `types.ts` the issue #384 appeared again. This PR is similar to #385 - it fixes the TS error by adding necessary exports to the `index.ts`

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A no docs needed
- [ ] ~~Added unit/integration tests~~ N/A not testable
- [x] Added changesets if applicable

## Related
- #385 
- #384

PR #501, opened 2024-02-23
Test for `composite: true` errors in CI, ensuring we can't ship those issues

jpwilliams:
## Summary

We've seen some issues crop up when `composite: true` is present in `tsconfig.json` files. This PR adds a `composite: true` project where we test the Inngest package for compliance for the types that it exports.

We purposefully want to limit the number of types exported from the main `"inngest"` entrypoint, as each type exported here becomes part of the public API, where changing those types is a breaking change. Therefore, the composite check gives us a (albeit inaccurate) test that can stop us shipping these dangerous changes.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Bug fix and testing

## Related

- #384
- #385 
- #437
- #460


packages/inngest/test/composite_project/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+## Do I need to update it?
+
+The main tests this project is performing is that types are correctly inferred
+by the various functions exported by the `"inngest"`. Therefore, if adding new
+API methods and calls, it's ideal to add them into this project.
+
+This code is only built and is never run, so don't worry about it actually
+working.
```
djfarrelly, line 35: 😆

djfarrelly (approved):
Awesome! Tested on my own project as well - works great! Thank you Jack 👏

PR #520, opened 2024-03-18
INN-2882 Add selective header forwarding when sending events

jpwilliams:
## Summary

Read the incoming headers tracestate and traceparent. If it is set, copy the header to any outgoing event-sending requests.

This only covers `step.sendEvent()`; we'll add later support for `client.send()` as we need async hooks for this which has scattered runtime support.

## Checklist

## Related

- INN-2882


packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -722,12 +722,35 @@ export class InngestCommHandler<
...
+          );
+          if (value) {
+            return {
+              ...(await acc),
+              [header]: value,
+            };
+          }
+
+          return acc;
+        }, Promise.resolve({}));
```
darwin67, line 745: I can't read JS. Does this means take the headers passed in, if there's nothing, default to an empty map?

jpwilliams: @darwin67 Correct.

It looks a bit more complex as this hits the user's serve handler to access each header, where the function for retrieving that header may be async. This pattern is for making those calls in parallel and then rolling up the results into a map at the end, though it's an abstract one.

I'll refactor to make it more explicit.

darwin67 (approved):

darwin67:
@jpwilliams anything still waiting on?
I think this PR can be merged as is.

PR #540, opened 2024-04-15
Add throttling config

tonyhb:
Updates the TS SDK to include throttling.

### DIFF
--- a/packages/inngest/src/components/InngestFunction.ts
+++ b/packages/inngest/src/components/InngestFunction.ts
@@ -364,6 +364,40 @@ export namespace InngestFunction {
     };
 
     /**
+     * Rate limit workflows, only running them a given number of times (limit) per
+     * period. This can optionally include a `key`, which is used to further
+     * constrain throttling similar to idempotency.
+     */
+    throttle?: {
+      /**
+       *  An optional expression which returns a throttling key for controlling throttling.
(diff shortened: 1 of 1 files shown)


packages/inngest/src/components/InngestFunction.ts
```diff
@@ -363,6 +363,40 @@ export namespace InngestFunction {
       period: TimeStr;
     };
 
+    /**
+     * Rate limit workflows, only running them a given number of times (limit) per
```
goodoldneon, line 367: Does "workflow" mean "function run"? I don't think we use the term "workflow" much anymore

```suggestion
     * Rate limit function runs, only running them a given number of times (limit) per
```

packages/inngest/src/components/InngestFunction.ts
```diff
@@ -363,6 +363,40 @@ export namespace InngestFunction {
       period: TimeStr;
     };
 
+    /**
+     * Rate limit function runs, only running them a given number of times (limit) per
+     * period. This can optionally include a `key`, which is used to further
+     * constrain throttling similar to idempotency.
+     */
+    throttle?: {
```
jpwilliams, line 371: Does this description need to change? Looks like it's the same as the one for `rateLimit` atm.

Issue #546, opened 2024-04-18
More info needed when there's a NESTING_STEPS error


maktouch:
**Is your feature request related to a problem? Please describe.**
We've quickly migrated from graphile/worker to inngest. We now have over 160 functions. In my logs, I see "NESTING_STEPS" but there's no info to quickly debug that, as in, what function or step is causing this. 

**Describe the solution you'd like**
I'd love to get the parent function, or the event, or even the step name, so I can go and fix this. 

I can make a PR in this repo if you think it's valid!

jpwilliams:
This is shipped in [`3.18.1`](https://github.com/inngest/inngest-js/releases/tag/inngest%403.18.1) following #549.

Thanks, @maktouch!

I'm also noting https://github.com/inngest/inngest-js/pull/549#issuecomment-2070205258, which highlights how rubbish the stack trace is for this instance: no usable lines and the `prettyError` header/footer have survived. This will be tracked in #531.

![CleanShot 2024-04-22 at 12 47 33](https://github.com/inngest/inngest-js/assets/1134251/bc33c398-0ba8-4c95-8e89-6625e00e3abe)

Issue #531, opened 2024-03-27
Save and pull `stack` on entrypoints


jpwilliams:
## Summary

When the library emits a log to try and help the user diagnose or resolve an issue, we usually include a stack trace in order to provide them with a location to start the investigation.

More often than not, however, this stack trace is loaded with internal calls, flowing in and out of async, or other calls in the user's application.

![image](https://github.com/inngest/inngest-js/assets/1736957/2742838d-1fa4-4295-ad86-3272de3506e0)

On every entrypoint via userland code, we should store a stack trace ahead of time, such that any logs emitted to users via `prettyError()` can be attributed to those traces instead of our internal ones.

PR #549, opened 2024-04-20
Add step name or ID when there’s a NESTING_STEPS error

maktouch:
## Summary
When there's a NESTING_STEPS error, debugging it is tricky because the stack traces can't necessarily determine the correct faulty file. 

## Checklist
~~- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A, this is augmenting an existing error
~~- [ ] Added unit/integration tests~~ N/A, I don't think it's worth it
~~- [ ] Added changesets if applicable~~ N/A, I don't think it's worth it. 

## Related
- INN-2958
- #546

### DIFF
--- a/.changeset/three-shoes-shave.md
+++ b/.changeset/three-shoes-shave.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Add step name or ID when there’s a NESTING_STEPS error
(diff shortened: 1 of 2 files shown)


jpwilliams:
This is great, @maktouch! Thanks for the contribution.

In terms of the original issue (#546) and not being able to see which step is causing the problem, would #531 also be a part of helping with this?

Basically, is the current stack trace useless as it's mostly listing calls internal to this library?

maktouch:
Ideally, I think better stack traces are the way to go, so #531 makes sense. 

This is how it currently looks

CleanShot 2024-04-22 at 12 47 33

PR #557, opened 2024-04-29
Add environment variable handling to Hono example

arjunyel:
## Summary

Add environment variable handling to Hono example

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A
- [ ]  ~~Added unit/integration tests~~ N/A
- [ ]  ~~Added changesets if applicable~~ N/A

Issue #556, opened 2024-04-28
[BUG] Inngest not working with hono


BernardoSM:
**Describe the bug**
3 days ago was merged hono framework for inngest that you can see [here](https://github.com/inngest/inngest-js/commit/24f1e7de92d609e0a9ee9df1e3dfa0d2f7c8f682).

I tried to run the [example](https://github.com/inngest/inngest-js/tree/main/examples/framework-hono) and works pretty well, but when I run [inngest dev server](https://www.inngest.com/docs/local-development#inngest-dev-server) I'm receiving some errors.

**To Reproduce**

1. Follow [Getting started](https://github.com/inngest/inngest-js/tree/main/examples/framework-hono#getting-started), run app locally and open `http://localhost:3000/api/inngest` to check if inngest is working
2. Run `npx inngest-cli@latest dev` in terminal

**Expected behavior**
Not show any error on terminal that is running hono (cloudflare worker) with inngest.
Not show two apps on inngest dev server

**Code snippets / Logs / Screenshots**

<details><summary>Hono terminal errors:</summary>

```bash
[wrangler:inf] GET /api/inngest 500 Internal Server Error (2ms)
✘ [ERROR] workerd/server/server.c++:3340: error: Uncaught exception: kj/async-io-unix.c++:569: disconnected: ::read(fd, buffer, maxBytes): Connection reset by peer

  stack:
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@55c6383
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@55c43e5
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@405a071
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@40778a0
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@40bfeb1
  /home/bernardosm/blindpaylabs2/inngest-hono/node_modules/@cloudflare/workerd-linux-64/bin/workerd@40835e0
  workerd/server/server.c++:3340: error: Uncaught exception:
  kj/async-io-unix.c++:186: disconnected: ::write(fd, buffer, size):
  Connection reset by peer
  stack:
[...]

PR #571, opened 2024-05-15
Fix `"inngest/hono"` serve handler not parsing environment variables

### LINKED ISSUE #560: [BUG] Hono Adapter not working on Cloudflare Worker
**Describe the bug**
The Hono adapter does not work when deployed with Cloudflare Workers. All request throw `Failed calling `url` from serve handler when starting to handle request; Invalid URL string.","stack":"TypeError: Failed calling `url` from serve handler when starting to handle request; Invalid URL string.`.

The reason is that the hono adapter is building the url from the host header:
https://github.com/inngest/inngest-js/blob/f053ef994709441dba20756378d72140811d1cd6/packages/inngest/src/hono.ts#L38

but in Cloudflare, the host does not include the protocol. The Cloudflare adapter therefore adds it manually:
https://github.com/inngest/inngest-js/blob/f053ef994709441dba20756378d72140811d1cd6/packages/inngest/src/cloudflare.ts#L38

To fix this, simply replace the url callback from hono with the one in Cloudflare. It would be nice if we could adapt the hono adapter so that it prefixes the protocol to the host if the host does not have it yet.

jpwilliams:
## Summary

The `"inngest/hono"` serve handler was not fetching environment variables, which are not reliably accessible via globals in common environments such as Cloudflare Workers.

In addition, `url` handling was a little wonky and didn't seem to handle many variants. The hope was that Hono handled that to ensure that `c.req.url` is always/never absolute, but this doesn't seem to be the case.

## Checklist

## Related

- Supersedes #557 as we should no longer be required to manually set signing and event keys
- Fixes #560's URL issues
- Fixes _some_ of the isues in #556, though there are still some unknowns for the C++ errors there

### DIFF
--- a/examples/framework-hono/package.json
+++ b/examples/framework-hono/package.json
@@ -4,7 +4,8 @@
     "deploy": "wrangler deploy --minify src/index.ts"
   },
   "dependencies": {
-    "hono": "^4.2.7"
+    "hono": "^4.2.7",
+    "inngest": "^3.0.0"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20240403.0",
--- a/packages/inngest/src/hono.ts
+++ b/packages/inngest/src/hono.ts
@@ -36,7 +36,47 @@ export const serve = (options: ServeHandlerOptions) => {
         transformResponse: ({ headers, status, body }) => {
           return c.body(body, { headers, status });
         },
-        url: () => new URL(c.req.url, c.req.header("host")),
+        url: () => {
+          try {
+            // If this is an absolute URL, use it right now.
+            return new URL(c.req.url);
+          } catch {
+            // no-op
+          }
+
+          // We now know that `c.req.url` is a relative URL, so let's try
+          // to build a base URL to pair it with.
+          const host = options.serveHost || c.req.header("host");
+          if (!host) {
+            throw new Error(
+              "No host header found in request and no `serveHost` given either."
+            );
+          }
+
+          let baseUrl = host;
+          // Only set the scheme if we don't already have one, as a user may
+          // have specified the protocol in `serveHost` as a way to force it
+          // in their environment, e.g. for testing.
+          if (!baseUrl.includes("://")) {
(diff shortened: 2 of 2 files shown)


packages/inngest/src/hono.ts
```diff
@@ -35,11 +36,52 @@ export const serve = (options: ServeHandlerOptions) => {
...
+            baseUrl = `${scheme}://${baseUrl}`;
+          }
+
+          return new URL(c.req.url, baseUrl);
+        },
         queryString: (key) => c.req.query(key),
         headers: (key) => c.req.header(key),
         method: () => c.req.method,
         body: () => c.req.json(),
+        env: () => c.env as Env,
```
jpwilliams, line 84: This should stop folks needing to manually specifying signing/event keys.

examples/framework-hono/package.json
```diff
@@ -4,7 +4,8 @@
     "deploy": "wrangler deploy --minify src/index.ts"
   },
   "dependencies": {
-    "hono": "^4.2.7"
+    "hono": "^4.2.7",
+    "inngest": "^3.0.0"
```
jpwilliams, line 8: The example didn't include `inngest` 🫠

packages/inngest/src/hono.ts
```diff
@@ -35,11 +36,52 @@ export const serve = (options: ServeHandlerOptions) => {
         transformResponse: ({ headers, status, body }) => {
           return c.body(body, { headers, status });
         },
-        url: () => new URL(c.req.url, c.req.header("host")),
+        url: () => {
```
arjunyel, line 39: Would `canParse` help here? it's safe to use becasue anyone using Hono would be using a runtime that supports it. https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static

jpwilliams: Ah, awesome. Aye, at a glance it looks like Hono supports 18+. Thanks!

PR #597, opened 2024-06-06
Fix/incorrect custom fetch wrapping

jpwilliams:
## Summary

A warning log for when a custom fetch implementation is suspected to be broken is a bit too spammy.

```
A request failed when using a custom fetch implementation; this may be a misconfiguration. Make sure that your fetch client is correctly bound to the global scope.
```

This log is intended to be shown when we suspect that the given fetch implementation is invalid and seems to not be functioning correctly, for example if it has lost its binding after being passed in an object.

At present, however, it is logged upon every failure, even if it is correctly handled by code further up the stack.

This PR ensures we only log if it's not an expected fetch failure like failing to reach a server, which still indicates that the implementation is functional.

It also fixes an issue where the fetch implementation used by the `InngestCommHandler` is wrapped and treated as custom if no custom implementation has been given to `new Inngest()`.

## Checklist

### DIFF
--- a/.changeset/hip-steaks-pay.md
+++ b/.changeset/hip-steaks-pay.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix serve handler's `fetch` implementation incorrectly being marked as custom if no custom `fetch` is given to `new Inngest()`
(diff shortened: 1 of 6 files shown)


darwin67 (approved):

jpwilliams:
In draft waiting on confirmation from a user.

darwin67:
> In draft waiting on confirmation from a user.

confirmed from user

PR #610, opened 2024-06-10
Replace shell.nix with flake.nix

darwin67:
## Summary
Simplify and make sure dev dependencies are locked with flakes.


jpwilliams (approved):
ty!

Issue #468, opened 2024-01-17
Improve testing experience


MonsterDeveloper:
**Is your feature request related to a problem? Please describe.**
Setting up unit tests for Inngest functions is currently challenging due to a lack of guidelines in the official documentation. Exporting from `inngest.createFunction` requires exporting both the function itself and the function passed to it for testing.

**Describe the solution you'd like**
Enable the manual execution of declared functions. For example:
```ts
export const myFunction = inngest.createFunction(...)

myFunction.run()
```
Additionally, include testing guidelines in the docs with practical examples.

**Describe alternatives you've considered**
Exporting the handler callbacks passed into `inngest.createFunction` and testing them using mock objects.

brijmcq:
Would it be possible to expose the test helpers?

jpwilliams:
It's a mentioned issue, but commenting to highlight it.

@MonsterDeveloper just exposed `InngestFunction#createExecution()` over at #618, released in [3.19.16](https://github.com/inngest/inngest-js/pull/621).

Note that it's an internal method that can still change at any time; we'll add first-party testing tooling at a later date, but for now `createExecution()` is indeed what we use for a lot of our internal unit tests.

MonsterDeveloper:
Yep, now it's *possible* to test functions as follows. But beware that it's unstable and is subject to change without notice.

![CleanShot 2024-06-18 at 21 33 42@2x](https://github.com/inngest/inngest-js/assets/16190582/4295cf33-f68f-45e8-a724-fd221ddf498c)

jpwilliams:
Please see inngest/inngest#1680 for further conversation and progress on this. 🙂

- inngest/inngest#1680

jpwilliams:
[@inngest/test](https://github.com/inngest/inngest-js/tree/main/packages/test) is now available to provide some of these options.

Closing this as it was the original discussion around generic testing tools. Please raise issues/PRs against the `@inngest/test` package if you'd like further changes! 🙂

PR #618, opened 2024-06-16
Remove `private` modifier from `createExecution` method on `InngestFunction`

MonsterDeveloper:
## Summary

Removed `private` modifier from `createExecution` method on `InngestFunction` in order to allow calling created functions in the apps that use `inngest-js`.

---

When I was researching opportunities on testing my inngest functions, I found out that Inngest already has an option to call the create function that it uses internally for testing: `InngestFunction.createExecution`. It is used the the Inngest's own test files, e.g. [`InngestFunction.test.ts`](https://github.com/inngest/inngest-js/blob/baa9e5e63397943cd3c65896d7e908bf451e3c20/packages/inngest/src/components/InngestFunction.test.ts#L149)

It even seems to be working in my setup, if I ignore the type errors.

However, the `createExecution` method is not typed, resulting in very bad DX:
CleanShot 2024-06-16 at 19 24 36@2x

To add type definitions to this method, I removed the `private` modifier so that declarations for this method ship with the library, allowing developers to call/test created functions.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A still too raw to add docs for testing
- [ ] ~~Added unit/integration tests~~

## Related
#468

### DIFF
--- a/.changeset/quiet-pumas-relax.md
+++ b/.changeset/quiet-pumas-relax.md
@@ -1,0 +1,7 @@
+---
+"inngest": patch
+---
+
+Expose `InngestFunction#createExecution()` as a `protected` method to allow custom unit testing.
+
+Note that this is an internal API and can change at any time; first-party testing tools will be adde at a later date.
--- a/packages/inngest/src/components/InngestFunction.test.ts
+++ b/packages/inngest/src/components/InngestFunction.test.ts
@@ -12,8 +12,6 @@ import {
   NonRetriableError,
   type EventPayload,
 } from "@local";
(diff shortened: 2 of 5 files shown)


jpwilliams:
Thanks, @MonsterDeveloper.

I've shifted it to `protected` instead, as that should correctly generate the typing needed whilst still only allowing access via `["` instead of `.`.

MonsterDeveloper:
@jpwilliams good point. Made the changes.

jpwilliams (approved):

PR #625, opened 2024-06-25
Log invalid function config

goodoldneon:
## Summary
When there's an invalid function config, log it with `ERROR` severity

## Checklist

- [ ] ~Added unit/integration tests~ N/A

### DIFF
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -36,6 +36,7 @@ import { createStream } from "../helpers/stream";
 import { hashEventKey, hashSigningKey, stringify } from "../helpers/strings";
 import { type MaybePromise } from "../helpers/types";
 import {
+  functionConfigSchema,
   logLevels,
   type AuthenticatedIntrospection,
   type EventPayload,
@@ -1234,10 +1235,27 @@ export class InngestCommHandler<
   }
 
   protected configs(url: URL): FunctionConfig[] {
-    return Object.values(this.rawFns).reduce<FunctionConfig[]>(
+    const configs = Object.values(this.rawFns).reduce<FunctionConfig[]>(
       (acc, fn) => [...acc, ...fn["getConfig"](url, this.id)],
       []
     );
+
+    for (const config of configs) {
+      try {
+        functionConfigSchema.parse(config);
+      } catch (e) {
+        if (e instanceof z.ZodError) {
+          this.log(
+            "error",
+            `Config invalid for function "${config.id}" : ${e.errors
+              .map((err) => err.message)
+              .join("; ")}`
+          );
+        }
+      }
+    }
+
+    return configs;
   }
 
   /**
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -1241,17 +1241,14 @@ export class InngestCommHandler<
     );
 
     for (const config of configs) {
-      try {
-        functionConfigSchema.parse(config);
-      } catch (e) {
-        if (e instanceof z.ZodError) {
-          this.log(
-            "error",
-            `Config invalid for function "${config.id}" : ${e.errors
-              .map((err) => err.message)
-              .join("; ")}`
-          );
-        }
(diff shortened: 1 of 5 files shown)


packages/inngest/src/types.ts
```diff
@@ -1019,52 +1019,116 @@ export interface AuthenticatedIntrospection
...
+//     limit: number;
+//     period: TimeStr;
+//     burst?: number;
+//   };
+//   cancel?: {
+//     event: string;
+//     if?: string;
+//     timeout?: string;
+//   }[];
+// }
```
jpwilliams, line 1131: Can remove this

```suggestion
```

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1234,10 +1235,27 @@ export class InngestCommHandler<
...
+        if (e instanceof z.ZodError) {
+          this.log(
+            "error",
+            `Config invalid for function "${config.id}" : ${e.errors
+              .map((err) => err.message)
+              .join("; ")}`
+          );
+        }
+      }
+    }
```
jpwilliams, line 1256: Nit: Could use `.safeParse()` to avoid having to use do the `instanceof ZodError` check:
```suggestion
    for (const config of configs) {
      const check = functionConfigSchema.safeParse(config);
      if (!check.success) {
        const errors = check.error.errors.map((err) => err.message).join("; ");

        this.log(
          "error",
          `Config invalid for function "${config.id}" : ${errors}`
        );
      }
    }
```

packages/inngest/src/types.ts
```diff
@@ -1019,52 +1019,116 @@ export interface AuthenticatedIntrospection
...
+});
+
+export type FunctionConfig = z.output<typeof functionConfigSchema>;
+
 /**
  * A block representing an individual function being registered to Inngest
  * Cloud.
  *
  * @internal
  */
```
jpwilliams, line 1091: Can we move this comment back up?
```suggestion
/**
 * The shape of an individual function being synced with Inngest.
 *
 * @internal
 */
export type FunctionConfig = z.output<typeof functionConfigSchema>;
```

packages/inngest/src/types.ts
```diff
@@ -1019,52 +1019,116 @@ export interface AuthenticatedIntrospection
   signing_key_hash: string | null;
 }
 
+export const functionConfigSchema = z.strictObject({
```
jpwilliams, line 1022: Can we add a comment here as a reminder that this should be used for asserting the shape of the config, but not for feature compatibility, e.g. disallowing use of feature X with feature Y.
```suggestion
/**
 * The schema used to represent an individual function being synced with
 * Inngest.
 *
 * Note that this should only be used to validate the shape of a config object
 * and not used for feature compatibility, such as feature X being exclusive
 * with feature Y; these should be handled on the Inngest side.
 */
export const functionConfigSchema = z.strictObject({
```

packages/inngest/src/types.ts
```diff
@@ -1019,52 +1019,116 @@ export interface AuthenticatedIntrospection
...
+      key: z.string().optional(),
+      limit: z.number(),
+      period: z.string(),
+    })
+    .optional(),
+  throttle: z
+    .strictObject({
+      key: z.string().optional(),
+      limit: z.number(),
+      period: z.string(),
```
jpwilliams, line 1069: Can we keep the resulting type of this as a `TimeStr`? Something like this should be fine:
```suggestion
      period: z.string().transform((x) => x as TimeStr),
```

packages/inngest/src/types.ts
```diff
@@ -1019,52 +1019,116 @@ export interface AuthenticatedIntrospection
...
+    .strictObject({
+      maxSize: z.number(),
+      timeout: z.string(),
+    })
+    .optional(),
+  rateLimit: z
+    .strictObject({
+      key: z.string().optional(),
+      limit: z.number(),
+      period: z.string(),
```
jpwilliams, line 1062: Can we keep the resulting type of this as a `TimeStr`? Something like this should be fine:
```suggestion
      period: z.string().transform((x) => x as TimeStr),
```

jpwilliams (changes requested):

PR #639, opened 2024-07-04
Add `@inngest/middleware-remote-state`

jpwilliams:
## Summary

Adds `@inngest/middleware-remote-state`, giving consumers an easy way to push state to a remote store with a dataloader-like API.

We could also:

- Add async `encrypt` and `decrypt` functions to `@inngest/middleware-encryption` 
- Ensure all `encrypt` and `decrypt` calls are triggered and run in parallel
- Use `@inngest/middleware-encryption` in `@inngest/middleware-remote-state` to gather all keys, then provide usual dataloader-like API

## Checklist

### DIFF
--- a/packages/middleware-remote-state/.gitignore
+++ b/packages/middleware-remote-state/.gitignore
@@ -1,0 +1,2 @@
+dist
+node_modules
--- a/packages/middleware-remote-state/LICENSE.md
+++ b/packages/middleware-remote-state/LICENSE.md
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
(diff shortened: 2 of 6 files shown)


JonParton:
Hi Guys! 

Keeping a watch on this and loving the progress on the Cross Compatible Encryption Middleware too 🙌

I have a query on how this is planned to be implemented that I can't tell from the stub code so far. 

Is the plan to allow you to store some event data in remote state rather then all of it? Similar to the Encryption Middleware [Customizing Event Encryption Options](https://www.npmjs.com/package/@inngest/middleware-encryption?activeTab=readme#:~:text=Customizing%20event%20encryption)? 

This is to make sure that functionality in Inngest is still functional such as triggering based on a [CEL If Statement](https://www.inngest.com/docs/reference/functions/create#trigger:~:text=When%20using%20an%20event%20trigger%2C%20you%20can%20optionally%20combine%20it%20with%20the%20if%20option%20to%20filter%20events) that references non sensitive top level data, whilst storing all the Data Residency sensitive stuff in a sub `encrypted` or `sensitive` etc property of the event payload.

```ts
await inngest.send({
  name: "app/account.created",
  data: {
    billingPlan: "pro", // Non sensitive and could be used to trigger functions with an if statement etc
    sensitive: { // The sensitive data being Remote State Stored or Encrypted etc
        accountId: "645e9f6794e10937e9bdc201",
        name: "Dave Smith"
        }
  },
  user: {
    external_id: "645ea000129f1c40109ca7ad",
    email: "taylor@example.com",
  } // Presume if you were being PII Sensitive you just wouldn't include proper PII in the user property as Inngest uses it for delete me etc. 
})

```

Of course All Step Data should be remote from my understanding as it could have all sorts of data in there from variables, but selecting partial event data to be remote makes sense! 

Is there also wheels in motion to implement this Remote State middleware functionality in Python too? 🙈

jpwilliams:
Hi @JonParton! 👋

Awesome! Aye, the cross-language encryption middleware is teetering on the edge of release, which also opens up some pretty possibilities for building on top of it, like this middleware.

As you say, just like the encryption middleware, pushing _some_ data to a remote store (say, S3) and keeping other parts available to Inngest for CEL expressions would indeed be one of the features of this middleware too.

It can be used for some of the same purposes as the E2E encryption middleware (protecting PII), but may also complement it and be used alongside encryption, perhaps for data residency purposes or for steps with very large outputs (generated files, for example) that may be best stored somewhere external.

No immediate plans yet for Python, but the aim is that this is a relatively thin layer on top of the base we build for `@inngest/middleware-encryption`; carrying it across to Python would hopefully be a small lift to write once we've settled on the APIs.

I believe you're using the encryption middleware already? I'd love to know:

- Do you use your own encryption service or the default provided by the package?
- Are you customizing which fields are encrypted outside of the default `data.encrypted`?

JonParton:
Hi @jpwilliams !

So we are not actually using the Encryption middleware just yet as we have some backend services in both Python and Typescript written by different teams, so until they hit parity it was a no go. However, the team is just working to change Event Schema's in preparation for using both of these middleware's. 

> Are you customizing which fields are encrypted outside of the default `data.encrypted`?

Our current idea is to structure it as the example above (and repeated below): 
```ts
await inngest.send({
  name: "app/account.created",
  data: {
    billingPlan: "pro", // Non sensitive and could be used to trigger functions with an if statement etc
    sensitive: { // The sensitive data being Remote State Stored or Encrypted etc
        accountId: "645e9f6794e10937e9bdc201",
        name: "Dave Smith"
     },
  },
  user: {
    external_id: "645ea000129f1c40109ca7ad",
  } // Presume if you were being PII Sensitive you just wouldn't include proper PII in the user property as Inngest uses it for delete me etc. 
})
```

Basically making it so we store all of the values that need encrypting (Or remote storing) in the sub `data.sensitive:` prop. We have gone with this over `data.encrypted` as depending on the current deployments need (We will have multiple environments for different clients) we will use either encryption for security, or the Remote State Storage to help with Data Residency. In this case the more generic `data.sensitive` makes more sense here and is where we will put any PII we are going to deal with. 

> Do you use your own encryption service or the default provided by the package?

Plan would be to use the default provided unless there is advise otherwise! Main thing is to make sure there is compatibility between the language SDK's! 


> No immediate plans yet for Python, but the aim is that this is a relatively thin layer on top of the base we build for @inngest/middleware-encryption; carrying it across to Python would hopefully be a small lift to write once we've settled on the APIs.

We would definitely love the state store middleware to be available in Python too 🙏 Our first strongly data residency sensitive client will need cross compatibility of this in 2-3 months, but will be using the TS version sooner. We can of course patch ourselves if the Middleware hooks are all available in python, but having it out of the box that we may be able to contribute back to would be the preference! 💖

-------------

Thanks for all the work on this Jack! And also for nailing one of the other ideas I discussed briefly with Tony the other day too around actually [using the Zod Schemas for validation](https://github.com/inngest/inngest-js/pull/657) without me even realising it was in flight! 🚀

PR #669, opened 2024-08-01
Update no-nested-steps.ts

swiecki:
Today we had an interesting and expensive learning experience- we had pushed step.sendEvent inside step.run. As part of the retrospective I noticed the eslint plugin seemed to only be picking up step.run inside step.run, not step.invoke or step.sendEvent etc.

UPDATED: eslint-plugin/src/rules/no-nested-steps.ts

- Updated rule to disallow use of any `step.*` within another `step.*` function
- Added check to report error if `step.*` is used within another `step.*` function
- issue with counting the depth of nested `step.run()` calls

## Checklist

### DIFF
--- a/.changeset/heavy-files-try.md
+++ b/.changeset/heavy-files-try.md
@@ -1,0 +1,5 @@
+---
+"@inngest/eslint-plugin": patch
+---
+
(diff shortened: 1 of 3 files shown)


jpwilliams:
Thanks, @swiecki! Apologies this caught you out.

PR #672, opened 2024-08-05
Set sentry transaction name according to inngest function

mattddean:
## Summary
Set transaction name in sentry according to inngest function name so that it's immediately obvious in sentry's UI which function had the error:

Screenshot 2024-08-05 at 11 53 08 AM

I tried calling `setTransactionName` next to `scope.setTags(sharedTags);` on line 84 but the transaction name was not appropriately set when I threw an error in my function.

## Checklist

## Related

- INN-

### DIFF
--- a/.changeset/shy-goats-heal.md
+++ b/.changeset/shy-goats-heal.md
@@ -1,0 +1,5 @@
+---
+"@inngest/middleware-sentry": patch
+---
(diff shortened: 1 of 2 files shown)


jpwilliams:
Thanks, @mattddean! 🙌

PR #674, opened 2024-08-07
Add trust probes and response signing

jpwilliams:
## Summary

Adds trust probes, capabilities to SDK introspection, and response signing.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Internal change
- [x] Added changesets if applicable (patch)

## Related

- inngest/inngest-py#142

### DIFF
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -677,16 +677,48 @@ export class InngestCommHandler<
           },
         });
 
-      const signature = await actions.headers(
-        "checking signature for request",
-        headerKeys.Signature
-      );
+      const assumedMode = getMode({ env: this.env, client: this.client });
+
+      if (assumedMode.isExplicit) {
+        this._mode = assumedMode;
+      } else {
+        const serveIsProd = await actions.isProduction?.(
+          "starting to handle request"
+        );
+        if (typeof serveIsProd === "boolean") {
+          this._mode = new Mode({
+            type: serveIsProd ? "cloud" : "dev",
+            isExplicit: false,
+          });
+        } else {
+          this._mode = assumedMode;
+        }
+      }
+
+      this.upsertKeysFromEnv();
+
+      const methodP = actions.method("starting to handle request");
+
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const body = await actions.body("checking body for request signing");
-      const signatureValidation = this.validateSignature(
-        signature ?? undefined,
-        body
-      );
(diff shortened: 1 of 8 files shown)


packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -677,31 +677,102 @@ export class InngestCommHandler<
...
+          this._mode = new Mode({
+            type: serveIsProd ? "cloud" : "dev",
+            isExplicit: false,
+          });
+        } else {
+          this._mode = assumedMode;
+        }
+      }
+
+      this.upsertKeysFromEnv();
```
jpwilliams, line 698: Fetching modes (and late actions like using environment variables passed in the request) is moved much earlier, so we can appropriately validate this early too.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1476,39 +1561,59 @@ export class InngestCommHandler<
...
+    }
   }
 
-  protected signResponse(): string {
-    return "";
+  protected getResponseSignature(key: string, body: string): string {
+    const now = Date.now();
+    const mac = signDataWithKey(body, key, now.toString());
+
+    return `s=${mac}&t=${now}`;
```
goodoldneon, line 1644:
```suggestion
    return `t=${now}&s=${mac}`;
```

Need to swap these because our backend compares by string, rather than parsed query params 😢

packages/inngest/src/types.ts
```diff
@@ -991,6 +991,10 @@ export interface RegisterRequest {
   deployId?: string;
 }
 
+export interface IntrospectionCapabilities {
```
goodoldneon, line 994:
```suggestion
export interface Capabilities {
```

Rename because capabilities are reported in both introspection and the sync request

packages/inngest/src/components/InngestCommHandler.ts
jpwilliams: @goodoldneon Does this need to change too?

https://github.com/inngest/inngest-py/blob/a760b9a7a8bdfd0d7538b3ccb9dbef8e3c9f877f/inngest/_internal/net.py#L276

goodoldneon: Yea, that fix is in my open trust probe PR:
https://github.com/inngest/inngest-py/pull/142/files#diff-0c9c1f402f1ba4641478323e51a10a386eed3476c1c3c022f6437a512ae8f085R278

goodoldneon (approved):

PR #682, opened 2024-08-20
Fix empty response body when streaming

goodoldneon:
## Summary
Fix a bug where streaming would send back an empty body. This happened because we stringified an unawaited `Promise` within `createStream`

## Checklist

### DIFF
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -814,7 +814,7 @@ export class InngestCommHandler<
            * appropriate response is always given.
            */
           void actionRes.then((res) => {
-            void prepareActionRes(res).then(finalize);
+            return finalize(prepareActionRes(res));
           });
 
           return timer.wrap("res", () => {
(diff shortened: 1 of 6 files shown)


packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -814,7 +814,7 @@ export class InngestCommHandler<
            * appropriate response is always given.
            */
           void actionRes.then((res) => {
-            return finalize(prepareActionRes(res));
+            void prepareActionRes(res).then(finalize);
```
tonyhb, line 817: You knwo what I'm about to say!  Haha.  Can we add tests so that this won't regress? 🙏

BrunoScheufler, line 817: How did TypeScript not catch this 😭

goodoldneon: `finalize` accepts an `unknown` type 😢

jpwilliams: Added integration tests that use streaming for supported platforms. 👍

PR #702, opened 2024-09-07
Add support for in-band syncing

jpwilliams:
## Summary

Adds in-band sync support by returning different data based on whether a `PUT` request is signed (and validated) or not.

In draft while functionality is built out elsewhere.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Accounted for elsewhere

## Related

- INN-3306


packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1451,6 +1441,134 @@ export class InngestCommHandler<
...
+    }
+
+    const body: InBandRegisterRequest = {
+      app_id: this.client.id,
+      capabilities: registerBody.capabilities,
+      env: introspectionBody.env,
+      framework: registerBody.framework,
+      functions: registerBody.functions,
+      inspection: introspectionBody,
+      platform: "?",
```
goodoldneon, line 1480: Is it possible to populate this?

jpwilliams: Baha yes, this is prior to our discussion. Will change.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1199,14 +1136,75 @@ export class InngestCommHandler<
...
+            .then((kind) => {
+              return kind === syncKind.InBand;
+            }),
+        ]);
+
+        if (inBandSyncRequested) {
+          // Validation can be successful if we're in dev mode and did not
+          // actually validate a key. In this case, also check that we did indeed
+          // use a particular key to validate.
+          const sigCheck = await signatureValidation;
```
jpwilliams, line 1171: This is always true if the SDK's mode is `"dev"`, but will fail if it's `"cloud"` and the request could not be validated.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1199,14 +1136,75 @@ export class InngestCommHandler<
...
+
+          if (!sigCheck.success) {
+            return {
+              status: 401,
```
jpwilliams, line 1175: A failed validation when we are allowing an in-band sync and one has been requested is always a `401`.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1199,14 +1136,75 @@ export class InngestCommHandler<
...
+              body: stringify({
+                code: "sig_verification_failed",
+                message: "",
+                name: "SigVerificationFailedError",
```
jpwilliams, line 1179: @amh4r I used the same JSON as the Python SDK; not sure if this code is used anywhere else yet.

amh4r: I don't think it's used anywhere else. We should be consistent with the error code across SDKs but the name and message can be excluded

amh4r (approved):

PR #704, opened 2024-09-07
Initial `@inngest/test` package

jpwilliams:
## Summary

The initial bare-bones release of `@inngest/test` based on the RFC at inngest/inngest#1680 following the preliminary changes to open up the execution APIs in #688.

To see how to use the library in its current state, see [`README.md`](https://github.com/inngest/inngest-js/tree/feat/inngest-test/packages/test).

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Package has a readme - will follow up with docs
- [ ] ~Added unit/integration tests~ N/A Package still `0.` - will pad out utests later

## Related

- Follows #688
- Based on RFC at inngest/inngest#1680
- Python: inngest/inngest-py#152


djfarrelly (changes requested):
User reported issue using this with vitest where Jest complains: 
```
Do not import `@jest/globals` outside of the Jest test environment
```

There may be a couple ways to fix this:
1. Add [`jest-mock`](https://www.npmjs.com/package/jest-mock) to `dependencies` (currently Jest is in dev deps only) and use that as a standalone package that is agnostic to the runner. Also needs to remove the internal `expect`.
2. Use a different mocking implementation like vitest's `@vitest/spy`

amh4r (approved):

djfarrelly (approved):

PR #707, opened 2024-09-10
Fix probe response sig with streaming

goodoldneon:
## Summary
Fix no response signature header when streaming is enabled.

## Checklist

## Related

- INN-


jpwilliams (approved):
Thanks! I refactored the `getQueryString()` usage a touch - it was already spread everywhere prior to this PR.

PR #745, opened 2024-11-13
Do not reuse the response body when parsing errors

### LINKED ISSUE #586: [BUG] TypeError: Body has already been used
**Describe the bug**
We are seeing error logs coming from Inngest thrown in `getResponseError` because the body has already been used. I do not know what the root error is, and is happening rarely, but I am a bit afraid this can lead to loss of data because we rely on inngest to be accepting events at all times. The full error is:
```
TypeError: Body has already been used. It can only be used once. Use tee() first if you need to read it twice. at Gl.getResponseError (index.js:25:50970) at Gl._send (index.js:25:53022) at async index.js:25:315335 at async Promise.all (index 0) at async index.js:25:315291 at async Promise.all (index 0) at async index.js:25:315244 at async i (index.js:25:229977) at async i (index.js:25:229977) at async index.js:25:246409
```

**To Reproduce**
Repeatedly send events to inngest. We are running in cf worker.

jpwilliams:
## Summary

Fixes some unlucky reuse of the `Response` body when attempting to parse errors from the `inngest.send()` response.

## Checklist

## Related

- Fixes #586


jacobheric (approved):

tonyhb (approved):
good catch

PR #762, opened 2024-11-26
`ai.infer`: anthropic types and adapters

tonyhb:
This add types from the Anthropic SDK to support Anthropic models.


jacobheric (approved):
You have some type declarations linting and some openai comments and vars on the new stuff, but otherwise looks good to me!

jpwilliams (approved):
A note here - we support Node 14+ and TS 4.7+ across both CJS and ESM, which hugely restricts our ability to utilize other libraries if they have a more limited toolchain, like `@anthropic-ai/sdk`.

More critically, we support ES5, meaning our compiled output must be compatible with that low target and any changes to update our bundling could break builds for downstream users.

This support means we're far behind in terms of language features we can use _and_ other libraries we can use, even just for types.

For v4, we can cut a lot of this support and enforce we're using much more modern standards across the board. This PR frustratingly requires that we manually use the types which hugely increases ongoing maintenance; doing this breaking change to upgrade our bundling would save a tonne of time.

PR #773, opened 2024-12-07
Fix Nuxt and H3 use https when no NODE_ENV

amh4r:
## Summary
Fix the SDK reporting its scheme as `https` when the `NODE_ENV` env var isn't set.


jpwilliams (approved):
Thank you!

PR #776, opened 2024-12-10
Experimental `getAsyncCtx()`

jpwilliams:
## Summary

Adds a new `getAsyncCtx()` function exported from `"inngest/experimental"` that will attempt to retrieve the function input arguments.

- Uses `@inngest/test@workspace:^` internally, exclusively for the `inngest` package. This can be a bit weird, as `inngest` is a peer dep of `@inngest/test` so typing can be flaky.
- Altered exports of `@inngest/test` to have simpler compilation and publishing.
- Added the experimental `getAsyncCtx()`

```ts
import { getAsyncCtx } from "inngest/experimental";

const ctx = await getAsyncCtx();
```

### Questions

- [x] Should we remove the store for a particular execution context when the run completes?
No. It will still be garbage collected and info may still be useful.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Experimental for now

### DIFF
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -36,7 +36,13 @@ runs:
       if: ${{ inputs.install-dependencies == 'true' }}
       run: pnpm install
       shell: bash
-      working-directory: ${{ inputs.working-directory }}/packages/inngest
+      working-directory: ${{ inputs.working-directory }}
+
+    - name: Build test dependencies
+      if: ${{ inputs.install-dependencies == 'true' }}
+      run: pnpm run build
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}/packages/test
 
     - name: Build
       if: ${{ inputs.build == 'true' }}
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -49,9 +49,3 @@ runs:
       run: pnpm run build
       shell: bash
       working-directory: ${{ inputs.working-directory }}/packages/inngest
-
-    - name: Install dependencies to link
-      if: ${{ inputs.install-dependencies == 'true' }}
-      run: pnpm install
-      shell: bash
-      working-directory: ${{ inputs.working-directory }}
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -49,3 +49,9 @@ runs:
       run: pnpm run build
       shell: bash
       working-directory: ${{ inputs.working-directory }}/packages/inngest
+
+    - name: Install dependencies to link
+      if: ${{ inputs.install-dependencies == 'true' }}
+      run: pnpm install
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
--- a/packages/inngest/jest.config.js
+++ b/packages/inngest/jest.config.js
@@ -6,7 +6,7 @@ module.exports = {
   roots: ["<rootDir>/src"],
   moduleNameMapper: {
     "(\\..+)\\.js": "$1",
-    inngest: "<rootDir>/src",
+    "^inngest$": "<rootDir>/src",
     "^@local$": "<rootDir>/src",
     "^@local/(.*)": "<rootDir>/src/$1",
     "^@local/(.*)\\.js": "<rootDir>/src/$1",
--- a/packages/inngest/jsr.json
+++ b/packages/inngest/jsr.json
@@ -17,6 +17,7 @@
   ],
   "exports": {
     ".": "./src/index.ts",
+    "./experimental": "./src/experimental.ts",
     "./astro": "./src/astro.ts",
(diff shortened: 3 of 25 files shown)


.github/actions/setup-and-build/action.yml
```diff
@@ -36,7 +36,13 @@ runs:
...
       run: pnpm install
       shell: bash
-      working-directory: ${{ inputs.working-directory }}/packages/inngest
+      working-directory: ${{ inputs.working-directory }}
+
+    - name: Build test dependencies
+      if: ${{ inputs.install-dependencies == 'true' }}
+      run: pnpm run build
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}/packages/test
```
jpwilliams, line 45: `inngest` now relies on `@inngest/test@workspace:^`, so we make sure that's built during CI too.

packages/inngest/jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   roots: ["<rootDir>/src"],
   moduleNameMapper: {
     "(\\..+)\\.js": "$1",
-    inngest: "<rootDir>/src",
+    "^inngest$": "<rootDir>/src",
```
jpwilliams, line 9: This regex was picking up anything containing `inngest` and redirecting it to `src/` - bit of a problem when importing `@inngest/test`!

packages/inngest/jsr.json
```diff
@@ -17,6 +17,7 @@
   ],
   "exports": {
     ".": "./src/index.ts",
+    "./experimental": "./src/experimental.ts",
```
jpwilliams, line 20: A new entrypoint to mark out experimental APIs.

packages/inngest/src/components/execution/v1.ts
```diff
@@ -459,44 +460,48 @@ class V1InngestExecution extends InngestExecution implements IInngestExecution {
    * and middleware hooks where appropriate.
    */
   private async startExecution(): Promise<void> {
-    /**
-     * Mutate input as neccessary based on middleware.
-     */
-    await this.transformInput();
+    return getAsyncLocalStorage().then((als) =>
+      als.run({ ctx: this.fnArg }, async (): Promise<void> => {
```
jpwilliams, line 464: When executing, we now wrap the entire operation in our `AsyncLocalStorage.run()` call. This also ensures that middleware is included in the async chain.

packages/inngest/src/components/execution/als.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Get a singleton instance of `AsyncLocalStorage` used to store and retrieve
+ * async context for the current execution.
+ */
+export const getAsyncLocalStorage = async (): Promise<AsyncLocalStorageIsh> => {
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises, no-async-promise-executor
+  als ??= new Promise<AsyncLocalStorageIsh>(async (resolve) => {
+    try {
+      const { AsyncLocalStorage } = await import("node:async_hooks");
```
jpwilliams, line 36: Bun, Deno, and Node all expose this as `node:async_hooks`.

Cloudflare Workers requires a compatibility flag of `node_compat` or `node_compat_v2`. A singular `node_als` should be possible but doesn't currently work with any version of Wrangler.

In the case that these flags have not been specified, this `AsyncLocalStorage` instance will essentially become a no-op; no store will ever be returned and `.run()` will just run the function given directly.

packages/inngest/src/components/execution/als.ts
```diff
@@ -0,0 +1,52 @@
...
+ * A local-only variable to store the async local storage instance.
+ */
+let als: Promise<AsyncLocalStorageIsh> | undefined;
+
+/**
+ * Retrieve the async context for the current execution.
+ */
+export const getAsyncCtx = async (): Promise<AsyncContext | undefined> => {
+  return getAsyncLocalStorage().then((als) => als.getStore());
+};
```
jpwilliams, line 26: This is the only piece exposed to the public API, used to retrieve the input to the current function run.

packages/inngest/src/components/execution/als.ts
```diff
@@ -0,0 +1,52 @@
+import { type Context } from "../../types.js";
+
+export interface AsyncContext {
+  ctx: Context.Any;
+}
```
jpwilliams, line 5: If found, this is the value returned when a developer calls `getAsyncCtx()` in userland code. It will likely expand to contain more information over time.

packages/inngest/src/components/execution/als.ts
jacobheric: Nice! 

On a side note, how many lines of code would it take you to do this in go!?

` als ??= new Promise<AsyncLocalStorageIsh>(async (resolve) => ...`

jacobheric (approved):

tonyhb (approved):
This is AMAZING work!!!

PR #831, opened 2025-01-31
Read signing key from env var

BrunoScheufler:
## Summary

This adds fallback handling for retrieving the signing key for connect.

## Checklist

## Related

- INN-

### DIFF
--- a/packages/inngest/src/components/connect/index.ts
+++ b/packages/inngest/src/components/connect/index.ts
@@ -1,7 +1,7 @@
 import { WaitGroup } from "@jpwilliams/waitgroup";
 import debug, { type Debugger } from "debug";
 import { ulid } from "ulid";
-import { headerKeys, queryKeys } from "../../helpers/consts.js";
+import { envKeys, headerKeys, queryKeys } from "../../helpers/consts.js";
 import { allProcessEnv, getPlatformName } from "../../helpers/env.js";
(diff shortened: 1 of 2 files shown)


packages/inngest/src/components/connect/index.ts
```diff
@@ -141,6 +141,12 @@ class WebSocketWorkerConnection implements WorkerConnection {
     if (!Array.isArray(options.handleShutdownSignals)) {
       options.handleShutdownSignals = DEFAULT_SHUTDOWN_SIGNALS;
     }
+
+    const env = allProcessEnv();
+    options.signingKey = options.signingKey ?? env[envKeys.InngestSigningKey];
+    options.signingKeyFallback =
+      options.signingKeyFallback ?? env[envKeys.InngestSigningKeyFallback];
```
jpwilliams, line 148: Nit: This should probably be `||` instead of `??` so that we look at the environment if someone passes an empty key like `signingKey: ""`.
```suggestion
    options.signingKey = options.signingKey || env[envKeys.InngestSigningKey];
    options.signingKeyFallback =
      options.signingKeyFallback || env[envKeys.InngestSigningKeyFallback];
```

jpwilliams (approved):

packages/inngest/src/components/connect/index.ts
BrunoScheufler: Good point!

PR #848, opened 2025-02-19
Fix serve ID not considered for in-band syncs

amh4r:
## Summary
Fix serve ID not considered for in-band syncs.


jpwilliams (approved):
Should we also make sure this is aligned for introspections too?

https://github.com/inngest/inngest-js/blob/903a85127b5fa3e3d0e34f62baa9883dd7d21b56/packages/inngest/src/components/InngestCommHandler.ts#L1703

PR #851, opened 2025-02-20
Allow multi-app connections

BrunoScheufler:
## Checklist

## Related

- INN-

### DIFF
--- a/packages/inngest/src/components/connect/index.ts
+++ b/packages/inngest/src/components/connect/index.ts
@@ -51,6 +51,11 @@ const WorkerHeartbeatInterval = 10_000;
 interface connectionEstablishData {
   marshaledCapabilities: string;
   manualReadinessAck: boolean;
+  apps: {
+    appName: string;
+    appVersion?: string;
+    functions: Uint8Array;
+  }[];
 }
 
 const ConnectWebSocketProtocol = "v0.connect.inngest.com";
@@ -273,8 +278,6 @@ class WebSocketWorkerConnection implements WorkerConnection {
       connect: "v1",
     };
 
-    const appName = this.inngest.id;
-
     const functionConfigs: Record<
       string,
       {
@@ -290,7 +293,7 @@ class WebSocketWorkerConnection implements WorkerConnection {
         functions: functions.flatMap((f) =>
           f["getConfig"]({
             baseUrl: new URL("wss://connect"),
-            appPrefix: appName,
+            appPrefix: client.id,
             isConnect: true,
           })
         ),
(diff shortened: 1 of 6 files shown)


packages/inngest/src/components/connect/index.ts
```diff
@@ -129,12 +138,27 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
+        client: Inngest.Like;
+        functions: InngestFunction.Any[];
+      }
+    > = {};
+    for (const app of this.options.apps) {
+      functions[app.client.id] = {
+        client: app.client,
+        functions: (app.functions as InngestFunction.Any[]) ?? [],
+      };
+    }
```
jpwilliams, line 160: We could find that some app IDs collide here; maybe we should log a warning if this happens.

packages/inngest/src/components/connect/index.ts
```diff
@@ -129,12 +138,27 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
-    return (
-      (this.options.functions as InngestFunction.Any[]) ??
-      this.inngest["localFns"] ??
-      []
-    );
+  private get functions(): Record<
+    string,
+    {
+      client: Inngest.Like;
+      functions: InngestFunction.Any[];
```
jpwilliams, line 145: Could this also be `InngestFunction.Like[]`? Or do you need some of the private accesses?

packages/inngest/src/components/connect/index.ts
```diff
@@ -254,126 +278,158 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
+      this.functions
+    )) {
+      functionConfigs[appId] = {
+        client: client,
+        functions: functions.flatMap((f) =>
+          f["getConfig"]({
+            baseUrl: new URL("wss://connect"),
+            appPrefix: client.id,
+            isConnect: true,
+          })
```
jpwilliams, line 298: If you're feeling generous, a `getConnectConfig()` method that returns the correct data instead of adapting `getConfig()` would be awesome - `getConfig()` is soo old 😄

jpwilliams (commented):
An alternate API could be:
```ts
connect({
  client: inngest, // contains critical config such as signing key
  apps: [{
    appId: "foo",
    functions: [...],
  }, {
    appId: "bar",
    functions: [...],
  }],
});
```
This meets the same fate as we've seen elsewhere, though, where direct import `step.invoke()`s now fail as they never have the correct app ID.

In the future it feels likely we'll change invoke to be app-based instead of function-based and remove app IDs from functions altogether.

For now, it means we're probably better tackling the current quirks rather than switching to this.

packages/inngest/src/components/connect/index.ts
```diff
@@ -254,126 +278,158 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
+            },
+            isProduction: () => {
+              try {
+                // eslint-disable-next-line @inngest/internal/process-warn
+                const isProd = process.env.NODE_ENV === "production";
+                return isProd;
+              } catch (err) {
+                // no-op
+              }
+            },
```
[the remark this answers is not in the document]
BrunoScheufler: Sure I can remove that 😄

packages/inngest/src/components/connect/index.ts
BrunoScheufler: getConfig() 😭

jpwilliams (approved):

PR #864, opened 2025-02-26
Replace `ulid` with `ulidx`

### LINKED ISSUE #863: Edge Runtime Compatibility Issue
**Describe the bug**
When using Inngest with Next.js Edge Runtime, I encounter an error because Edge Runtime doesn't support Node.js 'crypto' module, which appears to be a dependency of the Inngest SDK. Version 3.30.0 works correctly in Next.js Edge Runtime, but the latest version 3.31.11 fails with a crypto module error.

**To Reproduce**
1. Create a Next.js API route with Edge Runtime enabled:

```typescript
import { sentryMiddleware } from '@inngest/middleware-sentry';
import { Inngest } from 'inngest';
import type { NextRequest } from 'next/server';

export const runtime = 'edge';
export const preferredRegion = ['cle1'];

const inngest = new Inngest({
  id: 'platform-workflows',
  middleware: [sentryMiddleware()],
});

export async function POST(request: NextRequest) {
  const event = await request.json();
  await inngest.send({
    name: 'storyblok/story.${event.action}',
    data: event,
  });
  
  return new Response('OK');
}
```

2. With Inngest 3.31.11, making a POST request to this endpoint fails
3. The request fails with an error about the crypto module

**Expected behavior**
Inngest SDK should work in Edge Runtime environments, or there should be clear documentation about Edge Runtime compatibility limitations and potential workarounds.

**Screenshots / Stack trace dump**
```
 ⨯ Error: The edge runtime does not support Node.js 'crypto' module.
Learn More: https://nextjs.org/docs/messages/node-module-in-edge-runtime
    at Array.map (<anonymous>)
    at async POST (app/api/webhooks/storyblok/route.ts:13:4)
  11 |
  12 |   if (event.text.includes('...')) {
> 13 |     await inngest.send({
     |    ^
  14 |       name: `storyblok/story.${event.action}`,
  15 |       data: event,
  16 |     });

```

**System info**
- inngest-cli version: 1.4.6
- OS: macOS 15.3.1
- Inngest versions tested: 3.30.0 (works), 3.31.11 (fails)

**Additional context**
None.

jpwilliams:
## Summary

We use [`ulid`](https://www.npmjs.com/package/ulid) to generate some internal IDs, including when sending events in order to appropriately idempotently retry.

The `ulid` library, however, is dated. Most notably, when attempting to find a `crypto` module in order to securely generate random data, it checks for `window.crypto` (to support browsers), but not `global.crypto` or `globalThis.crypto` which are available in edge runtimes.

https://github.com/ulid/javascript/blob/b269f93944fe41a3c2798c3d883a33c7a2d4bf19/lib/index.ts#L118-L146

This creates an issue whereby edge runtimes do not find `crypto` in the global namespace, so fall back to attempting to import `node:crypto` and subsequently failing.

The `ulid` library does export a `factory()` that can be used to pass in a custom crypto module, meaning we could easily discover it ourselves. However, the [`ulidx`](https://www.npmjs.com/package/ulidx) effort also exists which handles these and other compatibility issues since `ulid` became unmaintained.

We'll switch to `ulidx` for now to resolve the issue, as this does a much more thorough job of discovering the `crypto` module:

https://github.com/perry-mitchell/ulidx/blob/5043c511406fb9b836ddf126583c80ffb90cbb73/source/ulid.ts#L70-L114

> [!NOTE]
> The creator of `ulidx` has been added as a collaborator to [ulid/javascript](https://github.com/ulid/javascript) repo with the intent to merge the efforts back into the `ulid` library. 🥳 

## Checklist

- [ ] ~Added unit/integration tests~ N/A Deploy env specific
- [x] Wait for fix confirmation for #863 

## Related

- Fixes #863 
- Hopefully `ulidx` is merged into `ulid` - ulid/javascript#113

### DIFF
--- a/.changeset/beige-yaks-reply.md
+++ b/.changeset/beige-yaks-reply.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix event sending failing in some edge environments due to not finding `global.crypto` or `globalThis.crypto` when creating idempotency IDs
--- a/packages/inngest/package.json
+++ b/packages/inngest/package.json
@@ -212,7 +212,7 @@
     "ms": "^2.1.3",
     "serialize-error-cjs": "^0.1.3",
     "strip-ansi": "^5.2.0",
-    "ulid": "^2.3.0",
+    "ulidx": "^2.4.1",
     "zod": "~3.22.3"
   },
   "devDependencies": {
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -1,4 +1,4 @@
-import { ulid } from "ulid";
+import { ulid } from "ulidx";
 import { InngestApi } from "../api/api.js";
 import {
   defaultDevServerHost,
--- a/packages/inngest/src/components/connect/index.ts
+++ b/packages/inngest/src/components/connect/index.ts
@@ -1,6 +1,6 @@
 import { WaitGroup } from "@jpwilliams/waitgroup";
 import debug, { type Debugger } from "debug";
-import { ulid } from "ulid";
+import { ulid } from "ulidx";
 import { envKeys, headerKeys, queryKeys } from "../../helpers/consts.js";
 import { allProcessEnv, getPlatformName } from "../../helpers/env.js";
 import { parseFnData } from "../../helpers/functions.js";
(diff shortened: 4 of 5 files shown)


kang8 (approved):

jpwilliams:
Fix confirmed at https://github.com/inngest/inngest-js/issues/863#issuecomment-2686972704.

BrunoScheufler (approved):

PR #884, opened 2025-03-11
`@inngest/ai`: Gemini adapter + Grok OpenAI-compatible support

charlypoly:
- **Gemini**: does not support tool calling via their OpenAI compatible API -> full adapter
- **Grok**: OpenAI compatible API but does not support strict parameters with function calls -> partial adapter on top of openai

### DIFF
--- a/.changeset/new-ghosts-train.md
+++ b/.changeset/new-ghosts-train.md
@@ -1,0 +1,5 @@
+---
+"@inngest/ai": minor
+---
+
+`@inngest/ai`: Gemini adapter + Grok OpenAI-compatible support
--- a/.changeset/new-ghosts-train.md
+++ b/.changeset/new-ghosts-train.md
@@ -1,5 +1,5 @@
 ---
-"@inngest/ai": minor
+"@inngest/ai": patch
 ---
 
 `@inngest/ai`: Gemini adapter + Grok OpenAI-compatible support
--- a/packages/ai/src/adapters/index.ts
+++ b/packages/ai/src/adapters/index.ts
@@ -1,5 +1,4 @@
 export * from "./anthropic.js";
 export * from "./gemini.js";
 export * from "./openai.js";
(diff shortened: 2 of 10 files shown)


packages/ai/src/models/gemini.ts
```diff
@@ -12,40 +13,63 @@ export const gemini: AiAdapter.ModelCreator<
...
-  const model = options.model as OpenAi.Model;
+  if (!baseUrl.endsWith("/")) {
+    baseUrl += "/";
+  }
 
-  return openai({
-    ...options,
-    apiKey,
+  const url = new URL(
+    `models/${options.model}:generateContent?key=${authKey}`,
```
charlypoly, line 27: cf: https://ai.google.dev/api/generate-content

packages/ai/src/models/grok.ts
```diff
@@ -0,0 +1,68 @@
...
+  const model = options.model as Grok.Model;
+
+  const adapter = openai({
+    ...options,
+    apiKey,
+    baseUrl,
+    model,
+  }) as unknown as Grok.AiModel;
+
+  adapter.format = "grok";
```
charlypoly, line 28: the adapter in AgentKit reuse 99% of OpenAI

packages/ai/src/adapters/index.ts
jpwilliams, line 1: Should we export the `GrokAiAdapter` here too?

.changeset/new-ghosts-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inngest/ai": minor
```
jpwilliams, line 2: Nit, but as this is `0.x.y` atm should this be a `patch`? I think `minor` means a breaking change in `0.x.y`?

jpwilliams (approved):

PR #920, opened 2025-03-24
v4

jpwilliams:
## Summary

- TypeScript 5.8+
- Node 20+
- vitest
- tsup
- biome

## Checklist


wiesson:
FYI, I have switched to this PR due to vite/sveltekit/esm compatibility and it works quite well in prod 😅

punkpeye:
Is it safe to start using this in production?

It's a great PR, but waiting for v4 compatibility the most.

jpwilliams:
> @punkpeye: Is it safe to start using this in production?
> 
> It's a great PR, but waiting for v4 compatibility the most.

Yeah this should be pretty safe. It's v4 but this will end up being a non-breaking change!

punkpeye:
Would it be a possible to release version of the latest changes as beta on npm?

jpwilliams:
> @punkpeye: Would it be a possible to release version of the latest changes as beta on npm?

Yeah! The release bot here keeps a tag up to date: https://github.com/inngest/inngest-js/pull/920#issuecomment-2747771334

```sh
npm install inngest@pr-920
```

punkpeye:
```diff
inngest.send({
  name: 'glama/chat-session.created',
-  data: { chatSessionReferenceId: '123', },
+  chatSessionReferenceId: '123',
})
```

Am I doing something wrong or did the data get flattened into the event like this?

If so, that feels like a step backwards in terms of developer experience.

I really don't want to blend payload with inngest properties, e.g. many of my events had 'name' payload property.

This will mean that any property that you introduce as Inngest-reserved will have to be a breaking change.

punkpeye:
The logger interface needs to be exported:

```diff
- interface Logger {
+ export interface Logger {
  info(...args: LogArg[]): void;
  warn(...args: LogArg[]): void;
  error(...args: LogArg[]): void;
  debug(...args: LogArg[]): void;
}
```

or else users will get errors like:

> Exported variable 'updatePopularMcpServerSearchesJob' has or is using name 'Logger' from external module "/../node_modules/.pnpm/inngest@3.32.10-pr-920.20_express@5.1.0_fastify@5.5.0_typescript@5.9.2_zod@4.1.5/node_modules/inngest/dist/InngestCommHandler-C5pG18ux" but cannot be named.ts(4023)

p.s. The use of `tsdown` (or any bundler) for your node.js packages also makes it almost impossible to patch the issues. There is really no need to bundle node.js code.

jpwilliams:
> @punkpeye: Am I doing something wrong or did the data get flattened into the event like this?

Hmm no change has been made like this. What's the bug you're seeing here? Are types complaining? Or not working at runtime?

---

> @punkpeye: The logger interface needs to be exported

Oop! Thanks for the note - changing! 🙂 

---

> @punkpeye: The use of tsdown (or any bundler) for your node.js packages also makes it almost impossible to patch the issues. There is really no need to bundle node.js code.

It mainly helps us with a couple of things:

- Handling both ESM and CJS builds
- Easy targeting of specific Node.js versions, which helps with e.g. polyfills for older Node versions

That said, I guess what you're talking about here is the fact it seems to be creating file chunks like `foo-dPKxJ1yL.js`, which means patches can't be reliably applied?

If so, that's entirely unintentional - the bundling part of `tsdown` isn't what we care about here and can be disabled!

punkpeye:
Example:

```ts
await step.sendEvent('sync-github-repository-event', {
  data: {
    fullName: mcpServer.repository.githubRepository.fullName,
  },
  // Using the pushedAt timestamp to ensure that we don't sync the same repository multiple times within 24 hours unless the repository has been updated.
  id: `glama/github-repository.sync-${mcpServer.repository.githubRepository.id}-${githubRepositoryInfo.pushedAt}`,
  name: 'glama/github-repository.sync',
});
```

This is giving an error:

> Property 'fullName' is missing in type '{ data: { fullName: string; }; id: string; name: "glama/github-repository.sync"; }' but required in type '{ ts?: number | undefined; id?: string | undefined; name: "glama/github-repository.sync"; data?: any; user?: any; fullName: string; }'.ts(2345)

and schemas is defined as:

```ts
const schemas = new EventSchemas().fromSchema({
  'glama/chat-session.created': z.object({
    chatSessionReferenceId: z.uuid(),
  }),
  'glama/clear-all-mcp-server-caches': z.object({}),
  'glama/github-repository.sync': z.object({
    fullName: z.string(),
  }),
  // ...
```

jpwilliams:
@punkpeye Got it - thanks for the repro!

This should be fixed along with some new tests - could you check the latest version of `inngest@pr-920`? :)

punkpeye:
Confirming that this fixed type issues.

PR #927, opened 2025-03-31
Fix connecting to prod realtime

jpwilliams:
## Summary

Fix bug where WebSocket connections do not allow being redirected, as is the case when hitting the prod API with `ws://`, causing a `301`.

Vercel's edge runtime appropriately complies with the WebSocket spec here, whereas their local `dev` doesn't, hence the discrepancy.

## Checklist

## Related

- Complements inngest/inngest#2292

### DIFF
--- a/packages/realtime/src/subscribe/TokenSubscription.ts
+++ b/packages/realtime/src/subscribe/TokenSubscription.ts
@@ -367,8 +367,20 @@ export class TokenSubscription {
       };
 
       this.#ws.onerror = (event) => {
-        console.error("WebSocket error observed:", event);
-        ret.reject(event);
+        const serializedErr = {
+          event,
+          message: (event as ErrorEvent).message,
+          type: (event as ErrorEvent).type,
+          target: (event as ErrorEvent).target,
+          error: (event as ErrorEvent).error,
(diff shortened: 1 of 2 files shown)


packages/realtime/src/subscribe/TokenSubscription.ts
```diff
@@ -71,8 +71,12 @@ export class TokenSubscription {
 
     if (this.#app.apiBaseUrl) {
       url = new URL(path, this.#app.apiBaseUrl);
+
+      // If INNGEST_API_BASE_URL=url is set instead of INNGEST_DEV=url, assume
+      // that we should be connecting to a dev server
+      url.protocol = "ws:";
```
djfarrelly, line 77: Instead of inferring whether TLS is enabled for the WebSocket URL, should we use `url.protocol`'s current value and use `ws` or `wss` depending on whether it's http or https?

djfarrelly (approved):
Approved with suggestion

packages/realtime/src/subscribe/TokenSubscription.ts
jpwilliams: Brain fart! 💯

PR #947, opened 2025-04-07
Loosen `zod` dependency range

### LINKED ISSUE #933: [BUG] Update zod dependency version in inngest lib
**Describe the bug**
Zod is listed as a dependency using `~3.22.3` in the inngest client library (it's at version 3.24.2 as of this bug report).

In a monorepo with other packages that list zod as a dependency or peer dependency but are less restrictive on the versioning, zod doesn't get hoisted and leads to version mismatch issues.

**To Reproduce**
1. Set up a monorepo with npm workspaces
2. `npm install inngest ai -w packages/my-package`
3. Create a tool, like:
```
import { tool } from "ai";
import { z } from "zod";

const weatherTool = tool({
  description: 'Get the weather in a location',
  parameters: z.object({
    location: z.string().describe('The location to get the weather for'),
  }),
  execute: async ({ location }) => ({
    location,
    temperature: 72 + Math.floor(Math.random() * 21) - 10,
  }),
});
```

**Expected behavior**
Typescript doesn't complain.

**Actual behavior**
Typescript emits the following error on `parameters`:
```
No overload matches this call.
  Overload 1 of 2, '(tool: Tool<ToolParameters, unknown> & { execute: (args: any, options: ToolExecutionOptions) => PromiseLike<unknown>; }): Tool<...> & { ...; }', gave the following error.
    Type 'ZodObject<{ location: ZodString; }, "strip", ZodTypeAny, { location: string; }, { location: string; }>' is not assignable to type 'ToolParameters'.
      Type 'ZodObject<{ location: ZodString; }, "strip", ZodTypeAny, { location: string; }, { location: string; }>' is missing the following properties from type 'ZodType<any, any, any>': "~standard", "~validate"ts(2769)
```

**Additional context**
I think either upgrading zod or switching `~` -> `^` the package.json would fix this

jpwilliams:
## Summary

See https://github.com/inngest/inngest-js/issues/933#issuecomment-2778464713. If a user has issues with this unlocked, they just need to upgrade for now.

It's imperfect, but unblocks some users as our pinned `zod` version is lagging behind.

#920 alleviates this issue.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A KTLO
- [ ] ~Added unit/integration tests~ N/A KTLO

## Related

- Fixes #933
- See https://github.com/inngest/inngest-js/issues/933#issuecomment-2778464713
- #920 solves the issue

### DIFF
--- a/.changeset/silver-drinks-sing.md
+++ b/.changeset/silver-drinks-sing.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Loosen `zod` dependency range to allow for minor versions `>=3.22`
--- a/packages/inngest/package.json
+++ b/packages/inngest/package.json
@@ -214,7 +214,7 @@
     "strip-ansi": "^5.2.0",
     "temporal-polyfill": "^0.2.5",
     "ulidx": "^2.4.1",
-    "zod": "~3.22.3"
+    "zod": "^3.22.3"
   },
   "devDependencies": {
     "@actions/core": "^1.10.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -128,8 +128,8 @@ importers:
         specifier: ^2.4.1
         version: 2.4.1
       zod:
-        specifier: ~3.22.3
-        version: 3.22.3
+        specifier: ^3.22.3
+        version: 3.24.2
     devDependencies:
       '@actions/core':
         specifier: ^1.10.0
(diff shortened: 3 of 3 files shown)


jpwilliams:
This continues to fail type checks (and so breaks the user's build) if another dependency within a user's project pulls the version higher than we can support.

PR #950, opened 2025-04-09
Add connect leases

BrunoScheufler:
## Summary

This PR implements connect lease extensions.

## Checklist

## Related

- INN-

### DIFF
--- a/.changeset/rude-shrimps-arrive.md
+++ b/.changeset/rude-shrimps-arrive.md
@@ -1,0 +1,5 @@
+---
+"inngest": minor
+---
+
+Connect: Reliability improvements
--- a/.changeset/rude-shrimps-arrive.md
+++ b/.changeset/rude-shrimps-arrive.md
@@ -1,5 +1,5 @@
 ---
-"inngest": minor
+"inngest": patch
 ---
 
 Connect: Reliability improvements
--- a/packages/inngest/package.json
+++ b/packages/inngest/package.json
@@ -213,7 +213,6 @@
     "serialize-error-cjs": "^0.1.3",
     "strip-ansi": "^5.2.0",
     "temporal-polyfill": "^0.2.5",
-    "ulidx": "^2.4.1",
     "zod": "~3.22.3"
   },
   "devDependencies": {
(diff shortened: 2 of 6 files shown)


.changeset/rude-shrimps-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"inngest": minor
```
jpwilliams, line 2: Is this `patch` instead of `minor` if it's just internal reliability improvements?
```suggestion
"inngest": patch
```

packages/inngest/package.json
```diff
@@ -213,7 +213,6 @@
     "serialize-error-cjs": "^0.1.3",
     "strip-ansi": "^5.2.0",
     "temporal-polyfill": "^0.2.5",
-    "ulidx": "^2.4.1",
```
jpwilliams, line 216: Nice. Serendipitous that this happened near #944 so we can remove this!

packages/inngest/src/components/connect/index.ts
```diff
@@ -1069,49 +1176,60 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
+                `Gateway heartbeat missed (${connectionId})`,
+                attempt
+              )
+            );
+            return;
+          }
+
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
+          this.messageBuffer.flush(hashedSigningKey);
+        }, heartbeatIntervalMs / 2);
```
jpwilliams, line 1230: A nit not about this change, but should this be something like `heartbeatIntervalMs * .9`? atm this gives the server only half the time to respond.

To be a bit more succinct, maybe this code could send a heartbeat on the interval and then check if the _previous_ heartbeat was responded to, rather than waiting for the current one with a `setTimeout`.

jpwilliams (approved):

packages/inngest/src/components/connect/index.ts
BrunoScheufler: Fixed this, should be cleaner now

PR #982, opened 2025-05-16
feat: infer ID

simon-the-shark:
## Summary
Small typescript changes.

id of the function is no longer a `string`, but a generic type `<ID extends string>` which allows us for very specific literal type inferring and superior typesafety. I've tried to do this without changing your code, but it's pretty much impossible. I hope the changes are not controversial, I've tried to be minimal, just to get the results I want - but I didn't dive deeply into your types, so hopefully didn't break any convention. 

thanks to this you can do smth like this:

### Example
```ts
import { inngest } from "@backend/lib/inngest"

const functionTest1 = inngest.createFunction(
  {
    id: "function-test-1",
  },
  {
    event: "test/event-1",
  },
  async ({ logger }) => {
    logger.info("Hello, world!")
  }
)

const functionTest2 = inngest.createFunction(
  {
    id: "function-test-2",
  },
  {
    event: "test/event-2",
  },
  async ({ logger }) => {
    logger.info("Hello, world!")
  }
)

export const functions = [functionTest1, functionTest2]

export type FunctionsNames = (typeof functions)[number]["rawId"]
```

previously `FunctionsNames` was just a `string` and after this change it's a `type FunctionsNames = "function-test-1" | "function-test-2"` literal

Zrzut ekranu 2025-05-16 o 23 38 53

and that's only a fraction of what you can do with types after this change!

## Checklist

## Related

- INN-

### DIFF
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -736,7 +736,10 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
   /**
    * Runtime-only validation.
    */
-  private sanitizeOptions<T extends InngestFunction.Options>(options: T): T {
+  private sanitizeOptions<
+    ID extends string,
+    T extends InngestFunction.Options<ID>,
+  >(options: T): T {
     if (Object.prototype.hasOwnProperty.call(options, "fns")) {
       // v2 -> v3 migration warning
       console.warn(
@@ -934,9 +937,11 @@ export namespace Inngest {
         >
       >
     >,
+    ID extends string = string,
   >(
     options: Omit<
       InngestFunction.Options<
+        ID,
         TClient,
         TMiddleware,
         AsArray<TTrigger>,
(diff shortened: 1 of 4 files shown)


jpwilliams:
Thanks, @simon-the-shark! 🙌 

What are you using this for internally? I'm curious why you're looking to grab the literal ID of a function.

simon-the-shark:
so I want to have a middleware that keeps track of different functions running and do some actions based on **some** of them. and thanks to this we can have better typesafety there. 

if someone changes id of the function / adds new one / remove one then we can get a type error there indicating you need to adjust the middleware too. 

Also later on it should also be possible to map these ids to events and paylods with typescript. And currently by default we don't know pretty much anything in the middleware when it comes to types, the payload is `any` or smth like that and the id/names are just strings. And since the functions are pretty much known at compile time (you can correct me if I'm wrong) - we could have had full literal typesafety. 

In ideal world, you would just do
```ts
// here the id is literraly typed to e.g. `"test_id_1" | "test_id_2"` (all possible functions for the client or smth)
switch(fn.id){
  case "test_id_1": {
      fn.event.data // here the data is exactly typed to the possible payloads that are inferred from the zod schemas defined, but only those related to "test_id_1" function
   }
}
```
I believe this should be possible to make it work in your sdk, but it's prolly not that easy.

I've pretty much managed to make it work at first with some wrappers and utils and the typesafety was super great and fully working but it caused circular dependency, so no running the app for me in such case :((. 

I hope I'll manage to make this work after this small change without circular dependencies, but didn't do that fully yet. but the small example I've shown you above should be working.

simon-the-shark:
@jpwilliams do i still need to do smth more here or you'll just process this PR in your own time?

PR #991, opened 2025-05-28
Add support for singletons

KiKoS0:
## Summary
Adding support for configuring singletons. Other configs don't seem to have unit/integration tests so I skipped them but lmk if I missed them.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Followup when releasing singletons.
- [ ] Added unit/integration tests N/A
- [ ] Added changesets if applicable N/A

## Related
https://linear.app/inngest/issue/INN-3904

Needs: https://github.com/inngest/inngest/pull/2459

### DIFF
--- a/packages/inngest/src/components/InngestFunction.ts
+++ b/packages/inngest/src/components/InngestFunction.ts
@@ -155,6 +155,7 @@ export class InngestFunction<
       debounce,
       timeouts,
       priority,
+      singleton,
     } = this.opts;
 
     /**
@@ -197,6 +198,7 @@ export class InngestFunction<
       debounce,
       priority,
       timeouts,
+      singleton,
     };
 
     if (cancelOn) {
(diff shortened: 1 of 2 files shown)


darwin67 (approved):

packages/inngest/src/components/InngestFunction.ts
```diff
@@ -565,6 +567,29 @@ export namespace InngestFunction {
...
+       * e.g. "event.data.user_id".
+       */
+      key?: string;
+
+      /**
+       * Determines how to handle new runs when one is already active for the same key.
+       * - `"ignore"` skips the new run.
+       * - `"cancel"` cancels the existing run and starts the new one.
+       */
+      mode: "ignore" | "cancel";
```
amh4r, line 590: Just a thought: would `skip` be a clearer option than `ignore`? Since we have the "skipped" status elsewhere for skipped runs

KiKoS0: hmm good point i think it's more consistent

PR #997, opened 2025-06-04
Export types for convenience and discoverability

djfarrelly:
## Summary
With all types exported on the `"inngest/connect"` path along side the main `connect()` function, users can more easily click into this file and discover related types they may want to use. This is helpful compared to previously having to dig into the code or view the docs to know where these types are located. This also makes import statements simpler for end users as they can have all related imports on a single line.

## Checklist

## Related

- INN-5009


jpwilliams (approved):
Thanks!

PR #999, opened 2025-06-05
Allow arbitrary field/method access via ProxyLogger

amh4r:
## Summary
Allow access to arbitrary fields/methods via the `ProxyLogger`. For example, if I have a custom logger with the method `foo` then it'll be callable via the `ProxyLogger`.

## Checklist


jpwilliams (approved):
🙌

PR #793, opened 2025-01-06
Ensure `node:async_hooks` import is conditional

jpwilliams:
## Summary

This could help Webpack avoid bundling issues when scanning for imports.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Bug fix
- [ ] Added unit/integration tests
- [ ] Added changesets if applicable

## Related

- Exploration for #791

PR #1020, opened 2025-06-25
obscure "node:async_hooks" import from bundlers

thomasballinger:
## Summary
Obscure the dynamic import of an optional Node.js dependency to prevent bundlers from attempting to inline the code. This ensures this import will happen at runtime even when this code is bundled.

This makes the bundled code work in environments without Node.js APIs like the Convex JS runtime without erroring on undefined imports that would otherwise require a stub/polyfill. Maybe this helps in Cloudflare too when Node.js compatibilty is not turned on.

---

The code that previously compiled to
```js
        try {
            const { AsyncLocalStorage } = await Promise.resolve().then(() => __importStar(require("node:async_hooks")));
            resolve(new AsyncLocalStorage());
        }
```
(see https://app.unpkg.com/inngest@3.39.2/files/components/execution/als.js#L60-61)

now compiles to
```js
        try {
            // Obscure this import to keep bundlers from bundling it.
            const dynamicImport = (path) => {
                const safePath = path.split("/").join("/");
                return Promise.resolve(`${safePath}`).then(s => __importStar(require(s)));
            };
            const { AsyncLocalStorage } = 
            // eslint-disable-next-line @typescript-eslint/consistent-type-imports
            await dynamicImport("node:async_hooks");
            resolve(new AsyncLocalStorage());
        }
```
This is a pattern that esbuild (tested) and hopefully any reasonable bundler will leave as a runtime import/require.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A just for library developers
- [ ] ~~Added unit/integration tests~~ N/A this would require some plumbing to test, lmk if that's desirable

## Related
Discord thread with @djfarrelly on the Convex server https://discord.com/channels/1019350475847499849/1387074958487720047/1387074958487720047

### DIFF
--- a/.changeset/fair-seals-rush.md
+++ b/.changeset/fair-seals-rush.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+The optional import of Node.js module "node:async_hooks" is more likely to by dynamic, even after bundling this library. This change makes this feature-detecting dynamic import work correctly when bundled for the Convex JS runtime.
--- a/packages/inngest/src/components/execution/als.ts
+++ b/packages/inngest/src/components/execution/als.ts
@@ -53,7 +53,16 @@ export const getAsyncLocalStorage = async (): Promise<AsyncLocalStorageIsh> => {
       // eslint-disable-next-line @typescript-eslint/no-misused-promises, no-async-promise-executor
       async (resolve) => {
         try {
-          const { AsyncLocalStorage } = await import("node:async_hooks");
+          // Obscure this import to keep bundlers from bundling it.
+          const dynamicImport = <T>(path: string): Promise<T> => {
+            const safePath = path.split("/").join("/");
+            return import(safePath) as Promise<T>;
+          };
+          const { AsyncLocalStorage } =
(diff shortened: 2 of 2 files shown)


jpwilliams (approved):
Thanks for this, @thomasballinger! 🙌 

Always a little nervous altering this code, as the many bundlers can be so finicky about dynamic imports, especially non-literal ones.

CI's looking good though. We'll have to set up a Convex project here too.

Linking some related PRs the tread a similar path:

- #931 
- #793

thomasballinger:
Thanks! Ha yeah those PRs tell a story, glad to have contributed to the next chapter :)
And there's a working Convex example now too, that's great https://github.com/inngest/inngest-convex-test

jpwilliams:
@thomasballinger FYI this was reverted 🙁 
- Released in `3.40.0`
- Reverted in #1028 (released in `3.40.1`)

Unfortunately it borked Next.js's complex Webpack resolution. For now, stick on `3.40.0` for Convex and we'll assess how best to reintroduce this. 🤔

PR #1039, opened 2025-07-30
Skip step targeting when the already fulfilled

amh4r:
## Summary

When a step is targeted but also fulfilled, don't run the step. This happens when the Executor mistakenly sends an extra parallel step target (the `stepId` query param) even though the step is already cached in the request body's `steps` object

## Checklist


jpwilliams:
Closing this is I'm viewing it as an executor bug.

PR #1055, opened 2025-08-22
Add support for Standard Schemas

jpwilliams:
## Summary

Instead of attempting to directly support both Zod v3 and v4 together, this PR introduces an `EventSchemas#fromSchema()` method, which allows any schema to be specified so long as it's compliant as a [Standard Schema](https://standardschema.dev/).

This means we deprecate `.fromZod()` and encourage use of `.fromSchema()` instead, as this supports `zod@>=3.40.0` and a multitude of other validation libraries.

## Checklist

- [ ] ~Added a [docs PR](https://github.com/inngest/website) that references this PR~ N/A Documenting in #920

## Related

- Targeted for release with #920
- Addresses #933


punkpeye:
This is great. Is this planned to be in v3?

jpwilliams:
> @punkpeye: This is great. Is this planned to be in v3?

Yes! It'll be shipped with #920.

PR #1080, opened 2025-09-23
Support Vite's import.meta.env

djfarrelly:
## Summary
When testing a Vite app for Python SDK realtime support, Vite does not pass `VITE_` prefix env vars to the client via `process.env`. 

**NOTE** - This might break other bundlers who really hate accessing `import.meta` - DO NOT MERGE FOR NOW

## Checklist

## Related
- https://github.com/inngest/inngest-py/pull/281
- [EXE-245](https://linear.app/inngest/issue/EXE-245)


packages/realtime/src/env.ts
```diff
@@ -27,7 +27,27 @@ export type ExpectedEnv = {
  */
 const env: ExpectedEnv | undefined = (() => {
   try {
-    // Nodeish, including Vite
+    // Vite
+    try {
+      // @ts-expect-error - import.meta.env is not available in CommonJS output
+      if (typeof import.meta !== 'undefined' && import.meta.env) {
+        // @ts-expect-error - import.meta.env is not available in CommonJS output
+        const env = import.meta.env;
```
djfarrelly: ts lint does not like using `import.meta` at all, so I've silenced it for now here, but maybe there is a better way. We may need to test other environments. I only tested vanilla Vite.

jpwilliams: This is fine.

'tis an ugly thing to have to do, I believe we can't do things like `declare const import: ...` like we do with Deno/Netlify.

As long as it's in that no-op try/catch, we're fine here! 🙌

PR #1091, opened 2025-09-24
Add streaming support to inngest/node

amh4r:
## Summary
Add streaming support to `inngest/node`

### DIFF
--- a/.changeset/itchy-moons-yawn.md
+++ b/.changeset/itchy-moons-yawn.md
@@ -1,0 +1,5 @@
+---
+"inngest": minor
+---
+
+Add streaming support to inngest/node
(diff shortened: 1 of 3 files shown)


jpwilliams (changes requested):
Awesome! 🙌

Could you add this to the integration tests?

Should just need an entry here: https://github.com/inngest/inngest-js/blob/726a289f8a8f78ed5631289a20fe345c6751cf54/.github/workflows/pr.yml#L247-L254

Like

```
- example: node
  streaming: force
```

adam-subframe (approved):
Thank you so much!

jpwilliams (approved):

PR #1129, opened 2025-10-09
Export types for the step result

de-tester:
## Summary

#1111 removed direct access to the files/helpers. `Jsonify` is useful when one need to pass step output to some helper methods. It's quite hard to type that helper method params without wrapping existing type with the conversion that inngest does to serialize step result. 

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ 
- [ ] ~~Added unit/integration tests~~

### DIFF
--- a/.changeset/large-balloons-study.md
+++ b/.changeset/large-balloons-study.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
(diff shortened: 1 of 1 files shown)


jpwilliams:
Thanks, @de-tester!

PR #1130, opened 2025-10-10
Add Exports for Connect Types

billdybas:
## Summary
When using TypeScript & ESM, the [example code for Connect](https://www.inngest.com/docs/setup/connect#getting-started) isn't currently valid and errors with:

`"Cannot find module 'inngest/components/connect/types' or its corresponding type declarations."`.

Adds the necessary `exports` to `package.json` so the example code compiles.

### DIFF
--- a/packages/inngest/package.json
+++ b/packages/inngest/package.json
@@ -225,6 +225,14 @@
       },
       "import": "./internals.js",
       "require": "./internals.cjs"
+    },
+    "./components/connect/types": {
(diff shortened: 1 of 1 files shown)


jpwilliams:
Thanks, @billdybas!

Closing this in favour of #1133; we should be importing this via `"inngest/connect"` instead of directly from the file.

Also documenting this in inngest/website#1295.

PR #1134, opened 2025-10-14
Support for latest OpenAI responses API

tedjames:
## Summary
Created model and adapter for the latest OpenAI responses API.

We currently only support the OpenAI chat API which is now considered to be deprecated. This PR introduces a new model called `openaiResponses` which can be implemented like this:

```typescript
import { openaiResponses } from "@inngest/ai";
import { inngest } from "./client";

export const runTest = inngest.createFunction(
  {
    id: "run-openai-test",
    name: "OpenAI Responses Test",
  },
  { event: "agent/chat.requested" },
  async ({ step }) => {
    try {
      const result = await step.ai.infer("test-openai", {
        model: openaiResponses({
          model: "gpt-5",
          defaultParameters: { reasoning: { effort: "medium" } },
        }),
        body: {
          input: [
            {
              role: "user",
              content:
                "Hello! Can you explain to me how redux state management works for React?",
            },
          ],
        },
      });

      return {
        success: true,
        result: result.output_text,
      };
    } catch (error) {
      console.error(error);
      throw error;
    }
  }
);

```

I've also created a few unit and smoke tests. Make sure to add a .env file to the AI package with an `OPENAI_API_KEY` to run the smoke tests. To run the tests, add your env vars and run: `pnpm test` and `pnpm test:smoke`.

This will run tests for both gemini and openai (gpt-5) models.

Once this is merged in, we can work on adapters to integrate this into AgentKit as well.

## Checklist

## Notes/Questions
- Does it make more sense to name the previous openai model "openaiChat" and the new one just "openai" instead? This would potentially be a breaking change for users but is more concise / cleaner overall.

## Related

[OpenAI Responses API Docs](https://platform.openai.com/docs/api-reference/responses)

[Issue #204 in AgentKit](https://github.com/inngest/agent-kit/issues/204)

### DIFF
--- a/.changeset/gentle-tools-raise.md
+++ b/.changeset/gentle-tools-raise.md
@@ -1,5 +1,5 @@
 ---
-"@inngest/ai": minor
+"@inngest/ai": patch
 ---
 
 added support for latest openai responses api


packages/ai/src/models/openai.ts
```diff
@@ -6,6 +6,7 @@ import { envKeys, processEnv } from "../env";
  * Create an OpenAI model using the OpenAI chat format.
  *
  * By default it targets the `https://api.openai.com/v1/` base URL.
+ * @deprecated Use `openaiResponses` instead.
```
jpwilliams, line 9: Oh damn, this is deprecated now?

Is the format very different from the `openai-chat` one?

packages/ai/src/adapter.ts
```diff
@@ -127,6 +128,7 @@ export namespace AiAdapter {
  */
 const adapters = {
   "openai-chat": null as unknown as OpenAiAiAdapter,
+  "openai-responses": null as unknown as AiAdapter.Any,
```
jpwilliams: Can this be `as OpenAiResponsesAdapter`?

I think this is used downstream in a few places for typing, so needs to have the I/O types on here.

packages/ai/src/adapter.ts
```diff
@@ -137,3 +139,6 @@ const adapters = {
  * All AI adapters available for use.
  */
 export type AiAdapters = typeof adapters;
+
+// Mark as used at runtime to satisfy no-unused-vars while keeping type inference
+void adapters;
```
jpwilliams, line 145: 🧠

.changeset/gentle-tools-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inngest/ai": minor
```
jpwilliams:
```suggestion
"@inngest/ai": patch
```

Super nit - we're in v0 so bumping the minor represents a breaking change; `patch` represents minor/patch changes atm.

tedjames: Makes sense! Just wondering though - should we name the previous openai model "openaiChat" and the new one just "openai" instead? This would be a breaking change for users but is more concise / cleaner overall.

packages/ai/src/models/openai.ts
tedjames: Chat completions are still supported but will not be receiving the latest capabilities like native tool calls, stateful context and reasoning strength.

Format is a bit different as well. You can now pass in instructions and input as opposed to just messages.

**Before:**

```
curl https://api.openai.com/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
      "model": "gpt-5",
      "messages": [
          {"role": "system", "content": "You are a helpful assistant."},
          {"role": "user", "content": "Hello!"}
      ]
  }'
```

**After:**
```
curl https://api.openai.com/v1/responses \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
      "model": "gpt-5",
      "instructions": "You are a helpful assistant.",
      "input": "Hello!"
  }'
```

More on all of this here: [https://platform.openai.com/docs/guides/migrate-to-responses](https://platform.openai.com/docs/guides/migrate-to-responses)

Issue #1066, opened 2025-09-10
[BUG] @inngest/test step mock handler getting called too many times


jb-chief:
**Describe the bug**
When mocking a step using @inngest/test, the handler gets called many times and is called with no args. This makes it difficult (impossible?) to do any actual work in the handler that you need in order to simulate an actual run.

**To Reproduce**

1. Create a function like:
```
export const testAThing = inngest.createFunction(
  { id: "test/test-a-thing" },
  { event: "test/test-a-thing.run" },
  async ({ step }) => {
    const result1 = await step.run("do-thing-1", async () => {
      console.log("DO THING 1");
      return Promise.resolve("result1");
    });
    const result2 = await step.run("do-thing-2", async () => {
      console.log("DO THING 2");
      return Promise.resolve("result2");
    });
    const result3 = await step.run("do-thing-3", async () => {
      console.log("DO THING 1");
      return Promise.resolve("result3");
    });
    return { result1, result2, result3 };
  },
);
```
2. Write a test like:
```
it("test a thing", async () => {
  const t = new InngestTestEngine({ function: testAThing });
  const { result } = await t.execute({
    events: [
      {
        name: "test/test-a-thing.run",
      },
    ],
    steps: [
      {
        id: "do-thing-1",
        handler() {
          console.log("MOCK THING 1");
          return "my-custom-result-1";
        },
      },
      {
        id: "do-thing-3",
        handler() {
          console.log("MOCK THING 3");
          return "my-custom-result-3";
        },
      },
    ],
  });
  expect(result).toBeDefined();
});
```

**Expected behavior**
I would expect to see:
```
MOCK THING 1
DO THING 2
MOCK THING 3
```

Instead, I see:
```
MOCK THING 1
MOCK THING 3
DO THING 2
MOCK THING 1
MOCK THING 3
```

In this minimal example case, it's not a big deal. But in some cases, I'd like for those handlers to actually do some work to simulate what the step would actually do (e.g. update the DB in some way, return different data based on args, etc.). The fact that the "do-thing-3" mock handler gets called before step 2 is actually run creates very strange behavior that I'm unsure how to resolve.

PR #1140, opened 2025-10-22
@inngest/test: call mock step handlers only once and at the right time

jb-chief:
- **@inngest/test: call mock step handlers only once and at the right time**

## Summary
When mocking steps using the `InngestTestEngine`, only call each handler once, at the time that the step would run during function execution. This allows consumers of the `InngestTestEngine` to write more complex / correct tests that perform some logic partway through function execution.

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A docs already made it seem like this behavior was the case
- [x] ~~Added changesets if applicable~~ N/A am I supposed to do this or is this part of the PR process?

## Related
https://github.com/inngest/inngest-js/issues/1066

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -202,6 +202,22 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
+  "test_test":
+    name: "test: Test"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jb-chief:
🙌 woo thank you!

Issue #1138, opened 2025-10-20
[BUG] various `@opentelemetry/*` dependencies are long outdated


BastiDood:
### Describe the bug
The main `inngest` package currently depends on `@opentelemetry/*` packages that have long been outdated. This outdated constraint by `inngest` causes peer dependency warnings in various packages that keep up with the latest updates.

### To Reproduce
For my use case, I use `@vercel/otel` at work, which keeps its dependencies up to date. So, simply running these commands will trigger the peer dependency incompatibility.

```shell
pnpm install inngest @vercel/otel
```

### Expected behavior
Zero peer dependency incompatibilities for `@vercel/otel` (due to outdated direct dependencies of `inngest`).

## Code snippets / Logs / Screenshots
We should upgrade these dependencies. For the `0.x.y` packages, we could consider relaxing the constraint so that it's compatible with `0.200.x` up to `0.300.y` [(which is what `@vercel/otel` does)](
https://github.com/vercel/otel/blob/14f773c2666fcfe418e9a1a0d4205d08c8f23fd1/packages/otel/package.json#L91-L99).

https://github.com/inngest/inngest-js/blob/b3e7305fd5be0148ca76c47476ab3e8f7cfe1c80/packages/inngest/package.json#L278-L284

btree1970:
Plus one to this, I am having issues setting up telemetry for my AI agents with langfuse. Please fix this

Linell:
@btree1970 @BastiDood I think that the latest release (`3.46.0`) should help sort this out for you!

gianpaj:
and now there's a high level vulnerability 

https://github.com/advisories/GHSA-q7rr-3cgh-j5r3

PR #1156, opened 2025-11-05
refactor(otel): bump OTEL deps && tweaks for compatibility

Linell:
## Summary
Bumps the OTEL packages and tweaks our usages to be compatible with those changes. Mentioned in #1138 

## Checklist

## Related

- INN-5471

### DIFF
--- a/.changeset/late-toes-dance.md
+++ b/.changeset/late-toes-dance.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Updated OpenTelemetry dependencies to support broader version ranges and adapted code for OTEL 2.x API compatibilty 


packages/inngest/src/components/execution/otel/util.ts
```diff
@@ -31,10 +76,6 @@ export const createProvider = (
     instrumentations: instrList,
   });
 
-  p.register({
-    contextManager: new AsyncHooksContextManager().enable(),
-  });
```
jpwilliams, line 36: Do we no longer need to do `p.register()` at all?

packages/inngest/package.json
```diff
@@ -275,13 +275,13 @@
...
     "@inngest/ai": "^0.1.3",
     "@jpwilliams/waitgroup": "^2.1.1",
-    "@opentelemetry/api": "^1.9.0",
-    "@opentelemetry/auto-instrumentations-node": "^0.66.0",
-    "@opentelemetry/context-async-hooks": "^1.30.1",
-    "@opentelemetry/exporter-trace-otlp-http": "^0.57.2",
-    "@opentelemetry/instrumentation": "^0.57.2",
-    "@opentelemetry/resources": "^1.30.1",
-    "@opentelemetry/sdk-trace-base": "^1.30.1",
+    "@opentelemetry/api": ">=1.9.0 <2.0.0",
```
jpwilliams: What's the difference between `>=1.9.0 <2.0.0` and `^1.9.0`? Aren't they the same?

Is it okay that this is `<2.0.0` as well, seeing as we mention that we're using `OpenTelemetry v2` in a few spots?

packages/inngest/src/components/execution/otel/util.ts
```diff
@@ -11,17 +11,62 @@ import { InngestSpanProcessor } from "./processor.ts";
...
+
   const processor = new InngestSpanProcessor();
 
   const p = new BasicTracerProvider({
     spanProcessors: [processor],
   });
 
+  const contextManager = new AsyncHooksContextManager().enable();
+  context.setGlobalContextManager(contextManager);
+  trace.setGlobalTracerProvider(p);
```
jpwilliams: I guess this is a requirement now to power `trace.getTracerProvider()` instead of being implicit?

packages/inngest/src/components/execution/otel/util.ts
Linell: Correct, we've just got to do the trace and context setting manually.

packages/inngest/package.json
Linell: The v2 that was mentioning was some SDK bits, not the api proper. You're right about them being the same though I think? I just mimicked [what Vercel did](https://github.com/vercel/otel/blob/14f773c2666fcfe418e9a1a0d4205d08c8f23fd1/packages/otel/package.json#L92).

Issue #1142, opened 2025-10-24
Realtime middleware types are broken


gdaybrice:
**Describe the bug**
When adding the realtimeMiddleware, type inference is broken. The publish method works as expected though, so purely a type issue.

```
The inferred type of 'inngest' cannot be named without a reference to '.pnpm/inngest@3.44.3_@vercel+node@5.3.24_rollup@4.50.2__express@5.1.0_fastify@5.6.1_hono@4.9._08dfbc80aa6d893fa674cc3b8b0dd522/node_modules/inngest'. This is likely not portable. A type annotation is necessary.
```

```
Property 'publish' does not exist on type 'Context<Inngest<{ id: string; middleware: [InngestMiddleware<{ name: string; init({ client }: { client: Any; fn?: Any | undefined; }): { onFunctionRun({ ctx: { runId } }: Readonly<{ readonly steps: Readonly<{ id: string; error?: any; data?: any; input?: any; }>[]; readonly fn: Any; readonly reqArgs: readonly unknown...'.
```

**To Reproduce**
```
export const inngest = new Inngest({
  id: 'app',
  middleware: [realtimeMiddleware()]
});
```

**System info**
- "@inngest/realtime": "0.4.4"
- "inngest": "3.44.3"

jpwilliams:
Hi @gdaybrice! 👋 

Thanks for the report!

Could I grab your `tsconfig.json` and TypeScript version too?

gdaybrice:
@jpwilliams here you go, let me know if you need anything else,

For context, using a turbo monorepo with bun, `"typescript": "5.9.3"`

*tsconfig.json* in nextjs app
```
{
  "extends": "@monorepo/typescript-config/nextjs.json",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@/*": ["./*"]
    },
    "plugins": [
      {
        "name": "next"
      }
    ],
    "declaration": false
  },
  "include": [
    "next-env.d.ts",
    "next.config.ts",
    "**/*.ts",
    "**/*.tsx",
    ".next/types/**/*.ts"
  ],
  "exclude": ["node_modules"]
}
```

*@monorepo/typescript-config/nextjs.json*

```
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Next.js",
  "extends": "./base.json",
  "compilerOptions": {
    "plugins": [{ "name": "next" }],
    "module": "ESNext",
    "moduleResolution": "Bundler",
[...]

PR #1160, opened 2025-11-07
Fix naughty Zod versions in peer deps

jpwilliams:
## Summary

Fixes Zod peer deps based on their [For library authors](https://zod.dev/library-authors) page.

A hopeful fix for #1142, though testing with canary builds might be frustrating due to dual `inngest` installs. 🤔 

## Checklist

- [ ] ~Added unit/integration tests~ N/A

## Related

- Potential fix for #1142


packages/inngest/src/components/InngestFunctionReference.test.ts
```diff
@@ -452,7 +452,7 @@ describe("referenceFunction", () => {
       // Output should be a discriminated union
       type ExpectedOutput =
         // biome-ignore lint/suspicious/noExplicitAny: z.any() returns any type
-        { type: "success"; data: any } | { type: "error"; message: string };
+        { type: "success"; data?: any } | { type: "error"; message: string };
```
jpwilliams, line 455: This is really just testing that the Zod type is consistent as it passes through a few generics; it's okay that we tweak this due to the version drop.

PR #1164, opened 2025-11-11
Update to use latest pre-built deps

darwin67:
## Summary

Was just trying to look at some stuff quickly and nix started a nodejs compilation.

Saw this was from May 2025 and was building an RC version of node 24, so bumped the flake version and have it use the latest prebuilt for faster loading.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


darwin67:
Never mind, seems like I just didn't pull down the latest `main`. Closing

PR #1168, opened 2025-11-14
feat: add Anthropic OTEL instrumentation

Linell:
## Summary
Adds the [@traceloop/instrumentation-anthropic](https://github.com/traceloop/openllmetry-js/tree/main/packages/instrumentation-anthropic) package so that we're able to automatically instrument Anthropic calls like we do for OpenAI. We're adding this one manually because it isn't automatically included in auto-instrumentation-node.

Note that the data it's returning _is_ slightly different from the OpenAI results, and that there's a display bug seemingly caused by one of the captured attributes being the entire prompt.

Screenshot 2025-11-13 at 11 22 31 PM

Screenshot 2025-11-13 at 11 23 48 PM

## Related

- INN-

### DIFF
--- a/packages/inngest/src/components/execution/otel/util.ts
+++ b/packages/inngest/src/components/execution/otel/util.ts
@@ -6,6 +6,7 @@ import {
   registerInstrumentations,
 } from "@opentelemetry/instrumentation";
 import { BasicTracerProvider } from "@opentelemetry/sdk-trace-base";
+import { AnthropicInstrumentation } from "@traceloop/instrumentation-anthropic";
 import { InngestSpanProcessor } from "./processor.ts";
 
 export type Behaviour = "createProvider" | "extendProvider" | "off" | "auto";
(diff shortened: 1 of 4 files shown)


packages/inngest/src/components/execution/otel/util.ts
```diff
@@ -27,6 +28,13 @@ export const createProvider = (
     ...getNodeAutoInstrumentations(),
   ];
 
+  try {
+    require.resolve("@anthropic-ai/sdk");
+    instrList.push(new AnthropicInstrumentation());
```
jpwilliams: Detecting installation (especially via `require.resolve()`) may not be viable in all runtimes; could we just always add this instrumentation?

jpwilliams (changes requested):
Hell yeah 🙌

Linell:
@rhino1998 already has a Linear ticket for the UI problem [here](https://linear.app/inngest/issue/EXE-931/extended-trace-details-value-column-overflows-with-long-values).

PR #1176, opened 2025-11-19
feat(metadata): add step.update.metadata

Linell:
Adds the ability for users to add metadata about their runs and steps

- [x] inside step context, target defined
- [x] inside step context, target undefined
- [x] outside step context, target defined
- [x] outside step context, target undefined

## Summary

You can see how this currently ends up being used in [the examples here](https://github.com/Linell/inngest-sandbox/pull/1/files). tl;dr:
```ts
export const metadataExampleOne = inngest.createFunction(
  { id: "metadata-inside-step-context-target-undefined" },
  { event: "metadata/inside/undefined-target" },
  async ({ event, step }) => {
    const result = await step.run("example-step", async () => {
      await inngest.metadata.update({ key: "metadata/inside/undefined-target" });

      return { success: true };
    });

    return result;
  },
);

```

## Checklist

## Related

[EXE-816: Implement step.metadata.update](https://linear.app/inngest/issue/EXE-816/implement-stepmetadataupdate)

### DIFF
--- a/packages/inngest/src/components/InngestStepTools.ts
+++ b/packages/inngest/src/components/InngestStepTools.ts
@@ -792,7 +792,7 @@ export type InternalStepTools = GetStepTools<Inngest.Any> & {
     idOrOptions: StepOptionsOrId,
     ...args: Parameters<typeof fetch>
   ) => Promise<{
-    status: number;
+    status_code: number;
     headers: Record<string, string>;
     body: string;
   }>;
--- a/.changeset/heavy-bags-stare.md
+++ b/.changeset/heavy-bags-stare.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix step.fetch response status code
--- a/.changeset/selfish-lions-camp.md
+++ b/.changeset/selfish-lions-camp.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix Connect heartbeater stops too soon during shutdown
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1766070988,
-        "narHash": "sha256-G/WVghka6c4bAzMhTwT2vjLccg/awmHkdKSd2JrycLc=",
+        "lastModified": 1767379071,
+        "narHash": "sha256-EgE0pxsrW9jp9YFMkHL9JMXxcqi/OoumPJYwf+Okucw=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "c6245e83d836d0433170a16eb185cefe0572f8b8",
+        "rev": "fb7944c166a3b630f177938e478f0378e64ce108",
         "type": "github"
       },
       "original": {
--- a/packages/inngest/src/components/Fetch.ts
+++ b/packages/inngest/src/components/Fetch.ts
@@ -81,7 +81,7 @@ const createFetchShim = (): StepFetch => {
 
     return new Response(jsonRes.body, {
       headers: jsonRes.headers,
-      status: jsonRes.status,
+      status: jsonRes.status_code,
     });
   };
 
--- a/packages/inngest/src/components/connect/index.ts
+++ b/packages/inngest/src/components/connect/index.ts
@@ -1231,13 +1231,7 @@ class WebSocketWorkerConnection implements WorkerConnection {
       }, heartbeatIntervalMs);
     }
 
-    conn.cleanup = () => {
-      this.debug("Cleaning up worker heartbeat", {
-        connectionId,
-      });
-
-      clearInterval(heartbeatInterval);
-
+    conn.cleanup = async () => {
       if (closed) {
         return;
       }
@@ -1258,6 +1252,13 @@ class WebSocketWorkerConnection implements WorkerConnection {
       this.debug("Closing connection", { connectionId });
       ws.onerror = () => {};
       ws.onclose = () => {};
+
+      // We must wait for all in-flight requests to complete before closing
+      // the connection and stopping the heartbeater. If we don't, then
+      // Inngest Server will mark the step as failed since it lost contact
+      // with the worker
+      await this.inProgressRequests.wg.wait();
+
       ws.close(
         1000,
         workerDisconnectReasonToJSON(WorkerDisconnectReason.WORKER_SHUTDOWN),
@@ -1266,6 +1267,12 @@ class WebSocketWorkerConnection implements WorkerConnection {
       if (this.currentConnection?.id === connectionId) {
         this.currentConnection = undefined;
       }
+
+      this.debug("Cleaning up worker heartbeat", {
+        connectionId,
+      });
+
+      clearInterval(heartbeatInterval);
     };
 
     return conn;


packages/inngest/src/types.ts
```diff
@@ -1542,6 +1542,38 @@ export interface StepOptions {
  */
 export type StepOptionsOrId = StepOptions["id"] | StepOptions;
 
+/**
+ * Either an object containing info to target a run/step/step attempt/span.
+ *
+ * @public
```
jpwilliams: Appreciate that you kept these tags! It was for an old automated system so we could track API breaks - no need to keep them around any more though.

packages/inngest/src/types.ts
```diff
@@ -1542,6 +1542,38 @@ export interface StepOptions {
  */
 export type StepOptionsOrId = StepOptions["id"] | StepOptions;
 
+/**
+ * Either an object containing info to target a run/step/step attempt/span.
```
jpwilliams: Should this be:
```suggestion
 * An object containing info to target a run/step/step attempt/span, used for attaching metadata.
```

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+ * ```ts
+ * import { metadataMiddleware } from "inngest/experimental";
+ *
+ * const inngest = new Inngest({
+ *   id: "my-app",
+ *   middleware: [metadataMiddleware()],
+ * });
+ * ```
+ */
+export const metadataMiddleware = () => {
```
jpwilliams, line 353: Is the thought of the middleware to ensure that it's opt-in? Any reason to use middleware over an flag in the client options?

Not being prescriptive about either, just curious of the thought process! :)

packages/inngest/src/api/api.ts
```diff
@@ -408,6 +409,80 @@ export class InngestApi {
...
+      headers?: Record<string, string>;
+    },
+  ): Promise<Result<void, ErrorResponse>> {
+    const url = await this.getTargetUrl(
+      `/v1/runs/${args.target.run_id}/metadata`,
+    );
+    const payload = { target: args.target, metadata: args.metadata };
+
+    console.log("Fetching URL: ", url);
+    console.log("With Payload: ", JSON.stringify(payload, null, 2));
```
jpwilliams: Dangling logs. We could use `client.logger.*()` if we do still need to add something.

packages/inngest/src/api/api.ts
```diff
@@ -408,6 +409,80 @@ export class InngestApi {
...
+        kind: string;
+        op: string;
+        values: Record<string, unknown>;
+      }>;
+    },
+    options?: {
+      headers?: Record<string, string>;
+    },
+  ): Promise<Result<void, ErrorResponse>> {
+    const url = await this.getTargetUrl(
```
jpwilliams: I added `this.req()` a little while ago that can help clean up some of this boilerplate

https://github.com/inngest/inngest-js/blob/94d80ab4ffbd4c72d35b0033e2afcee63b425f2b/packages/inngest/src/api/api.ts#L152-L178

packages/inngest/src/api/api.ts
```diff
@@ -408,6 +409,80 @@ export class InngestApi {
...
+          "Content-Type": "application/json",
+          ...options?.headers,
+        },
+      },
+    })
+      .then(async (res) => {
+        console.log(
+          `Metadata update response: ${res.status} ${res.statusText}`,
+        );
+        console.log(res);
```
jpwilliams: Dangling logs. We could use `client.logger.*()` if we do still need to add something.

packages/inngest/src/experimental.ts
```diff
@@ -2,7 +2,11 @@
...
 export { getAsyncCtx } from "./components/execution/als.ts";
-export type { ExtendedTracesMiddlewareOptions } from "./components/execution/otel/middleware.ts";
+
 // Extended Traces (OpenTelemetry)
+export type { ExtendedTracesMiddlewareOptions } from "./components/execution/otel/middleware.ts";
 export { extendedTracesMiddleware } from "./components/execution/otel/middleware.ts";
 export { PublicInngestSpanProcessor as InngestSpanProcessor } from "./components/execution/otel/processor.ts";
+
+// Step Metadata
+export { metadataMiddleware } from "./components/InngestMetadata.ts";
```
jpwilliams, line 12: Any thoughts on this entrypoint? It feels a bit clunky.

I almost wonder if we should also further separate it e.g. `"inngest/experimental/metadata"`, otherwise all of the experimental features are all bundled together.

packages/inngest/src/components/execution/InngestExecution.ts
```diff
@@ -144,4 +149,12 @@ export class InngestExecution {
 export interface IInngestExecution {
   version: ExecutionVersion;
   start(): Promise<ExecutionResult>;
+
+  addMetadata(
+    stepID: string,
```
jpwilliams: Nit: casing
```suggestion
    stepId: string,
```

packages/inngest/src/components/execution/v1.ts
```diff
@@ -164,6 +170,25 @@ class V1InngestExecution extends InngestExecution implements IInngestExecution {
...
+    values: Record<string, unknown>,
+  ) {
+    if (!this.state.metadata) {
+      this.state.metadata = new Map();
+    }
+
+    if (!this.state.metadata.has(stepId)) {
+      this.state.metadata.set(stepId, []);
+    }
+    this.state.metadata.get(stepId)!.push({ kind, scope, op, values });
```
jpwilliams: Can we do an extra check here to ensure we don't need the `!`?

packages/inngest/src/components/execution/v2.ts
```diff
@@ -145,6 +151,25 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
...
+    values: Record<string, unknown>,
+  ) {
+    if (!this.state.metadata) {
+      this.state.metadata = new Map();
+    }
+
+    if (!this.state.metadata.has(stepId)) {
+      this.state.metadata.set(stepId, []);
+    }
+    this.state.metadata.get(stepId)!.push({ kind, scope, op, values });
```
jpwilliams: Can we do an extra check here to avoid the `!`?

packages/inngest/src/components/Inngest.ts
```diff
@@ -169,6 +174,12 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
 
   private _appVersion: string | undefined;
 
+  /**
+   * @internal
+   * Flag set by metadataMiddleware to enable step.metadata()
+   */
+  public _experimentalMetadataEnabled = false;
```
jpwilliams: Can we make this `protected` or `private` instead?

- `protected` would mean it's still typed as a `boolean` for the user
- `private` would resolve to `any` for the user

Either one ensures it's not exposed directly if we're initially using middleware to enable the feature.

Also, that means there's no need for the `_`; it can just be `experimentalMetadataEnabled`.

packages/inngest/src/components/Inngest.ts
```diff
@@ -185,6 +196,28 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+   * ```
+   */
+  get metadata(): MetadataBuilder {
+    if (!this._experimentalMetadataEnabled) {
+      throw new Error(
+        'inngest.metadata is experimental. Enable it by adding metadataMiddleware() from "inngest/experimental" to your client middleware.',
+      );
+    }
+    return new UnscopedMetadataBuilder(this);
+  }
```
jpwilliams, line 223: Beautifully simple! 🙌

packages/inngest/src/components/Inngest.ts
```diff
@@ -477,6 +510,36 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
     );
   }
 
+  // biome-ignore lint/correctness/noUnusedPrivateClassMembers: used in the SDK
+  private async _updateMetadata({
```
jpwilliams: No need for the `_` here.

Other methods in this class have `_`, but only because there's also a public facing method without that, e.g. `send()` and `_send()`.

We usually do this to have some internal logic that we can run from elsewhere, but also have a particular method that is user-facing.

packages/inngest/src/components/Inngest.ts
```diff
@@ -477,6 +510,36 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+      values: Record<string, unknown>;
+    }>;
+    headers?: Record<string, string>;
+  }): Promise<void> {
+    const res = await this.inngestApi.updateMetadata(
+      { target, metadata },
+      { headers: { ...this.headers, ...headers } },
+    );
+    if (res.ok) {
+      console.log("We got this response: ", res);
```
jpwilliams: Dangling logs. We could use `client.logger.*()` if we do still need to add something.

packages/inngest/src/components/Inngest.ts
```diff
@@ -477,6 +510,36 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+      return res.value;
+    }
+
+    console.log("Failed to update metadata!");
```
jpwilliams: Dangling logs. We could use `client.logger.*()` if we do still need to add something.

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -683,6 +749,26 @@ export const createStepTools = <TClient extends Inngest.Any>(
...
+     * await step.metadata("update-status").update({ status: "processing" });
+     *
+     * // Update metadata for a different run
+     * await step.metadata("notify-parent")
+     *   .run(parentRunId)
+     *   .update({ childCompleted: true });
+     * ```
+     */
+    metadata: (memoizationId: string): MetadataStepTool =>
+      createStepMetadataWrapper(memoizationId),
```
jpwilliams: Fantastic! Did what you needed to do here make sense? I'd quite like us to move to this in v4 🤔

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+  ): Omit<MetadataBuilder<Extras>, "run" | "step">;
+  attempt(
+    index?: number,
+  ): Omit<MetadataBuilder<Extras>, "run" | "step" | "attempt">;
+  span(
+    id?: string,
+  ): Omit<MetadataBuilder<Extras>, "run" | "step" | "attempt" | "span">;
+  update(values: Record<string, unknown>, kind?: string): Promise<void>;
+  set(values: Record<string, unknown>, kind?: string): Promise<void>;
+  delete(values: string[], kind?: string): Promise<void>;
```
jpwilliams: Are we having multiple ops initially? Could `delete` and `set` be removed?

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+import { InngestMiddleware } from "./InngestMiddleware.ts";
+export interface BuilderConfig {
+  runId?: string | null;
+  stepId?: string | null;
+  stepIndex?: number;
+  attempt?: number | null;
+  spanId?: string;
+}
+
+export type MetadataScope = "run" | "step" | "step_attempt" | "extended_trace";
```
jpwilliams, line 11: Could every `export` in here have a comment? Generally just knowing how the various types/methods are intended to be used is enough!

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+import type { MetadataTarget } from "../types.ts";
+import type { Inngest } from "./Inngest.ts";
+import { InngestMiddleware } from "./InngestMiddleware.ts";
+export interface BuilderConfig {
+  runId?: string | null;
+  stepId?: string | null;
+  stepIndex?: number;
+  attempt?: number | null;
+  spanId?: string;
+}
```
jpwilliams: It's great that this came out as simply as you folks had first envisioned! 🙌

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+
+export type MetadataUpdate = {
+  kind: MetadataKind;
+  scope: MetadataScope;
+  op: MetadataOpcode;
+  values: Record<string, unknown>;
+};
+
+export type MetadataBuilder<Extras = {}> = {
+  run(id?: string): Omit<MetadataBuilder<Extras>, "run">;
```
jpwilliams: Wondering about the types here - this will be the main type that users interact with.

It's awesome that it's just a simple type that's away from the implementation, though the `Omit<>` means that it's a little hard to read, as we see signatures for methods that won't exist.

Image

We could use `Simplify<>`, which resolves types a little deeper, e.g. `Simplify<Omit<MetadataBuilder<Extras>, "run" | "step">>`. We end up with something a little more readable and explicit, but also maybe a bit too verbose.

https://github.com/inngest/inngest-js/blob/94d80ab4ffbd4c72d35b0033e2afcee63b425f2b/packages/inngest/src/helpers/types.ts#L250

Image

Alternatively we could set the `get metadata()` type on the `Inngest` client to return an `UnscopedMetadataBuilder` directly. That means the signatures are always very clean, but then the user has to type or go to the implementation to see what to do, which isn't great.

What do you think? I'm leaning towards `Simplify<>` 🤔

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -0,0 +1,299 @@
...
+  step(
+    id?: string,
+    index?: number,
+  ): Omit<MetadataBuilder<Extras>, "run" | "step">;
+  attempt(
```
jpwilliams: Each of the methods in this builder needs comments for the user, as this will be the main things they interact with.

packages/inngest/src/components/InngestMetadata.test.ts
```diff
@@ -0,0 +1,228 @@
...
+const mockClient = () =>
+  ({
+    _updateMetadata: vi.fn().mockResolvedValue(undefined),
+  }) as unknown as Inngest;
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe("buildTarget", () => {
```
jpwilliams, line 51: Fantastic tests! 🙌🙌🙌

Can we also add some type tests? e.g.

https://github.com/inngest/inngest-js/blob/94d80ab4ffbd4c72d35b0033e2afcee63b425f2b/packages/inngest/src/components/EventSchemas.test.ts#L33-L43

Vitest should already be set up to do this. Ideally we can simply test that the builder scoping works nicely, as that's the main complexity to make sure we don't break.

Some docs for it here: https://vitest.dev/guide/testing-types

packages/inngest/src/types.ts
rhino1998: Yeah, that's what I get for writing comments at night

packages/inngest/src/components/InngestMetadata.ts
rhino1998: [dff156d](https://github.com/inngest/inngest-js/pull/1176/commits/dff156d605bdb7518d96616acfb4627f45eecebe)

I wasn't aware that something like simplify worked like that. Very cool

packages/inngest/src/components/InngestMetadata.ts
Linell: Yup, literally just gating it similarly to extended traces so that it's more explicitly opt-in for the time being. Definitely open to better ways to do that!

packages/inngest/src/types.ts
rhino1998: [e8b1782](https://github.com/inngest/inngest-js/pull/1176/commits/e8b17828451e283ce798f3b94a68463837372606)

packages/inngest/src/components/execution/InngestExecution.ts
rhino1998: [df27157](https://github.com/inngest/inngest-js/pull/1176/commits/df2715799f494bc985473d4816729166fa14d1e8)

packages/inngest/src/components/execution/v2.ts
rhino1998: [bea62fb](https://github.com/inngest/inngest-js/pull/1176/commits/bea62fb4814615dc0d105ef3c0059eaba27ec6da)

packages/inngest/src/components/InngestMetadata.ts
rhino1998: Removed them for now [e9efa92](https://github.com/inngest/inngest-js/pull/1176/commits/e9efa923c2d6d8523876e988113367218274fc38)

packages/inngest/src/api/api.ts
rhino1998: Nice, that simplifies things [e666e80](https://github.com/inngest/inngest-js/pull/1176/commits/e666e807722723886b028adf28167adda06eb58a)

packages/inngest/src/components/execution/v1.ts
rhino1998: [bea62fb](https://github.com/inngest/inngest-js/pull/1176/commits/bea62fb4814615dc0d105ef3c0059eaba27ec6da)

packages/inngest/src/components/Inngest.ts
rhino1998: [c8e909e](https://github.com/inngest/inngest-js/pull/1176/commits/c8e909ec67a68ae665495d8f054655747f6c593a)

packages/inngest/src/components/Inngest.ts
rhino1998: [3e0d27e](https://github.com/inngest/inngest-js/pull/1176/commits/3e0d27e1bc72bb75e022fe4c2da0e24c515730f7)

packages/inngest/src/components/InngestMetadata.test.ts
rhino1998: Addes some type tests [a805b00](https://github.com/inngest/inngest-js/pull/1176/commits/a805b00bf11df271d783389addd68f38f0cf2304)

jpwilliams (approved):

PR #1204, opened 2025-12-18
fix(extended-traces): include client headers in trace exports

Linell:
## Summary

Include client headers in trace exports, which should hopefully help with extended traces not showing up as expected in branch environments because they depend on `x-inngest-env`:

> If you are sending events without an Inngest SDK, you'll need to pass the x-inngest-env header along with your request. For more information about this and sending events from any environment with the Event API, [read the send() reference](https://www.inngest.com/docs/reference/events/send#send-events-via-http-event-api).

<small>[Relevant documentation.](https://www.inngest.com/docs/platform/environments#branch-environments). </small>

### DIFF
--- a/packages/inngest/src/components/execution/otel/processor.ts
+++ b/packages/inngest/src/components/execution/otel/processor.ts
@@ -289,6 +289,7 @@ export class InngestSpanProcessor implements SpanProcessor {
             url: url.href,
 
             headers: {
+              ...app["headers"],
               Authorization: `Bearer ${app["inngestApi"]["signingKey"]}`,
             },
           });


djfarrelly (approved):
This should fix it - the request needs both the environment headers for API requests for branch environments, otherwise it will just goto the "branch parent."

It would be nice to improve this to add either tests or some other first class method that would be able to catch this with type checking if we refactor in the future. I might anticipate a regression as we're using a "private" property via bracket notiation workaround.

I think we also could ship this in a bug fix then land the refactor after to get this out ASAP.

jpwilliams (approved):
I think this would also need another patch (doesn't have to be this PR) to handle environments where we receive environment variables with the request, such as Cloudflare Workers.

In this case, the `InngestCommHandler` (serve handler) sets the `mode` of the client which also contains `env` with the delayed env vars.

Maybe `Inngest#headers` should be a `getter` that calculates the headers using the latest env every time, which should make sure we're always getting the latest here. 🤔

PR #1211, opened 2025-12-26
Emit warnings as metadata

rhino1998:
## Summary

This wraps executor uses of console.warn to also emit metadata warnings when metadata is enabled

## Checklist

~- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR~
~- [ ] Added unit/integration tests~

## Related

- EXE-944

### DIFF
--- a/packages/inngest/src/components/InngestFunction.test.ts
+++ b/packages/inngest/src/components/InngestFunction.test.ts
@@ -382,7 +382,7 @@ describe("runFn", () => {
                 describe("warning logs", () => {
                   t.expectedWarnings?.forEach((warning, i) => {
                     test(`warning log #${i + 1} includes "${warning}"`, () => {
-                      expect(globalThis.console.warn).toHaveBeenNthCalledWith(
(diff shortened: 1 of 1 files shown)


packages/inngest/src/components/Inngest.ts
```diff
@@ -540,6 +544,30 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
     );
   }
 
+  // biome-ignore lint/correctness/noUnusedPrivateClassMembers: used in the SDK
+  private async warnMetadata(
+    target: MetadataTarget,
+    kind: ErrCode,
+    text: string,
+  ) {
+    console.warn(text);
```
jpwilliams: Could we use `this.logger` here? There's a ticket somewhere for ensuring these warnings come through custom loggers if they're specified, so we can solve two problems in one here!

rhino1998: [1e82917](https://github.com/inngest/inngest-js/pull/1211/commits/1e82917135c7e5ea2c14f3471a9688c2a922bfa3)

PR #1219, opened 2026-01-05
fix: step retry handling for sync execution

zshannon:
## Summary

Fix step retry handling for sync execution mode. Previously, step errors with cached state were always marked as non-retriable, causing functions to fail prematurely. Now the SDK correctly:

- Only marks step errors as non-retriable on the final attempt
- Re-runs the step function when cached error exists but retries remain

### Related server issues discovered during testing

While testing with the experimental sync endpoint wrapper (`createExperimentalEndpointWrapper`), we discovered several server-side issues that should be tracked/fixed independently:

1. **Missing signature header**: The dev server doesn't send the `X-Inngest-Signature` header on sync mode callbacks, causing [`isInngestReq()`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/InngestCommHandler.ts#L695-L708) to return false and creating infinite loops.

```typescript
// InngestCommHandler.ts:695-708
private async isInngestReq(actions): Promise<boolean> {
  const [runId, signature] = await Promise.all([
    actions.headers(reqMessage, headerKeys.InngestRunId),
    actions.headers(reqMessage, headerKeys.Signature),
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//  🚨 ISSUE: Dev server sends empty/missing signature header
  ]);
  return Boolean(runId && typeof signature === "string");
//                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                        🚨 ISSUE: Returns false - request not recognized as from Inngest
}
```

2. **Missing attempt count**: The server doesn't return the current attempt number in sync mode callbacks. The `attempt` field is always 0, making it impossible for the SDK to know which retry attempt it's on. The Go SDK has the same limitation:

```go
// https://github.com/inngest/inngestgo/blob/main/step/run.go#L197-L210
func shouldUseStepFailed(mgr sdkrequest.InvocationManager) bool {
    callCtx := mgr.Request().CallCtx
    if callCtx.MaxAttempts == nil {
        return false
    }
    // ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
    // XXX: When using V2 API, there is no endpoint that returns response
    // with attempt count. We need to implement this as a new API endpoint
    // first before sync step.run is officially supported.
    // ⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
    return callCtx.Attempt+1 >= *callCtx.MaxAttempts
//         ^^^^^^^^^^^^^^^^
//         🚨 ISSUE: Attempt is always 0 in sync mode
}
```

3. **Missing step metadata**: The server callback payload needs to include `max_attempts` and current `attempt` for proper retry handling. The SDK parses these from the request [`ctx`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/helpers/functions.ts#L146-L168) but the values aren't populated in sync mode callbacks:

```typescript
// helpers/functions.ts:138-171 (V1 execution version schema)
[ExecutionVersion.V1]: () => ({
  version: ExecutionVersion.V1,
  ...z.object({
    event: z.record(z.any()),
    events: z.array(z.record(z.any())).default([]),
    steps: stepsSchemas[ExecutionVersion.V1],
    ctx: z.object({
      run_id: z.string(),
      fn_id: z.string().optional(),
      attempt: z.number().default(0),
//    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//    🚨 ISSUE: Always 0 - server doesn't send actual attempt count in sync mode
      max_attempts: z.number().optional(),
//    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//    🚨 ISSUE: Always undefined - server doesn't send this in sync mode
      disable_immediate_execution: z.boolean().default(false),
      use_api: z.boolean().default(false),
      qi_id: z.string().optional(),
      stack: z.object({
        stack: z.array(z.string()).nullable().transform((v) => (Array.isArray(v) ? v : [])),
        current: z.number(),
      }).passthrough().optional().nullable(),
    }).optional().nullable(),
  }).parse(data),
}) as const,
```

## Checklist

- [ ] ~~Added a [docs PR](https://github.com/inngest/website) that references this PR~~ N/A No docs changes needed

### DIFF
--- a/.changeset/fix-step-retry-handling.md
+++ b/.changeset/fix-step-retry-handling.md
@@ -1,0 +1,8 @@
+---
+"inngest": patch
+---
+
+Fix step retry handling for sync execution mode
+
+- Only mark step errors as non-retriable on final attempt
+- Re-run step function when cached error exists but retries remain


jpwilliams:
Thanks for this @zshannon!

I'm replacing this with #1292 as this shouldn't really be done in the SDK; we need it outside to ensure retries are safer and more durable!

PR #1229, opened 2026-01-08
Add new "event type" schema approach

amh4r:
## Summary

Add new "decentralized event schemas" approach. This is the replacement for v3's centralized `schema` field on the Inngest client.

"Event types" can be defined without a schema (e.g. `eventType("my-event")`) or with a schema (`e.g. `eventType("my-event", { schema: z.object({ msg: z.string }) })`). These schemas:
- Have correct static types
- Are validated at runtime: event producing (i.e. `Inngest.send()`), event receiving (i.e. running an Inngest function), and `step.waitForEvent`.

Utests in `packages/inngest/src/components/triggers` document static types and event producing validation. Some new itests prove that runtime validation works.

## Checklist

## Related
https://linear.app/inngest/issue/EXE-1138

https://www.notion.so/inngest/Event-schemas-TS-SDK-v4-2e6b64753bbd8072b818e917c8fe1192

### DIFF
--- a/packages/inngest/src/types.ts
+++ b/packages/inngest/src/types.ts
@@ -995,20 +995,14 @@ export interface RegisterOptions {
    * restrictive request timeouts and other limitations. It is only available if
    * the serve handler being used supports streaming.
    *
-   * If this is `"allow"`, the SDK will attempt to stream responses back to
-   * Inngest if it can confidently detect support for it by verifyng that the
-   * platform and the serve handler supports streaming.
-   *
-   * If this is `"force"`, the SDK will always attempt to stream responses back
-   * to Inngest regardless of whether we can detect support for it or not. This
-   * will override `allowStreaming`, but will still not attempt to stream if
-   * the serve handler does not support it.
+   * If this is `"true"`, the SDK will attempt to stream responses back
+   * to Inngest. If the serve handler does not support streaming, an error will be thrown.
    *
    * If this is `false`, streaming will never be used.
    *
    * Defaults to `false`.
    */
-  streaming?: "allow" | "force" | false;
+  streaming?: true | false;
 }
 
 /**
--- a/packages/inngest/MIGRATION.md
+++ b/packages/inngest/MIGRATION.md
@@ -79,6 +79,22 @@ app.use("/api/inngest", serve({ client: inngest, functions }));
 > [!NOTE]
 > If you were relying on environment variables (e.g., `INNGEST_SIGNING_KEY`) rather than passing these options explicitly, no changes are required—the client will automatically read from the environment.
 
+## Streaming Option Simplified
+
+The `streaming` option in `serve()` has been simplified from `"allow" | "force" | false` to `true | false`.
+
+- `"force"` → `true` (enable streaming; throws error if handler doesn't support it)
+- `"allow"` → removed (use `true` instead)
+- `false` → `false` (unchanged)
+
+```typescript
+// Old (v3)
+serve({ client, functions, streaming: "force" });
+
+// New (v4)
+serve({ client, functions, streaming: true });
+```
+
 ## Edge Environment Improvements
 
 In v4, fetch and configuration are now resolved lazily at first use rather than eagerly at client construction. This means you no longer need to manually bind `globalThis.fetch` before creating an Inngest client in edge environments (Cloudflare Workers, Vercel Edge, Deno, etc.).
--- a/packages/inngest/src/components/InngestCommHandler.test.ts
+++ b/packages/inngest/src/components/InngestCommHandler.test.ts
@@ -1,3 +1,5 @@
+import httpMocks from "node-mocks-http";
+import { envKeys, headerKeys } from "../helpers/consts.ts";
 import { InngestCommHandler } from "../index.ts";
 import { serve } from "../next.ts";
 import { createClient } from "../test/helpers.ts";
@@ -30,6 +32,97 @@ describe("ServeHandler", () => {
       ] as const;
 
       serve({ client: inngest, functions });
+    });
+  });
+
+  describe("streaming: force validation", () => {
+    const inngest = createClient({ id: "test", isDev: true });
+
+    const fn = inngest.createFunction(
+      { id: "test" },
+      { event: "demo/event.sent" },
+      () => "test",
+    );
+
+    const runHandler = async (
+      handler: ReturnType<typeof serve>,
+      opts?: {
+        env?: Record<string, string>;
+        actionOverrides?: Record<string, unknown>;
+      },
+    ) => {
+      const body = {
+        ctx: { fn_id: "test-test", run_id: "run-123", step_id: "step" },
+        event: { name: "demo/event.sent", data: {} },
+        events: [{ name: "demo/event.sent", data: {} }],
+        steps: {},
+        use_api: false,
+      };
+
+      const req = httpMocks.createRequest({
+        method: "POST",
+        url: "/api/inngest?fnId=test-test&stepId=step",
+        headers: {
+          host: "localhost:3000",
+          "content-type": "application/json",
+          "content-length": `${JSON.stringify(body).length}`,
+          [headerKeys.InngestRunId]: "run-123",
+          [headerKeys.Signature]: "",
+        },
+        body,
+      });
+      const res = httpMocks.createResponse();
+
+      const prevEnv = process.env;
+      if (opts?.env) {
+        process.env = { ...prevEnv, ...opts.env };
+      }
+
+      try {
+        const args: unknown[] = [req, res];
+        if (opts?.actionOverrides) {
+          args.push({ actionOverrides: opts.actionOverrides });
+        }
+
+        await (handler as (...args: unknown[]) => Promise<unknown>)(...args);
+
+        return {
+          status: res.statusCode,
+          body: res._getData(),
+        };
+      } finally {
+        process.env = prevEnv;
+      }
+    };
+
+    test("throws error when streaming is true but handler doesn't support it", async () => {
+      const handler = serve({
+        client: inngest,
+        functions: [fn],
+        streaming: true,
+      });
+
+      const result = await runHandler(handler, {
+        actionOverrides: { transformStreamingResponse: undefined },
+      });
+
+      expect(result.status).toBe(500);
+      expect(result.body).toMatch(/streaming/i);
+    });
+
+    test("throws error when INNGEST_STREAMING=true env var but handler doesn't support it", async () => {
+      const handler = serve({
+        client: inngest,
+        functions: [fn],
+      });
+
+      const result = await runHandler(handler, {
+        env: { [envKeys.InngestStreaming]: "true" },
+        actionOverrides: { transformStreamingResponse: undefined },
+      });
+
+      expect(result.status).toBe(500);
+      expect(result.body).toMatch(/streaming/i);
     });
   });
 });
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -474,7 +474,7 @@ export class InngestCommHandler<
 
     const defaultStreamingOption: typeof this.streaming = false;
     this.streaming = z
-      .union([z.enum(["allow", "force"]), z.literal(false)])
+      .boolean()
       .default(defaultStreamingOption)
       .catch((ctx) => {
         this.log(
@@ -486,7 +486,9 @@ export class InngestCommHandler<
 
         return defaultStreamingOption;
       })
-      .parse(options.streaming || this.env[envKeys.InngestStreaming]);
+      .parse(
+        options.streaming || parseAsBoolean(this.env[envKeys.InngestStreaming]),
+      );
 
     // Early validation for environments where process.env is available (Node.js).
     // Edge environments will skip this and validate at request time instead.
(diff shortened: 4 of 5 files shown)


packages/inngest/src/components/triggers/triggers.ts
```diff
@@ -0,0 +1,206 @@
...
+        v?: string;
+      }
+    : {
+        data: TData;
+        id?: string;
+        ts?: number;
+        v?: string;
+      };
+
+type ExtractSchema<TData> = TData extends StandardSchemaV1<infer IData, infer _>
```
jpwilliams: I believe `StandardSchema` provides access to separate I/O typing, as the input type for an event (when `.create()`ing it) may be different than the output. i.e. folks could use things like `z.transform(...)`.

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,311 @@
...
+      expect(et.event).toBe("event-1");
+      expectTypeOf(et.event).toEqualTypeOf<"event-1">();
+
+      expect(et.name).toBe("event-1");
+      expectTypeOf(et.name).toEqualTypeOf<"event-1">();
+
+      expect(et.schema).toBeDefined();
+      expectTypeOf(et.schema).toExtend<
+        StandardSchemaV1<{ message: string }, { message: string }>
+      >();
```
jpwilliams: FYI `any` extends anything; if `et.schema` is resolving to `any` accidentally, then this test will still pass.

I usually also do an additional `any` test, probably:
```ts
expectTypeOf(et.schema).not.toBeAny();
```

I believe `expectTypeOf().toEqualTypeOf()` is fine and this is only a problem with checking `extends`.

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+  test("createFunction", () => {
+    const inngest = new Inngest({ id: "app" });
+    inngest.createFunction({ id: "fn" }, cron("* * * * *"), ({ event }) => {
+      expectTypeOf(event.name).not.toBeAny();
+      expectTypeOf(event.name).toEqualTypeOf<
+        "inngest/scheduled.timer" | "inngest/function.invoked"
+      >();
+
+      expectTypeOf(event.data).not.toBeAny();
+      expectTypeOf(event.data).toEqualTypeOf<{}>();
```
jpwilliams: Doesn't `"inngest/scheduled.timer"` have a payload?

https://github.com/inngest/inngest-js/blob/ad044e05cf0aac7e593e753f73782d6b24bab0cc/packages/inngest/src/types.ts#L214-L227

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+  });
+
+  test("schema transform", async () => {
+    eventType("event-1", {
+      // @ts-expect-error - Transforms are not supported
+      schema: z.object({ input: z.string() }).transform((val) => {
+        return { output: val.input.length };
+      }),
+    });
+  });
```
jpwilliams, line 91: Nice! Thank you for this

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+      expectTypeOf(event.data).toEqualTypeOf<Record<string, any>>();
+    });
+
+    // With condition
+    inngest.createFunction(
+      { id: "fn2" },
+      {
+        event: et,
+        if: "event.data.foo == 'bar'",
+        schema: et.schema,
```
jpwilliams: Why do we pass the schema in too?

Can/does `inngest.createFunction()` get the schema from `event: et`?

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+
+    // Without condition
+    inngest.createFunction(
+      {
+        id: "fn",
+        cancelOn: [et],
+      },
+      et,
+      () => {},
+    );
```
jpwilliams, line 137: Makes me really want the unified handler stuff as this is weird to read.

```ts
inngest.createFunction({
  id: "fn",
  cancelOn: [et],
  trigger: et,
  handler: () => {},
});
```

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+        v?: string;
+      } | null>();
+    });
+  });
+});
+
+describe("eventType with schema", () => {
+  const et = eventType("event-1", {
+    schema: z.object({ msg: z.string() }),
+  });
```
jpwilliams, line 188: A typeguard against folks adding schemas which are incompatible with `data` would be interesting (but not required). i.e. `data` must be an object (the platform enforces this), but technically someone could add `schema: z.string()`.

A guard against this to show a type error might be nice.

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+    expectTypeOf(et.name).not.toBeAny();
+    expectTypeOf(et.name).toEqualTypeOf<"event-1">();
+
+    expect(et.schema).toBeDefined();
+    expectTypeOf(et.schema).not.toBeAny();
+    expectTypeOf(et.schema).toExtend<StandardSchemaV1<{ msg: string }>>();
+  });
+
+  test("create", async () => {
+    const created1 = et.create({ data: { msg: "hello" } });
```
jpwilliams, line 205: A final question on this - did we consider:
```ts
et.create({ msg: "hello" });
et.create({ msg: "hello" }, { id: "123", v: "1.0.0" });
```

i.e. cater for the common case (just setting data) with the first arg and all other settings come in the second.

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+    expectTypeOf(created2.data).toEqualTypeOf<{ msg: string }>();
+    expect(created2.id).toBe("123");
+    expectTypeOf(created2.id).not.toBeAny();
+    expectTypeOf(created2.id).toEqualTypeOf<string | undefined>();
+    expect(created2.ts).toBe(1715769600);
+    expectTypeOf(created2.ts).not.toBeAny();
+    expectTypeOf(created2.ts).toEqualTypeOf<number | undefined>();
+    expect(created2.v).toBe("1.0.0");
+    expectTypeOf(created2.v).not.toBeAny();
+    expectTypeOf(created2.v).toEqualTypeOf<string | undefined>();
```
jpwilliams, line 227: Noting for posterity that I think it's fine that we don't literally infer these values.

packages/inngest/src/components/triggers/triggers.ts
```diff
@@ -0,0 +1,261 @@
...
+    const event: UnvalidatedCreatedEvent<TName, ExtractSchemaData<TSchema>> = {
+      name: this.name,
+      data: params.data as ExtractSchemaData<TSchema>,
+      id: params.id,
+      ts: params.ts,
+      v: params.v ?? this.version,
+
+      // Method for validating and transforming the event data against the
+      // schema
+      validate: async (): Promise<void> => {
```
jpwilliams, line 163: Should `validate()` return the created event?
It may help it be a nice shortcut.

Before:
```ts
const event = myEvent.create({ data: { foo: "bar" } });
await event.validate();
```

After:
```ts
const event = await myEvent.create({ data: { foo: "bar" } }).validate();
```

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+
+  test("create", () => {
+    et.create({});
+    et.create({ data: { foo: "bar" } });
+    et.create({
+      data: { foo: "bar" },
+      id: "123",
+      ts: 1715769600,
+      v: "1.0.0",
+    });
```
jpwilliams, line 81: These are type-only; do they need runtime assertions for the returned event too?

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+      expectTypeOf(event.name).not.toBeAny();
+      expectTypeOf(event.name).toEqualTypeOf<
+        "event-1" | "inngest/function.invoked"
+      >();
+      expectTypeOf(event.data).not.toBeAny();
+      expectTypeOf(event.data).toEqualTypeOf<{ msg: string }>();
+    });
+  });
+
+  test("step.invoke", () => {
```
jpwilliams, line 252: In this test, can we add cases where:
- `child` is triggered by an event with no schema; no `data` in the invoke call should be fine
- `child` is triggered by an event with a schema of `Record<string, any>`; no `data` in the invoke call should be fine

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -0,0 +1,569 @@
...
+          | "inngest/function.invoked"
+        >();
+
+        expectTypeOf(event.data).not.toBeAny();
+        expectTypeOf(event.data).toEqualTypeOf<
+          | { a: string }
+          | { b: number }
+          | { name: string }
+          | { age: number }
+          | {}
```
jpwilliams: Another couple of spots in this test that should probably be the scheduled timer payload (just `cron: string`):

https://github.com/inngest/inngest-js/blob/ad044e05cf0aac7e593e753f73782d6b24bab0cc/packages/inngest/src/types.ts#L214-L227

packages/inngest/src/types.ts
```diff
@@ -1568,19 +1619,23 @@ export type PayloadForAnyInngestFunction<
...
-    : Simplify<
-        Omit<
-          TEvents[EventNameFromTrigger<TEvents, ITriggers[number]>],
-          "name" | "ts"
-        >
-      >
+  ? // First check: Does this function have an invoke trigger with a schema?
+    // If so, only invoke schemas should be used (not eventType schemas)
+    HasInvokeTriggerWithSchema<ITriggers> extends true
+    ? { data: ExtractInvokeSchemaInput<ITriggers> }
```
jpwilliams, line 1626: Could we add a comment to this type explaining that it's to be used for invokes? As this type doesn't make too much sense for non-invoke payload fetching any more.

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -431,10 +433,20 @@ export const createStepTools = <TClient extends Inngest.Any>(
...
+      <
+        TOpts extends {
+          event:
+            | string
+            // biome-ignore lint/suspicious/noExplicitAny: Allow any schema
+            | EventType<string, any>;
+          timeout: number | string | Date;
+          // biome-ignore lint/suspicious/noExplicitAny: Allow any schema
+          schema?: StandardSchemaV1<any>;
+        } & ExclusiveKeys<{ match?: string; if?: string }, "match", "if">,
```
jpwilliams, line 504: This makes the type pretty messy; we're gonna remove `match` though as it's been deprecated for a while?

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -431,10 +433,20 @@ export const createStepTools = <TClient extends Inngest.Any>(
...
-      <IncomingEvent extends WithoutInternalStr<string>>(
+      <
+        TOpts extends {
+          event:
+            | string
+            // biome-ignore lint/suspicious/noExplicitAny: Allow any schema
+            | EventType<string, any>;
+          timeout: number | string | Date;
+          // biome-ignore lint/suspicious/noExplicitAny: Allow any schema
+          schema?: StandardSchemaV1<any>;
```
jpwilliams: These are now user-facing types; we need comments for these!

packages/inngest/src/components/InngestFunction.ts
```diff
@@ -350,7 +359,8 @@ export namespace InngestFunction {
    */
   export type Trigger<T extends string> = StrictUnion<
     | {
-        event: T;
+        // biome-ignore lint/suspicious/noExplicitAny: schema can be any StandardSchemaV1
+        event: T | EventType<T, any>;
```
jpwilliams: Nit: Could have `type AnyStandardSchema = StandardSchemV1<any>` somewhere so we don't have to have these exceptions; they're in a few spots.

packages/inngest/src/components/triggers/typeHelpers.ts
```diff
@@ -0,0 +1,363 @@
+// biome-ignore-all lint/suspicious/noExplicitAny: it's fine
```
jpwilliams, line 1: 😂

packages/inngest/src/components/triggers/triggers.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    return event;
+  }
+}
+
+/**
+ * This type's only purpose is to clearly highlight static type error messages
+ * in our codebase. To end users, it's exactly the same as a normal string.
+ */
+type StaticTypeError<TMessage extends string> = TMessage;
```
jpwilliams, line 196: Awesome!

packages/inngest/src/components/execution/v0.ts
```diff
@@ -157,6 +159,22 @@ export class V0InngestExecution
...
+                const eventData: unknown = incomingOp.data;
+                await validateEvents(
+                  // @ts-expect-error - This is a full event object at runtime
+                  [eventData],
+
+                  [waitForEventOpts],
+                );
+              } catch (err) {
+                this.state.currentOp.reject(err);
+                continue;
```
jpwilliams: I don't think we need the `continue` here, but we have questions about v0.

A thought is that we may keep this around for "strict," but tbh refactoring the exe versions would be nice, so I think it dies anyway.

packages/inngest/src/components/triggers/trigger.test.ts
amh4r: Ah yea it does. Thanks!

packages/inngest/src/components/triggers/trigger.test.ts
amh4r: Oops, no this isn't supported -- bad copy-paste. It doesn't do anything

packages/inngest/src/components/triggers/trigger.test.ts
amh4r: Yea same, that's definitely better. But I think that's a separate convo we should have. Is its benefit worth increasing upgrade friction? Maybe

packages/inngest/src/components/triggers/trigger.test.ts
amh4r: I'm open to that. It does look really wonky if you specify the options, but the vast majority of events are prob data only

packages/inngest/src/components/triggers/triggers.ts
amh4r: Good point, I'll do that

packages/inngest/src/components/triggers/trigger.test.ts
amh4r: We can't do a runtime assertion, but it currently gives a static type error if you don't specify an object. I'll add a test proving that

packages/inngest/src/components/InngestStepTools.ts
amh4r: What should the comment be? I'm a little unclear on that

packages/inngest/src/components/execution/v0.ts
amh4r: Ah true. I was trying to avoid the resolving a few lines later, but they'll noop right?

PR #1230, opened 2026-01-08
Move Serve Options to Client

Linell:
## Summary

Move many of the options previously provided to `serve` to the `client` level instead, because they make more sense there.

## Checklist

- [ ] Confirm Cloudflare still works correctly
- [ ] Confirm Connect still works correctly

## Related
[EXE-1119: Move some `serve` options to the client](https://linear.app/inngest/issue/EXE-1119/move-some-serve-options-to-the-client)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Node packages
 node_modules
 
+# Local Inngest dev server
+.inngest/
+
 # Nix
 /.direnv/
 
--- a/packages/inngest/MIGRATION.md
+++ b/packages/inngest/MIGRATION.md
@@ -1,0 +1,57 @@
+# v3 -> v4 Migration Guide
+
+This guide covers how to migrate between the v3 and the v4 version of the `inngest` package.
+
+> [!WARNING]
+> This migration guide is a work in progress.
+
+## Serve Options Moved to Client
+
+Many of the options previously passed to the `serve` function have been moved up to the `client` level. These properties make more sense at this level and, because it only involves potentially reorganizing where you're setting values, should be a very straightforward migration.
+
+The options that you may need to reorganize are:
+
+- **baseUrl**
+- **fetch**
+- **logLevel**
+- **signingKey**
+- **signingKeyFallback**
+
+If you are passing any of these values to the `serve` function, or the `createServer` function, you will need to modify your code so that they are instead provided to the client.
+
+```typescript
+// Old (v3)
+import { Inngest } from "inngest";
+import { serve } from "inngest/express";
+
+const inngest = new Inngest({ id: "my-app" });
+
+app.use(
+  "/api/inngest",
+  serve({
+    client: inngest,
+    functions,
+    signingKey: "my-signing-key",
+    signingKeyFallback: "my-fallback-key",
+    logLevel: "debug",
+    baseUrl: "https://my-inngest-instance.example.com",
+  })
+);
+
+// New (v4)
+import { Inngest } from "inngest";
+import { serve } from "inngest/express";
+
+const inngest = new Inngest({
+  id: "my-app",
+  signingKey: "my-signing-key",
+  signingKeyFallback: "my-fallback-key",
+  logLevel: "debug",
+  baseUrl: "https://my-inngest-instance.example.com",
+});
+
+app.use("/api/inngest", serve({ client: inngest, functions }));
+```
+
+> [!NOTE]
+> If you were relying on environment variables (e.g., `INNGEST_SIGNING_KEY`) rather than passing these options explicitly, no changes are required—the client will automatically read from the environment.
--- a/packages/inngest/src/astro.ts
+++ b/packages/inngest/src/astro.ts
@@ -49,7 +49,6 @@ export const serve = (
 } => {
   const commHandler = new InngestCommHandler({
     frameworkName,
-    fetch: fetch.bind(globalThis),
     ...options,
     handler: ({ request: req }: { request: Request }) => {
       return {
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -46,6 +46,8 @@ import {
   type FailureEventArgs,
   type Handler,
   type InvokeTargetFunctionDefinition,
+  type LogLevel,
+  logLevels,
   type MetadataTarget,
   type SendEventOutput,
   type SendEventResponse,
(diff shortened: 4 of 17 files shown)


packages/inngest/src/components/Inngest.ts
```diff
@@ -455,6 +521,44 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+    if (
+      this._signingKeyFallback === undefined &&
+      env[envKeys.InngestSigningKeyFallback]
+    ) {
+      this._signingKeyFallback = env[envKeys.InngestSigningKeyFallback];
+    }
+
+    if (this._logLevel === undefined) {
+      this.trySetLogLevel(env[envKeys.InngestLogLevel]);
+    }
```
amh4r, line 548: Nit: would a helper function simplify this a bit?

```ts
function firstTruthy<T>(...args: T[]): T | undefined {
  for (const arg of args) {
    if (arg) {
      return arg;
    }
  }
  return undefined;
}
```

```suggestion
    // Set signing key from env if not already set via options
    this._signingKey = firstTruthy( this._signingKey, env[envKeys.InngestSigningKey]);

    // Set signing key fallback from env if not already set via options
    this._signingKeyFallback = firstTruthy(this._signingKeyFallback, env[envKeys.InngestSigningKeyFallback]);

    this.trySetLogLevel(firstTruthy(this._logLevel, env[envKeys.InngestLogLevel]));
```

packages/inngest/src/astro.ts
```diff
@@ -49,7 +49,6 @@ export const serve = (
 } => {
   const commHandler = new InngestCommHandler({
     frameworkName,
-    fetch: fetch.bind(globalThis),
```
amh4r, line 52: Is this something that users need to manually do on the client in v4?

packages/inngest/src/components/Inngest.ts
```diff
@@ -200,6 +229,36 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
     return this._appVersion;
   }
 
+  /**
+   * The signing key used to authenticate requests to and from Inngest.
+   */
+  get signingKey(): string | undefined {
+    return this._signingKey ?? processEnv(envKeys.InngestSigningKey);
+  }
```
amh4r, line 237: It feels a little weird that `this._signingKey` isn't always the signing key. Should we instead handle the env var in the constructor? Like maybe `this.signingKey` doesn't need to be a getter

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -441,13 +445,7 @@ export class InngestCommHandler<
...
     this.frameworkName = options.frameworkName;
     this.client = options.client as Inngest.Any;
-
-    if (options.id) {
-      console.warn(
-        `${logPrefix} The \`id\` serve option is deprecated and will be removed in v4`,
-      );
-    }
-    this.id = options.id || this.client.id;
+    this.id = this.client.id;
```
amh4r, line 448: Nit: maybe we remove `this.id` and directly use `this.client.id` elsewhere in this class. Then it becomes clearer that the client is the source of truth

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -567,7 +547,6 @@ export class InngestCommHandler<
    */
   protected get apiBaseUrl(): string {
     return (
-      this._options.baseUrl ||
```
amh4r, line 570: I think there might be a bug in here. If we set the base URL on the client (which should take precedence over env vars), then we'll still return the env vars. Maybe we can delete this `apiBaseUrl` getter and instead directly use `this.client.apiBaseUrl`

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -583,7 +562,6 @@ export class InngestCommHandler<
    */
   protected get eventApiBaseUrl(): string {
     return (
-      this._options.baseUrl ||
```
amh4r, line 586: Similar thought as the `apiBaseUrl` getter

packages/inngest/src/types.ts
```diff
@@ -997,46 +1029,6 @@ export type LogLevel = (typeof logLevels)[number];
...
-   *
-   * You can retrieve your signing key from the Inngest UI inside the "Secrets"
-   * section at {@link https://app.inngest.com/secrets}. We highly recommend
-   * that you add this to your platform's available environment variables as
-   * `INNGEST_SIGNING_KEY`.
-   *
-   * If no key can be found, you will not be able to register your functions or
-   * receive events from Inngest.
-   */
-  signingKey?: string;
```
jpwilliams, line 1016: Hmm do we still want to provide the ability to overwrite them at `serve()`-time too?

Some folks will create functions with a very generic client and then share those in multiple apps; when this happens, they'll need to override any client-level settings.

packages/inngest/MIGRATION.md
```diff
@@ -0,0 +1,57 @@
+# v3 -> v4 Migration Guide
```
jpwilliams, line 1: Oh, nice to include this!

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Node packages
 node_modules
 
+# Local Inngest dev server
+.inngest/
```
jpwilliams, line 5: 🙌

packages/inngest/src/components/Inngest.ts
jpwilliams: Aye, same for a few of these getters - it'd make sense if `this._signingKey` being set was if it was _explicitly_ set, but also it could be set if we've found it in a delayed env?

packages/inngest/src/components/Inngest.ts
```diff
@@ -178,6 +180,33 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+
+  /**
+   * Validates and sets the log level, warning if invalid.
+   */
+  private trySetLogLevel(level: string | undefined): void {
+    if (!level) return;
+    if (logLevels.includes(level as LogLevel)) {
+      this._logLevel = level as LogLevel;
+    } else {
+      console.warn(`${logPrefix} Unknown log level "${level}"; ignoring`);
```
jpwilliams, line 206: I think this `console.warn()` is copied from elsewhere, but can we use `this.logger` instead?

packages/inngest/src/types.ts
amh4r: I think they should put client creation in a factory function. They already need to have a different app ID for each client anyway

Linell:
Closed in favor of the changes added here: https://github.com/inngest/inngest-js/pull/1232

PR #1232, opened 2026-01-09
Default to Cloud Mode

Linell:
## Summary

This modifies the SDK so that it defaults to 'cloud mode' unless it's specified otherwise, with the intention of ensuring that no user accidentally deploys a dev mode server to production.

~Also adds a warning that detects when a user is likely to be unintentionally using cloud mode and logs out a warning telling them how to fix their issue. I feel like I mess this up myself often enough that it's worth some sort of heads up for new users.~ We now throw an error if we detect that a user has misconfigured the environment.

## Checklist

## Related
[EXE-1124: Default to "cloud mode"](https://linear.app/inngest/issue/EXE-1124/default-to-cloud-mode)

### DIFF
--- a/packages/inngest/src/cloudflare.ts
+++ b/packages/inngest/src/cloudflare.ts
@@ -127,12 +127,6 @@ export const serve = (
 ) => Promise<Response>) => {
   const handler = new InngestCommHandler({
     frameworkName,
-
-    /**
-     * Assume that we want to override the `fetch` implementation with the one
-     * globally available in the Cloudflare env. Specifying it here will
-     * ensure we avoid trying to load a Node-compatible version later.
-     */
     ...options,
     handler: (...args: Either<PagesHandlerArgs, WorkersHandlerArgs>) => {
       const { req, env } = deriveHandlerArgs(args);
--- a/packages/inngest/MIGRATION.md
+++ b/packages/inngest/MIGRATION.md
@@ -5,6 +5,23 @@ This guide covers how to migrate between the v3 and the v4 version of the `innge
 > [!WARNING]
 > This migration guide is a work in progress.
 
+## Default Mode Changed to Cloud
+
+The default mode is now `cloud` instead of `dev`. This prevents accidental production deployments in development mode and aligns with all other Inngest SDKs.
+
+**What this means:**
+
+- In `cloud` mode, a signing key is required (via `INNGEST_SIGNING_KEY` or the `signingKey` option)
+- For local development, explicitly set `isDev: true` on your client or set `INNGEST_DEV=1`
+
+```typescript
+// Local development
+const inngest = new Inngest({ id: "my-app", isDev: true });
+
+// Production (signing key required via env or option)
+const inngest = new Inngest({ id: "my-app" });
+```
+
 ## Serve Options Moved to Client
 
 Many of the options previously passed to the `serve` function have been moved up to the `client` level. These properties make more sense at this level and, because it only involves potentially reorganizing where you're setting values, should be a very straightforward migration.
@@ -55,3 +72,7 @@ app.use("/api/inngest", serve({ client: inngest, functions }));
 
 > [!NOTE]
 > If you were relying on environment variables (e.g., `INNGEST_SIGNING_KEY`) rather than passing these options explicitly, no changes are required—the client will automatically read from the environment.
+
+## Edge Environment Improvements
+
+In v4, fetch and configuration are now resolved lazily at first use rather than eagerly at client construction. This means you no longer need to manually bind `globalThis.fetch` before creating an Inngest client in edge environments (Cloudflare Workers, Vercel Edge, Deno, etc.).


packages/inngest/src/components/Inngest.test.ts
```diff
@@ -65,9 +65,9 @@ describe("new Inngest()", () => {
       return inngest;
     };
 
-    test("should default to inferred dev mode", () => {
+    test("should default to inferred cloud mode", () => {
       const inngest = createTestClient();
-      expect(inngest["mode"].isDev).toBe(true);
+      expect(inngest["mode"].isCloud).toBe(true);
       expect(inngest["mode"].isExplicit).toBe(false);
```
amh4r: We can remove `isExplicit` too since modes will only be explicit 🙌

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -2444,6 +2395,18 @@ export class InngestCommHandler<
       );
     }
 
+    const isAuthError =
+      status === 401 || status === 403 || error.includes("signing key");
+
+    if (isAuthError && this._mode?.isCloud && !this.signingKey) {
+      this.log(
```
Linell: Do I still want this log, even? 🤔

packages/inngest/src/api/api.ts
```diff
@@ -50,11 +47,10 @@ const checkpointNewRunResponseSchema = z.object({
 
 export namespace InngestApi {
   export interface Options {
-    baseUrl?: string;
+    baseUrl?: string | (() => string);
     signingKey: string;
```
Linell: Do we need similar getters for the signing keys too? Since this is internal, could we also drop the string so that it's always a function.

packages/inngest/src/components/execution/otel/processor.ts
```diff
@@ -259,27 +254,16 @@ export class InngestSpanProcessor implements SpanProcessor {
...
-                devHost,
-                app["fetch"],
-              );
-              if (hasDevServer) {
-                url = new URL(path, devHost);
-              }
-            } else if (app["mode"]?.explicitDevUrl) {
-              url = new URL(path, app["mode"].explicitDevUrl.href);
-            }
+          let baseUrl = app.apiBaseUrl || defaultInngestApiBaseUrl;
```
Linell: Can we use the logic from the API here, since it's just building the same URL?

Linell: The closer everything can live to the client, the better.

packages/inngest/src/components/Inngest.ts
```diff
@@ -121,7 +118,7 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
     defaultInngestEventBaseUrl,
   );
 
-  private headers!: Record<string, string>;
+  // private headers!: Record<string, string>;
```
amh4r:
```suggestion
```

packages/inngest/src/components/Inngest.ts
```diff
@@ -146,7 +143,9 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
    * settings, but should still check for the presence of an environment
    * variable if it is not set.
    */
-  private _mode!: Mode;
+
+  // private _mode: Mode | undefined; // TODO: do we even need this?
```
amh4r:
```suggestion
```

packages/inngest/src/components/Inngest.ts
```diff
@@ -156,12 +155,75 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+   */
+  // TODO: this is really more internal, right?
+  get getExplicitDevUrl(): URL | undefined {
+    const devEnvValue = this._env[envKeys.InngestDevMode];
+    if (typeof devEnvValue !== "string" || !devEnvValue) {
+      return undefined;
+    }
+
+    try {
+      return new URL(devEnvValue);
```
amh4r: Since `URL` can be strict, we might need a little more logic. For example, `"0.0.0.0:8288"` errors:
```
> new URL("0.0.0.0:8288")
Uncaught TypeError: Invalid URL
    at new URL (node:internal/url:818:25) {
  code: 'ERR_INVALID_URL',
  input: '0.0.0.0:8288'
}
```

Given that this is a dev URL, I think it's OK to assume `http` if a scheme isn't present

packages/inngest/src/components/Inngest.ts
```diff
@@ -668,12 +725,12 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
 
     // When sending events, check if the dev server is available.  If so, use the
     // dev server.
-    let url = this.sendEventUrl.href;
+    const url = this.sendEventUrl.href;
```
amh4r: Doesn't look like this URL is using `eventBaseUrl` under the hood. I think that means we could send events to the wrong URL

packages/inngest/src/components/Inngest.ts
```diff
@@ -668,12 +725,12 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
 
     // When sending events, check if the dev server is available.  If so, use the
     // dev server.
```
amh4r:
```suggestion
```

We no longer do this check

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -2332,23 +2263,13 @@ export class InngestCommHandler<
...
-    if (inferredDevMode) {
-      const host = devServerHost(this.env);
-      const hasDevServer = await devServerAvailable(host, this.fetch);
-      if (hasDevServer) {
-        registerURL = devServerUrl(host, "/fn/register");
-      }
-    } else if (this._mode?.explicitDevUrl) {
+    // Use explicit dev server URL if provided via INNGEST_DEV=<url>
+    if (this.client.getExplicitDevUrl) {
       registerURL = devServerUrl(
```
amh4r: Do we need `devServerUrl` anymore? I think we can just use `this.client.apiBaseUrl`, since it's handling the Dev Server URL shenanigans

packages/inngest/src/components/Inngest.ts
```diff
@@ -156,12 +155,75 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
   protected experimentalMetadataEnabled = false;
 
-  get apiBaseUrl(): string | undefined {
-    return this._apiBaseUrl;
+  /**
+   * Try to parse the `INNGEST_DEV` environment variable as a URL.
+   * Returns the URL if valid, otherwise `undefined`.
+   */
+  // TODO: this is really more internal, right?
+  get getExplicitDevUrl(): URL | undefined {
```
amh4r:
```suggestion
  private get getExplicitDevUrl(): URL | undefined {
```

I agree with your comment: this is an internal thing. Ideally we can confine the Dev Server URL shenanigans to just the client

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -442,12 +417,13 @@ export class InngestCommHandler<
...
+    // XXX: remove for v3->v4
     if (options.id) {
-      console.warn(
-        `${logPrefix} The \`id\` serve option is deprecated and will be removed in v4`,
+      throw new Error(
+        `${logPrefix} The \`id\` serve option is deprecated and has been removed in v4`,
       );
     }
-    this.id = options.id || this.client.id;
+    this.id = this.client.id;
```
amh4r:
```suggestion
```

Can we remove this? The client is the source of truth for the app ID. Duplicating the value creates an opportunity for state drift

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -566,13 +544,7 @@ export class InngestCommHandler<
    * each time it's accessed, as it may change during execution.
    */
   protected get apiBaseUrl(): string {
```
amh4r: This method is going away, right? Since the `baseUrl` option will no longer be in the `serve` functions

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -2448,38 +2369,33 @@ export class InngestCommHandler<
...
-    if (this.env[envKeys.InngestSigningKeyFallback]) {
-      if (!this.signingKeyFallback) {
-        this.signingKeyFallback = String(
-          this.env[envKeys.InngestSigningKeyFallback],
-        );
+  private validateModeConfiguration({
+    skipIfNoEnv = false,
+  }: {
+    skipIfNoEnv?: boolean;
+  } = {}) {
```
amh4r:
```suggestion
  private validateModeConfiguration(skipIfNoEnv: boolean = false) {
```

Only 1 option so we don't need object destructuring

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -2504,12 +2420,12 @@ export class InngestCommHandler<
       // Never validate signatures outside of prod. Make sure to check the mode
       // exists here instead of using nullish coalescing to confirm that the check
       // has been completed.
-      if (this._mode && !this._mode.isCloud) {
+      if (this.client.mode && this.client.mode !== "cloud") {
```
amh4r:
```suggestion
      if (this.client.mode !== "cloud") {
```
`this.client.mode` is always truthy

packages/inngest/src/components/execution/otel/processor.ts
```diff
@@ -15,12 +15,7 @@ import {
   type SpanProcessor,
 } from "@opentelemetry/sdk-trace-base";
 import Debug from "debug";
-import {
-  defaultDevServerHost,
-  defaultInngestApiBaseUrl,
-} from "../../../helpers/consts.ts";
-import { devServerAvailable } from "../../../helpers/devserver.ts";
-import { devServerHost } from "../../../helpers/env.ts";
+import { defaultInngestApiBaseUrl } from "../../../helpers/consts.ts";
```
amh4r:
```suggestion
```
Unused import

packages/inngest/src/helpers/env.ts
```diff
@@ -60,182 +59,87 @@ export const devServerHost = (env: Env = allProcessEnv()): EnvValue => {
...
+//    */
+//   // move to a helper in client
+//   public getUrl(defaultCloudUrl: string): string {
+//     if (this.explicitDevUrl) {
+//       return this.explicitDevUrl.href;
+//     }
+//
+//     return this.isCloud ? defaultCloudUrl : defaultDevServerHost;
+//   }
+// }
```
amh4r:
```suggestion
```

packages/inngest/src/helpers/env.ts
```diff
@@ -60,182 +59,87 @@ export const devServerHost = (env: Env = allProcessEnv()): EnvValue => {
...
+//     // Check environment variables for INNGEST_DEV=0|1|boolean
+//     const envIsDev = parseAsBoolean(env[envKeys.InngestDevMode]);
+//     if (typeof envIsDev === "boolean") {
+//       return new Mode({ type: envIsDev ? "dev" : "cloud" });
+//     }
+//   }
+//
+//   // 4. Default to cloud mode
+//   return new Mode({ type: "cloud" });
+// };
```
amh4r:
```suggestion
```

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -2448,38 +2361,29 @@ export class InngestCommHandler<
...
-      this.log(
-        "warn",
-        `Use of ${envKeys.InngestDevServerUrl} has been deprecated in v3; please use ${envKeys.InngestBaseUrl} instead. See https://www.inngest.com/docs/sdk/migration`,
+    if (this.client.mode === "cloud" && !this.client.signingKey) {
+      throw new Error(
+        `Inngest error: A signing key is required to run in Cloud mode, but no signing key was found.\n\n` +
+          `To fix this, choose one of the following:\n` +
+          `  - For local development, set INNGEST_DEV=1 to use the Dev Server (e.g. INNGEST_DEV=1 npm run dev)\n` +
+          `  - For production, set the ${envKeys.InngestSigningKey} environment variable\n\n` +
+          `Find your keys at https://app.inngest.com/secrets`,
```
amh4r:
```suggestion
          `Find your keys at https://app.inngest.com`,
```

We don't have a `/secrets` path, but that's a good idea!

packages/inngest/src/components/Inngest.ts
```diff
@@ -156,12 +139,79 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
+      this._env[envKeys.InngestBaseUrl] ||
+      this.resolveDefaultUrl(defaultInngestEventBaseUrl)
+    );
+  }
+
+  // NOTE:
+  // This being undefined makes sense here, but once it's accessed IRL
+  // it should cause issues!
+  get signingKey(): string | undefined {
+    return this._env[envKeys.InngestSigningKey];
```
amh4r: This isn't using the in-code option (currently in `serve`, but eventually on the client)

packages/inngest/src/components/Inngest.ts
```diff
@@ -156,12 +139,79 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
   }
 
-  get eventBaseUrl(): string | undefined {
-    return this._eventBaseUrl;
+  get signingKeyFallback(): string | undefined {
+    return this._env[envKeys.InngestSigningKeyFallback];
```
amh4r: This isn't using the in-code option (currently in `serve`, but eventually on the client)

packages/inngest/src/components/Inngest.ts
```diff
@@ -108,24 +110,18 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
    * The absolute URL of the Inngest Cloud API.
    */
-  private sendEventUrl: URL = new URL(
-    `e/${this.eventKey}`,
-    defaultInngestEventBaseUrl,
-  );
-
-  private headers!: Record<string, string>;
+  private get sendEventUrl(): URL {
+    return new URL(`e/${this.eventKey}`, this.eventBaseUrl);
```
amh4r: Event key is "tainted" by env vars in the same way as other fields (signing key, API URL, etc.).

Also, this method is only used in 1 spot so we can inline it:
```diff
diff --git a/packages/inngest/src/components/Inngest.ts b/packages/inngest/src/components/Inngest.ts
index b63c739c..383877d1 100644
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -736,7 +736,8 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
 
         // We don't need to do fallback auth here because this uses event keys and
         // not signing keys
-        const response = await this.fetch(this.sendEventUrl.href, {
+        const url = new URL(`e/${this.eventKey ?? dummyEventKey}`, this.eventBaseUrl)
+        const response = await this.fetch(url, {
           method: "POST",
           body: stringify(payloads),
           headers: { ...this.headers, ...headers },
```

packages/inngest/src/cloudflare.ts
```diff
@@ -133,7 +133,6 @@ export const serve = (
      * globally available in the Cloudflare env. Specifying it here will
      * ensure we avoid trying to load a Node-compatible version later.
      */
-    fetch: fetch.bind(globalThis),
```
amh4r, line 136: Should we mention this in the migration guide? Deleting it makes sense since it's on the client now, but some users may need to set `fetch: fetch.bind(globalThis)` on their client.  That said, I'm kinda unsure why this is needed 😆.

Also, we can delete the comment above

amh4r (approved):

packages/inngest/src/cloudflare.ts
Linell: It's theoretically no longer needed, but I'm definitely planning on testing the hoot out of edge environments as well just to be sure. I did add a snippet to the migration doc, though!

PR #1242, opened 2026-01-15
Fix NonRetriableError instanceof check in monorepos

### LINKED ISSUE #1241: NonRetriableError doesn't work in monorepo
Throwing `NonRetriableError` still causes retries in our monorepo setup. Looking at the code, the SDK uses `instanceof NonRetriableError` which fails when different packages load different instances of the inngest module.

```typescript
// Shared package
import { Inngest, NonRetriableError } from "inngest";
export const inngest = new Inngest({ id: "app" });

// Worker package
import { inngest } from "@shared/inngest";
import { NonRetriableError } from "inngest"; // different instance

inngest.createFunction(
  { id: "test" },
  { event: "test" },
  async () => {
    throw new NonRetriableError("no retry"); // still retries
  }
);
```

Adding `error?.name === "NonRetriableError"` as a fallback check should fix this. Happy to submit a PR if this approach works.

Related to #1181

pingren:
Fixes NonRetriableError not being recognized in monorepo setups where different packages can load separate instances of the inngest module.

The SDK currently only uses `instanceof` to detect NonRetriableError. This fails when the error comes from a different module instance - common in monorepos.

Added `(error as any)?.name === "NonRetriableError"` as a fallback check in v0/v1/v2 execution handlers.

Fixes #1241

### DIFF
--- a/packages/inngest/src/components/execution/v0.ts
+++ b/packages/inngest/src/components/execution/v0.ts
@@ -561,7 +561,10 @@ export class V0InngestExecution
        * Ensure we give middleware the chance to decide on retriable behaviour
        * by looking at the error returned from output transformation.
        */
-      let retriable: boolean | string = !(error instanceof NonRetriableError);
+      let retriable: boolean | string = !(
+        error instanceof NonRetriableError ||
+        (error as any)?.name === "NonRetriableError"
+      );
       if (retriable && error instanceof RetryAfterError) {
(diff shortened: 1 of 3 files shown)


Linell:
Fixed in [v3.51.0](https://github.com/inngest/inngest-js/releases/tag/inngest%403.51.0)

PR #1244, opened 2026-01-15
Ensure execution version doesn't slide when going to/from checkpointing

jpwilliams:
## Summary

Fixes a bug whereby the execution version used could slide from V1 to V2 when using both `checkpointing: true` and `optimizedParallelism: true` together.

## Checklist

- [ ] ~Added unit/integration tests~ POST-MERGE: test suite for this and #1245

### DIFF
--- a/packages/inngest/src/helpers/functions.ts
+++ b/packages/inngest/src/helpers/functions.ts
@@ -243,11 +243,18 @@ export const fetchAllFnData = async ({
 }): Promise<Result<FnData, ParseErr>> => {
   const result = { ...data };
 
+  // Ugly pattern, but ensures we always check every execution model correctly.
+  const shouldFetchData: Record<ExecutionVersion, () => boolean> = {
+    [ExecutionVersion.V0]: () =>
+      result.version === ExecutionVersion.V0 && result.use_api,
+    [ExecutionVersion.V1]: () =>
+      result.version === ExecutionVersion.V1 && Boolean(result.ctx?.use_api),
+    [ExecutionVersion.V2]: () =>
+      result.version === ExecutionVersion.V2 && Boolean(result.ctx?.use_api),
+  };
+
   try {
-    if (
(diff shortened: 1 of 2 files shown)


packages/inngest/src/helpers/functions.test.ts
```diff
@@ -25,6 +25,7 @@ describe("#parseFnData", () => {
       name: "should parse successfully for valid data",
       data: {
         version: 1,
+        sdkDecided: true,
```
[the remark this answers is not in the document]
jpwilliams: Correct here.

packages/inngest/src/helpers/functions.ts
```diff
@@ -84,29 +84,45 @@ const fnDataVersionSchema = z.object({
     .or(z.literal(1))
     .or(z.literal(2))
     .optional()
-    .transform<ExecutionVersion>((v) => {
+    .transform<{ version: ExecutionVersion; sdkDecided: boolean }>((v) => {
```
jpwilliams, line 87: When we parse the execution version given to us by the Executor, `-1` means "the SDK can decide."

We want to know if this is the case; if the SDK can decide then we're allowed to let options like `optimizedParallelism: true` affect the execution version used, whereas if it's already been set by an earlier execution then we **must** stay put.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1904,11 +1905,12 @@ export class InngestCommHandler<
...
-    let { version } = immediateFnData;
+    let { version, sdkDecided } = immediateFnData;
 
     // Handle opting in to optimized parallelism in v3.
     if (
       version === ExecutionVersion.V1 &&
+      sdkDecided &&
       fn.fn["shouldOptimizeParallelism"]?.()
     ) {
       version = ExecutionVersion.V2;
```
jpwilliams, line 1916: Only allow this version shift if the SDK is deciding on the version right now.

PR #1256, opened 2026-01-20
refactor: JIT event key setting

Linell:
## Summary
Instead of a dummy event key being set across the app, we *only* tack that in when we're sending in dev with an undefined key.

## Checklist

## Related

[EXE-1151: Client's `eventKey` field is never set to the dummy event key](https://linear.app/inngest/issue/EXE-1151/clients-eventkey-field-is-never-set-to-the-dummy-event-key)

### DIFF
--- a/packages/inngest/src/components/InngestCommHandler.ts
+++ b/packages/inngest/src/components/InngestCommHandler.ts
@@ -5,7 +5,6 @@ import { getAsyncCtx } from "../experimental";
 import {
   debugPrefix,
   defaultMaxRetries,
-  dummyEventKey,
   ExecutionVersion,
   envKeys,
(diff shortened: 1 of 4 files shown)


packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -535,10 +534,10 @@ export class InngestCommHandler<
   }
 
   private get hashedEventKey(): string | undefined {
-    if (!this.client["eventKey"] || this.client["eventKey"] === dummyEventKey) {
+    if (!this.client.eventKey) {
```
amh4r, line 537: Fewer private field accessors 😍

PR #1258, opened 2026-01-21
Connect: Add worker thread support

amh4r:
## Summary

Add worker thread support for Connect. Users can opt into worker thread mode by either:
- Setting `INNGEST_CONNET_USE_WORKER_THREAD` to a truthy value (`1` or `true`)
- Setting `useWorkerThread: true` in code

The existing main thread logic should remain exactly the same. But significant refactoring was necessary to share core logic between the 2 strategies.

Known issues:
- `rewriteGatewayEndpoint` does not work with worker threads. A followup PR will add flawed support and mark it as deprecated. The only way to do it is by having the worker thread ask the main thread to run the function (since function's can't be passed to worker threads), but that creates an opportunity for main thread event loop blocking to affect the worker thread (which defeats the whole purpose of the worker thread). We'll eventually replace it with a static override that can be passed to the worker thread

## Related
https://linear.app/inngest/issue/INN-5525

### DIFF
--- a/packages/inngest/src/components/connect/strategies/sameThread/index.ts
+++ b/packages/inngest/src/components/connect/strategies/sameThread/index.ts
@@ -1,4 +1,3 @@
-import debug, { type Debugger } from "debug";
 import { SDKResponse } from "../../../../proto/src/components/connect/protobuf/connect.ts";
 import { MessageBuffer } from "../../buffer.ts";
 import { BaseStrategy } from "../core/BaseStrategy.ts";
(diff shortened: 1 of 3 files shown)


BrunoScheufler (approved):

packages/inngest/src/components/connect/strategies/sameThread/index.ts
```diff
@@ -0,0 +1,127 @@
...
+export class SameThreadStrategy extends BaseStrategy {
+  private readonly config: StrategyConfig;
+  protected readonly debugLog: Debugger;
+  private readonly messageBuffer: MessageBuffer;
+  private readonly core: ConnectionCore;
+
+  constructor(config: StrategyConfig) {
+    super();
+    this.config = config;
+    this.debugLog = debug("inngest:connect");
```
Linell: How would you feel about moving this into the `BaseStrategy` and possibly namespacing it off based on the strategy?

PR #1266, opened 2026-01-23
EXE-1206: Ensure middleware transformers are called with checkpointing

djfarrelly:
## Summary

Ensure that checkpointing utilizes middleware transformations.

Failing tests: 
- https://github.com/inngest/inngest-js/actions/runs/21270191817/job/61218542645?pr=1266
- https://github.com/inngest/inngest-js/actions/runs/21270781426/job/61220304545

## Checklist

## Related
- EXE-1206

### DIFF
--- a/packages/inngest/src/components/InngestFunction.test.ts
+++ b/packages/inngest/src/components/InngestFunction.test.ts
@@ -3860,7 +3860,10 @@ describe("runFn", () => {
         const result = await execution.start();
 
         expect(result.type).toBe("function-resolved");
-        expect(transformOutputMock).toHaveBeenCalledTimes(1);
+        // In sync mode, transformOutput is called twice:
+        // 1. Once to transform data for checkpointing
+        // 2. Once to transform data for the SDK return value
+        expect(transformOutputMock).toHaveBeenCalledTimes(2);
         expect(transformOutputMock).toHaveBeenCalledWith(
           expect.objectContaining({
             result: expect.objectContaining({
@@ -4065,7 +4068,9 @@ describe("runFn", () => {
               id: "0737c22d3bfae812339732d14d8c7dbd6dc4e09c",
               op: "RunComplete",
               data: {
-                body: '{"encrypted":true,"original":{"stepData":"hello"}}',
+                // The final function result is { final: { stepData: "hello" } }
+                // which gets encrypted by transformOutput middleware
+                body: '{"encrypted":true,"original":{"final":{"stepData":"hello"}}}',
                 headers: {},
                 status: 200,
                 version: 2,
@@ -4191,7 +4196,9 @@ describe("runFn", () => {
               id: "0737c22d3bfae812339732d14d8c7dbd6dc4e09c",
               op: "RunComplete",
               data: {
-                body: '{"encrypted":true,"original":{"stepData":"hello"}}',
+                // The final function result is { final: { stepData: "hello" } }
+                // which gets encrypted by transformOutput middleware
+                body: '{"encrypted":true,"original":{"final":{"stepData":"hello"}}}',
                 headers: {},
                 status: 200,
                 version: 2,


packages/inngest/src/components/execution/v1.ts
```diff
@@ -409,9 +409,23 @@ class V1InngestExecution extends InngestExecution implements IInngestExecution {
...
+
+        // Transform data for checkpoint (middleware)
+        // Only call the transformOutput hook directly, not the full transformOutput method
+        // which has side effects like calling the finished hook
+        const transformedOutput = await this.state.hooks?.transformOutput?.({
+          result: { data: stepResult.data },
+          step: undefined,
+        });
+        const transformedData =
+          transformedOutput?.result?.data ?? stepResult.data;
```
djfarrelly, line 423: I don't love this code, but we do need to just transform the output with the hook for middleware without the other stuff and transforming the output w/ the function-resolved/rejected logic

GautierT:
```
let summary = await step.run("get-summary-1", async () => {
	logger.info(
		"[create-summary-v2] | get-summary-1 | summaryId : ",
		summaryId,
	);
	return await getFullSummaryForGeneration(summaryId, orgId);
});

logger.info(
	{ summary, prompt: summary.prompt },
	"[create-summary-v2] | summary",
);

if (!summary.prompt)
	throw new Error("Prompt not found after get-summary-1");
```

With this version I have he error `Prompt not found after get-summary-1`
In 3.49.3 it's working correctly so a bug was introduced.

GautierT:
working version here : https://github.com/inngest/inngest-js/compare/main...notae-ai:inngest-js:fix/encryption-checkpointing

(with the help of Claude)

packages/inngest/src/components/execution/v2.ts
```diff
@@ -471,27 +485,17 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
...
 
-        // Done - just return the value
-        return {
-          type: "function-resolved",
-          ctx: this.fnArg,
-          ops: this.ops,
-          data: checkpoint.data,
-        };
+        // Apply middleware transformation before returning
+        return await this.transformOutput({ data: checkpoint.data });
```
djfarrelly, line 500: @jpwilliams - I think this may need to be changed to transform before calling `this.checkpoint` to ensure the transformed value is sent to the checkpoint API.

packages/inngest/src/components/execution/v2.ts
```diff
@@ -471,27 +485,17 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
...
         if (this.inFinalAttempt()) {
-          return {
-            type: "function-rejected",
-            ctx: this.fnArg,
-            error: checkpoint.error,
-            ops: this.ops,
-            retriable: false,
-          };
+          // Apply middleware transformation before returning
+          return await this.transformOutput({ error: checkpoint.error });
```
djfarrelly, line 509: I assumed that we should be calling the transform here, but I don't know if this might cause another bug if we run middleware before returning the output to the client during sync mode? Not sure how this error gets pushed to the client if at all.

packages/inngest/src/components/InngestFunction.test.ts
```diff
@@ -3639,4 +3640,324 @@ describe("runFn", () => {
...
+        expect(result.type).toBe("function-resolved");
+        expect(transformOutputMock).toHaveBeenCalledTimes(1);
+        expect(transformOutputMock).toHaveBeenCalledWith(
+          expect.objectContaining({
+            result: expect.objectContaining({
+              data: { result: "success" },
+            }),
+          }),
+        );
+      });
```
djfarrelly, line 3879: TODO - More tests on what data gets checkpointed with the API.

packages/inngest/src/components/execution/v2.ts
```diff
@@ -407,9 +407,23 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
...
+          step: stepResult,
+        });
+        const transformedData =
+          transformedOutput?.result?.data ?? stepResult.data;
+
+        // Buffer a copy with transformed data for checkpointing
+        this.state.checkpointingStepBuffer.push({
+          ...stepToResume,
+          data: transformedData,
+        });
```
djfarrelly, line 427: This code is called from `asyncCheckpointingHandlers` via the `steps-found` handler. This ensures that if we `tryExecuteStep`s in that handler, we run the middleware transform before checkpointing to the API.

packages/inngest/src/components/execution/v2.ts
```diff
@@ -463,35 +477,35 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
       "": commonCheckpointHandler,
 
       "function-resolved": async (checkpoint, i) => {
+        // Transform data for checkpoint (middleware)
+        // Only call the transformOutput hook directly, not the full transformOutput method
+        // which has side effects like calling the finished hook
+        const transformedOutput = await this.state.hooks?.transformOutput?.({
+          result: { data: checkpoint.data },
+          step: this.state.executingStep,
```
djfarrelly, line 486: @jpwilliams  I'm unsure if this is the correct step to call to avoid a cached result - I mimic'd the `transformStep`'s internal usage.

jpwilliams (approved):

PR #1267, opened 2026-01-23
Maybe fix signature verification failures

amh4r:
## Summary

Fix a possible issue that incorrectly sets the body to an empty string during signature verification.

Previously, we had a simple falsy check on the `Content-Length` header. But this is flawed. For example, transfer chunked requests may not have the `Content-Length` header while still containing valid JSON.

## Context

The falsy check was originally added to support syncing with `curl -X PUT localhost:8000/api/inngest`. Doing that results in an empty body and no `Content-Length` header

### DIFF
--- a/packages/inngest/src/node.test.ts
+++ b/packages/inngest/src/node.test.ts
@@ -23,7 +23,8 @@ testFramework("Node", NodeHandler, {
     }
 
     // Mock the body data
-    const bodyData = Buffer.from(JSON.stringify(req.body));
+    const bodyString = req.body === undefined ? "" : JSON.stringify(req.body);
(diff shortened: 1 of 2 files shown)


Linell:
I was seeing lots of signature validation errors on Vercel myself, and after cherry-picking your code onto that branch the issues were resolved.

You can see it was pretty impactful before:
Screenshot 2026-01-24 at 5 48 49 PM
Then after:
Screenshot 2026-01-24 at 7 41 20 PM
You can see the errors drop off right after I deployed the rebased code.
Screenshot 2026-01-24 at 7 42 56 PM

PR #1268, opened 2026-01-23
fix: encrypt step data before checkpointing

GautierT:
Checkpointing bypasses middleware hooks - step data is sent directly to the API without going through `transformOutput`.

This breaks encryption middleware since data gets checkpointed in plaintext.

Fix: call `transformOutput` on step data before sending to checkpoint API. Uses shallow copies to avoid mutating local data.


GautierT:
I saw you working on it already : https://github.com/inngest/inngest-js/pull/1266 👍

PR #1275, opened 2026-01-28
feat(openai): add gpt-5.1 and gpt-5.2 model identifiers

### LINKED ISSUE #1274: feat(openai): add gpt-5.1 and gpt-5.2 model identifiers
## Is your feature request related to a problem? Please describe.

When configuring OpenAI models using the Inngest JavaScript SDK, newer incremental GPT-5 model versions such as `gpt-5.1` and `gpt-5.2` are not currently included in the `Model` type union.  
While the SDK intentionally supports arbitrary strings for forward compatibility, the absence of these officially released model identifiers limits autocomplete, discoverability, and type safety for developers using TypeScript.

---

## Describe the solution you'd like

Extend the `Model` union type to explicitly include the stable, officially released OpenAI model identifiers:

- `gpt-5.1`
- `gpt-5.2`

This change would be additive and backward compatible, improving developer experience without introducing any runtime or behavioral changes.

---

## Describe alternatives you've considered

- Relying exclusively on the `(string & {})` fallback, which maintains compatibility but does not provide explicit typing or IDE autocomplete for newly released models.
- Adding speculative or undocumented variants (for example, size-based or preview suffixes), which was intentionally avoided to keep the SDK aligned with documented and stable OpenAI model identifiers.

---

## Additional context

- The SDK already follows a clear pattern of listing stable OpenAI model identifiers while preserving forward compatibility.
- Incremental model versions (such as `gpt-4.1`) have been added previously for the same reason.
- This proposal aligns with existing naming conventions and does not introduce breaking changes or assumptions about future variants.

Anuj-K15:
## Summary

Adds support for newly released OpenAI GPT-5 incremental model versions by extending the `Model` type definition in `packages/ai/src/models/openai.ts`.

This update ensures developers can configure `gpt-5.1` and `gpt-5.2` with full TypeScript autocomplete and type safety when using Inngest AI integrations.

---

## Changes

- Added `gpt-5.1` and `gpt-5.2` to the `Model` union type
- No runtime or behavioral changes introduced
- Forward compatibility remains intact via the existing string fallback

---

## Motivation

OpenAI has released incremental GPT-5 model versions that are already usable via the API.  
Explicitly including these identifiers improves developer experience by making the latest models discoverable and type-safe in TypeScript projects, consistent with how previous model updates (e.g. `gpt-4.1`) are handled in the SDK.

---

## Testing

- Ran package-level tests for `@inngest/ai`
- All unit tests passed successfully (Vitest)
- No runtime behavior changes introduced

---

## Checklist

- [x] Change is additive and backward compatible
- [x] No documentation updates required (internal type definition only)
- [x] No runtime behavior changes
- [x] Changeset added (can be included if a release bump is required)

---

## Related

- Closes #1274
- Follows the same pattern as prior GPT-5 model additions


Anuj-K15:
Added a changeset for a patch release of `@inngest/ai.`

Anuj-K15:
### Integration test fix

I’ve addressed the failing integration test by updating the trigger used by the `cronInvokeFn` test function.

**Root cause**
The function was defined with a cron trigger (`{ cron: "59 23 31 12 *" }`), which cannot be invoked via `step.invoke()`. As a result, the invocation returned an error (`-1`) instead of the expected result object, causing the test assertion to fail.

**Fix**
- Updated the function trigger to use an event trigger instead:
  - Replaced `{ cron: "59 23 31 12 *" }` with `{ event: "demo/step.invoke.cron" }`
- This allows the function to be invoked correctly via `step.invoke()` during the integration test.

**File changed**
- `packages/inngest/src/test/functions/step-invoke/index.ts`

**Result**
The integration test now correctly receives:
- `{ eventInvokeDone: true }` from the event-triggered function
- `{ cronInvokeDone: true }` from the updated (event-triggered) cron test function

All integration tests now pass successfully.

PR #1276, opened 2026-01-29
refactor: v4 logger 🪵

Linell:
## Summary
Removes `prettyError` and refactors away from `console.log`

## Checklist

## Related

[EXE-1143: Remove "pretty errors"](https://linear.app/inngest/issue/EXE-1143/remove-pretty-errors)
[EXE-1142: Stop using `console.log`](https://linear.app/inngest/issue/EXE-1142/stop-using-consolelog)

### DIFF
--- a/packages/inngest/src/components/execution/otel/util.ts
+++ b/packages/inngest/src/components/execution/otel/util.ts
@@ -7,11 +7,8 @@ import {
 } from "@opentelemetry/instrumentation";
 import { BasicTracerProvider } from "@opentelemetry/sdk-trace-base";
 import { AnthropicInstrumentation } from "@traceloop/instrumentation-anthropic";
-import Debug from "debug";
-import { debugPrefix } from "./consts.ts";
+import { getLogger } from "../../../helpers/log.ts";
 import { InngestSpanProcessor } from "./processor.ts";
-
-const debug = Debug(`${debugPrefix}:middleware`);
 
 export type Behaviour = "createProvider" | "extendProvider" | "off" | "auto";
 export type Instrumentations = (Instrumentation | Instrumentation[])[];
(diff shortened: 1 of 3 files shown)


packages/inngest/src/components/execution/otel/util.ts
```diff
@@ -6,8 +6,12 @@ import {
   registerInstrumentations,
 } from "@opentelemetry/instrumentation";
 import { BasicTracerProvider } from "@opentelemetry/sdk-trace-base";
+import Debug from "debug";
```
amh4r: What's the reason for using this `debug` library on in otel stuff? Looks like we're using `getLogger().debug()` elsewhere

packages/inngest/src/helpers/net.ts
```diff
@@ -113,9 +114,11 @@ export async function signDataWithKey(
   if (!hasLoggedCryptoImplementation) {
     hasLoggedCryptoImplementation = true;
     if (subtle) {
-      console.debug("[inngest] Using native Web Crypto for request signing");
+      getLogger().debug(
+        "[inngest] Using native Web Crypto for request signing",
```
amh4r: I know these `[inngest]` prefixes were already here, but should we remove them?

packages/inngest/src/components/execution/otel/util.ts
Linell: The debug library adds some nice stuff that lets us filter things by namespace (ie `DEBUG=inngest:otel:*`. I've been viewing that as a little different than the logs from the logger:  essentially the debug library is for logs about "making the sausage" and the logger is for stuff we want the users to actually get.

Does that track, or do we want to make a different call? 🤔

PR #1281, opened 2026-02-03
Default to Optimized Parallelism

Linell:
## Checklist

- [x] Rebase this after the v4 logger changes are in, for ALS tweaks

### DIFF
--- a/packages/inngest/src/components/InngestStepTools.ts
+++ b/packages/inngest/src/components/InngestStepTools.ts
@@ -26,13 +26,7 @@ import {
   type StepOptionsOrId,
   type TriggerEventFromFunction,
 } from "../types.ts";
-import {
-  type AsyncContext,
-  getAsyncCtx,
-  getAsyncCtxSync,
-  getAsyncLocalStorage,
-  isALSFallback,
-} from "./execution/als.ts";
+import { getAsyncCtx, getAsyncCtxSync } from "./execution/als.ts";
 import type { InngestExecution } from "./execution/InngestExecution.ts";
 import { fetch as stepFetch } from "./Fetch.ts";
 import type {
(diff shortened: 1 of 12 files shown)


packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -1099,3 +1119,76 @@ type AiInferOpts<TModel extends AiAdapter> = {
...
+    throw new Error(
+      "`parallel()` must be called within an Inngest function execution",
+    );
+  }
+
+  const als = await getAsyncLocalStorage();
+
+  if (isALSFallback()) {
+    throw new Error(
+      "`parallel()` requires AsyncLocalStorage support, which is not available in this runtime. " +
```
Linell: According to Google, Cloudflare Workers _do_ have some basic support so we should ditch specifics here and just raise an error that it won't work.

packages/inngest/src/express.ts
```diff
@@ -26,6 +26,7 @@ import {
   InngestCommHandler,
   type ServeHandlerOptions,
 } from "./components/InngestCommHandler.ts";
+import { ExecutionVersion, headerKeys } from "./helpers/consts.ts";
```
amh4r:
```suggestion
```

packages/inngest/src/experimental.ts
```diff
@@ -1,12 +1,14 @@
 // AsyncLocalStorage
 
 export type { AsyncContext } from "./components/execution/als.ts";
-export { getAsyncCtx } from "./components/execution/als.ts";
-
+export {
+  getAsyncCtx,
+  getAsyncCtxSync,
+  getAsyncLocalStorage,
+} from "./components/execution/als.ts";
```
amh4r: Why does this stuff need to be exposed to users? Isn't it internal?

packages/inngest/src/index.ts
```diff
@@ -52,6 +52,7 @@ export { InngestCommHandler } from "./components/InngestCommHandler.ts";
 export type { InngestFunction } from "./components/InngestFunction";
 export type { InngestFunctionReference } from "./components/InngestFunctionReference";
 export { referenceFunction } from "./components/InngestFunctionReference.ts";
+export type { GroupTools } from "./components/InngestGroupTools.ts";
```
amh4r: Does the end user need the `GroupTools` type?

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -1098,3 +1119,82 @@ type AiInferOpts<TModel extends AiAdapter> = {
...
+ * // Or explicitly specify the mode
+ * const winner = await parallel({ mode: "race" }, async () => {
+ *   return Promise.race([
+ *     step.run("a", () => "a"),
+ *     step.run("b", () => "b"),
+ *   ]);
+ * });
+ * ```
+ */
+export const parallel = async <T>(
```
amh4r: Should we move this to `InngestGroupTools.ts`? Since it's a `group` method

PR #858, opened 2025-02-22
Optimize V1 execution memoization

jpwilliams:
## Summary

A pass to improve performance when using the `V1` request version.

- Using `Map` for busy hashmaps instead of regular `{}` objects
- Removal of some pure `.reduce()` calls that were recreating objects every cycle
- Optimized automatic step indexing to kill the ugly loop
  - This was done originally to handle the case where a user had manually specified `a:2`, but then also repeated `a`, ensuring we fill out previous step names as well as future ones. Still ugly though.
- Reduce the number of microtasks we extend the tick from `1000` to `10`
- After 10 passes of prolonging the event loop, let it tick to allow other code to run at a healthy rate

Below are some benchmarks when memoizing steps before and after this PR.

### Before

```
cpu: 12th Gen Intel(R) Core(TM) i5-12600K
runtime: node v22.12.0 (x64-linux)

benchmark                 time (avg)             (min … max)       p75       p99      p999
------------------------------------------------------------ -----------------------------
• execution
------------------------------------------------------------ -----------------------------
memoized-steps/0      30'010 ns/iter  (17'526 ns … 4'664 µs) 23'602 ns 95'624 ns  1'996 µs
memoized-steps/1         276 µs/iter     (219 µs … 1'210 µs)    255 µs    874 µs  1'089 µs
memoized-steps/10      1'339 µs/iter   (1'075 µs … 2'305 µs)  1'527 µs  2'055 µs  2'305 µs
memoized-steps/100    12'121 µs/iter (11'260 µs … 13'742 µs) 12'391 µs 13'742 µs 13'742 µs
memoized-steps/500    69'044 µs/iter (68'171 µs … 69'947 µs) 69'848 µs 69'947 µs 69'947 µs
memoized-steps/1000      194 ms/iter       (190 ms … 199 ms)    199 ms    199 ms    199 ms
memoized-steps/2000    1'102 ms/iter   (1'097 ms … 1'108 ms)  1'108 ms  1'108 ms  1'108 ms
memoized-steps/5000    7'551 ms/iter   (7'502 ms … 7'600 ms)  7'600 ms  7'600 ms  7'600 ms
memoized-steps/10000  31'382 ms/iter (31'286 ms … 31'479 ms) 31'479 ms 31'479 ms 31'479 ms
```

### After
```
cpu: 12th Gen Intel(R) Core(TM) i5-12600K
runtime: node v22.12.0 (x64-linux)

benchmark                 time (avg)             (min … max)       p75       p99      p999
[...]

PR #1295, opened 2026-02-05
fix: handle memoized steps in correct order

Linell:
## Summary

Fix a bug where when `optimizedParallelism` was set to true, an incorrect result could be returned from `Promise.race` that I found when testing #1281 

## Checklist

## Related
#1281

### DIFF
--- a/packages/inngest/src/components/execution/v2.ts
+++ b/packages/inngest/src/components/execution/v2.ts
@@ -1275,6 +1275,19 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
         .then(() => {
           foundStepsReportPromise = undefined;
 
+          // Handle memoized steps in completion order first to preserve
+          // Promise.race() semantics - the step that completed first during
+          // actual execution should resolve first during replay.
+          for (const hashedId of this.state.stepCompletionOrder) {
+            const step = unhandledFoundStepsToReport.get(hashedId);
+            if (step?.hasStepState && step.handle()) {
+              unhandledFoundStepsToReport.delete(hashedId);
+              if (step.fulfilled) {
+                foundStepsToReport.delete(step.hashedId);
+              }
+            }
+          }
+
           for (const [hashedId, step] of unhandledFoundStepsToReport) {
             // Note that we only run `step.handle()` if `step.hasStepState` is
             // `true`. This gives us a chance to handle checkpointing in another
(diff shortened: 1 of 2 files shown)


packages/inngest/src/components/execution/v2.ts
```diff
@@ -1337,6 +1275,19 @@ class V2InngestExecution extends InngestExecution implements IInngestExecution {
...
+          // actual execution should resolve first during replay.
+          for (const hashedId of this.state.stepCompletionOrder) {
+            const step = unhandledFoundStepsToReport.get(hashedId);
+            if (step?.hasStepState && step.handle()) {
+              unhandledFoundStepsToReport.delete(hashedId);
+              if (step.fulfilled) {
+                foundStepsToReport.delete(step.hashedId);
+              }
+            }
+          }
```
amh4r, line 1289: Does this look intentionally different from the existing implementation in `v1.ts`? I'm not sure if this is better/same/worse -- just wondering why it's different!

Linell: Well v2 handles all of them in one pass, but v1 yields between each of them. AI version:

> The v2 version handles all memoized steps in one synchronous pass, while v1 handles one step at a time,
  yielding between each (return void reportNextTick()). That yield is important: it lets a resolved step's .then() chain discover downstream steps before the next memoized step is handled. Without it, chained races like Promise.race([A, B.then(() => B2)]) break because A and B both resolve in the same pass before B2 is ever discovered, so Promise.race picks A even if B's chain should have won.

While asking Claude he pointed out that my changes break this sort of code:
```ts
  {
    id: "fn",
    retries: 0,
    optimizeParallelism: true,
  },
  [{ event: "event-0" }],
  async ({ step }) => {
    const winner = await Promise.race([
      step.run("slow", async () => {
        await sleep(4000);
        return "slow";
      }),
      (async () => {
        await step.run("fast-1", async () => {
          await sleep(100);
          return "fast-1";
        });
        return await step.run("fast-2", async () => {
          await sleep(100);
          return "fast-2";
        });
      })(),
    ]);
    console.log("winner", winner);
  },
);
```

We still get `slow` here 🤔

jpwilliams: For posterity, we're seeing that we should continue with the v1 engine (following #858) but still opt in to the Executor's understanding of the v2 engine.

PR #1302, opened 2026-02-10
refactor: use V1 execution engine everywhere

Linell:
## Summary
V1's simpler Promise resolution model and the improvements Jack has made mean that the V1 engine is the better choice in general.

V2 requests now route to the V1 engine internally.

Extracted from the changes [here](https://github.com/inngest/inngest-js/pull/1281).

## Checklist

## Related

https://github.com/inngest/inngest-js/pull/1281

### DIFF
--- a/packages/inngest/src/components/InngestCommHandler.test.ts
+++ b/packages/inngest/src/components/InngestCommHandler.test.ts
@@ -355,4 +355,74 @@ describe("response version header", () => {
       ExecutionVersion.V2.toString(),
     );
   });
+
+  test("responds with V1 when function opts out of optimized parallelism", async () => {
+    const client = createClient({ id: "test", isDev: true });
+
+    const optedOutFn = client.createFunction(
+      {
+        id: "test",
+        triggers: [{ event: "demo/event.sent" }],
+        optimizeParallelism: false,
+      },
+      () => "test",
+    );
+
+    const optedOutHandler = serve({ client, functions: [optedOutFn] });
+
+    const result = await runHandler(optedOutHandler);
+
+    expect(result.status).toBe(200);
+    expect(result.headers[headerKeys.RequestVersion]).toBe(
+      ExecutionVersion.V1.toString(),
+    );
+  });
+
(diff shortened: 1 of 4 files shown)


packages/inngest/src/components/InngestCommHandler.test.ts
```diff
@@ -126,3 +133,74 @@ describe("ServeHandler", () => {
...
+
+  const fn = inngest.createFunction(
+    { id: "test" },
+    { event: "demo/event.sent" },
+    () => "test",
+  );
+
+  const handler = serve({ client: inngest, functions: [fn] });
+
+  test("echoes V1 request version in response header", async () => {
```
amh4r: Do we need this? If we're gonna make `v1.ts` the only execution file, then it seems OK to always sending v2 back even when the request said v1. I'd double-check with Jack on that tho

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1846,43 +1841,18 @@ export class InngestCommHandler<
...
-
-    // Handle opting in to optimized parallelism in v3.
-    if (
-      version === ExecutionVersion.V1 &&
-      sdkDecided &&
-      fn.fn["shouldOptimizeParallelism"]?.()
-    ) {
-      version = ExecutionVersion.V2;
-    }
+    const version = ExecutionVersion.V2;
```
jpwilliams: I believe `optimizeParallelism` as a client-facing option is still available, just defaults to `true` now instead of `false` (opt-out instead of opt-in).

If that's the case, we need a way for the SDK to report that it wants `ExecutionVersion.V1` from the Executor, as that's what the Executor uses to decide whether to send the eager discovery steps of V1, or the coalesced discovery steps of V2.

So here we should probably keep the `if` statement but switch it around:
```ts
const version = ExecutionVersion.V2;

// Handle opting out of optimized parallelism
if (
  version === ExecutionVersion.V2 &&
  sdkDecided &&
  fn.fn["shouldOptimizeParallelism"]?.() === false
) {
  version = ExecutionVersion.V1;
}
```

PR #1305, opened 2026-02-11
Add new middleware

amh4r:
## Summary
Replace old middleware with new middleware

### DIFF
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -323,6 +323,9 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
    * has been initialized.
    */
   public get ready(): Promise<void> {
+    // Previously this was used to ensure that we could wait for middleware
+    // to be instantiated, but we now no longer have a set-up function
+    // that we await, so middleware is always ready to go.
     return Promise.resolve();
   }
 


packages/inngest/src/components/execution/engine.ts
```diff
@@ -1303,6 +1282,47 @@ class InngestExecutionEngine
...
+   * Apply mutations from `transformFunctionInput` back to execution state.
+   * Allows middleware to modify event data, step tools, memoized step data,
+   * and inject custom fields into the handler context.
+   */
+  private applyFunctionInputMutations(
+    result: Middleware.TransformFunctionInputArgs,
+  ): void {
+    const { event, events, step, ...extensions } = result.ctx;
+
+    // Mutate in place so the ALS store's reference to this.fnArg stays valid.
```
Linell, line 1250: 👏

packages/inngest/src/test/integration/utils.ts
```diff
@@ -137,26 +137,30 @@ export abstract class BaseSerializerMiddleware<
...
-    }
-    return this._serialize(result);
+  override async wrapStepHandler({ next }: Middleware.WrapStepHandlerArgs) {
+    const output = await next();
+    return this._serialize(output);
+  }
+
+  override async wrapStep({ next }: Middleware.WrapStepArgs) {
+    return this._deserialize(await next());
   }
```
jpwilliams, line 147: This actually turned out looking pretty clean

task.md
```diff
@@ -0,0 +1,9 @@
+Fix an issue with `wrapStep` when checkpointing is enabled. The `wrapStep` hook is called twice when a `step.run` is executed:
```
jpwilliams: I guess this is an accidental add

packages/inngest/src/components/middleware/middleware.ts
```diff
@@ -394,59 +441,100 @@ export namespace Middleware {
...
+     * returned output or thrown error before it's sent to the Inngest Server.
+     *
+     * Use cases:
+     * - Serialize step output before sending it to the Inngest Server.
+     * - Handle step attempt errors (before exhausting retries).
+     *
+     * Must call `next()` to continue processing. Do not mutate arguments.
+     */
+    // @privateRemark
+    // This hook exists because of checkpointing. For serialization middleware
```
jpwilliams, line 536: Thank you for these comments 🙏

packages/inngest/src/components/Inngest.ts
```diff
@@ -825,42 +765,39 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
...
       : payload
         ? ([payload] as [EventPayload])
         : [];
 
+    // Instantiate fresh middleware per send() call.
+    // Include client middleware plus all function-level middleware so that
+    // transformSendEvent fires for every registered middleware class.
+    const fnMiddleware = this.localFns.flatMap(
+      (fn) => fn.opts.middleware ?? [],
+    );
```
jpwilliams: Not a blocker.

Does this create middleware across _all_ local functions?

If I had 2 functions I had created with `inngest.createFunction()` and both had middleware, would `inngest.send()` run middleware for both?

We may need `.send()` be to aware of which function (if any) is running it, probably via ALS, so we can load the appropriate set of middleware.

packages/inngest/src/components/Inngest.ts
```diff
@@ -318,7 +323,7 @@ export class Inngest<TClientOpts extends ClientOptions = ClientOptions>
    * has been initialized.
    */
   public get ready(): Promise<void> {
-    return this.middleware.then(() => {});
+    return Promise.resolve();
```
jpwilliams, line 329:
```suggestion
    // Previously this was used to ensure that we could wait for middleware
    // to be instantiated, but we now no longer have a set-up function
    // that we await, so middleware is always ready to go.
    return Promise.resolve();
```

packages/inngest/src/components/Inngest.ts
amh4r: Fixed

PR #1313, opened 2026-02-15
Show the names and IDs of steps in steps not found errors

tonyhb:
to improve debugging, we want to show the steps that were found and not replayed when running parallel steps with a plan


packages/inngest/src/components/execution/v1.ts
```diff
@@ -1683,10 +1687,31 @@ class V1InngestExecution extends InngestExecution implements IInngestExecution {
...
+      .filter((step) => !step.hasStepState)
+      .map<StepNotFoundFoundStep>((step) => ({
+        id: step.hashedId,
+        name: step.name,
+        displayName: step.displayName,
+      }))
+      .sort((a, b) => a.id.localeCompare(b.id));
+
+    return {
+      foundSteps: foundSteps.slice(0, STEP_NOT_FOUND_MAX_FOUND_STEPS),
```
amh4r, line 1712: It feels weird to only show the first 25 found steps alphabetically. Will that add clarity or create its own confusion?

Also, is this all found steps? Or parallel found steps (irrespective of steps before parallel)?

tonyhb: Only steps that haven't ran and were found in the current parallel execution thread.  We need to limit this to some amount.  Specifically:
>  we want to show the steps that were found and not replayed

tonyhb: >  It feels weird to only show the first 25 found steps alphabetically. Will that add clarity or create its own confusion?

On this, anything is better than none.  In the current user's case, they have 2-3 parallel steps.  The human names are technically unbounded so lower > greater right now.  Being an int, we can always increase this.

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1758,16 +1736,37 @@ export class InngestCommHandler<
...
+            // we want to show the names and IDs of any steps that were found during the
+            // run process
+            const missingStepId = result.step.displayName || result.step.id;
+
+            let error = `Could not find step "${missingStepId}" to run; timed out.`;
+
+            if (result.foundSteps.length > 0) {
+              const foundStepsSummary = result.foundSteps
+                .map((step) => {
+                  const name = step.displayName || step.name || step.id;
```
jpwilliams: Don't use `step.name`; it's deprecated.
```suggestion
                  const name = step.displayName || step.id;
```

packages/inngest/src/components/execution/InngestExecution.ts
```diff
@@ -53,6 +57,12 @@ export type ExecutionResult = {
   >;
 }[keyof ExecutionResults];
 
+export interface StepNotFoundFoundStep {
+  id: string;
+  name?: string;
```
jpwilliams: Just kill `name` here - we don't want it!

```suggestion
```

PR #1316, opened 2026-02-17
Add new OpenAI models gpt-5.2 and gpt-5.1

Amit00008:
## Summary

This PR adds support for the new OpenAI models gpt-5.2 and gpt-5.1.

These models are now included in the available model list so they can be selected and used. This ensures compatibility with the latest OpenAI releases and allows users to take advantage of improved performance and capabilities.


Amit00008:
@jonmagic @joelhooks @cdzombak

PR #1321, opened 2026-02-19
feat: add staticSchema for type-only event schemas

Linell:
## Summary
Add `staticSchema()` for type-only event schemas without runtime validation

Screenshot 2026-02-19 at 9 43 04 AM

## Checklist

### DIFF
--- a/packages/inngest/src/components/triggers/trigger.test.ts
+++ b/packages/inngest/src/components/triggers/trigger.test.ts
@@ -4,7 +4,7 @@ import type { StandardSchemaV1 } from "@standard-schema/spec";
 import { describe, expect, test } from "vitest";
 import { z } from "zod";
 import { Inngest } from "../Inngest.ts";
-import { cron, eventType, invoke } from "./triggers.ts";
+import { cron, eventType, invoke, staticSchema } from "./triggers.ts";
 
 describe("cron", () => {
   test("return", () => {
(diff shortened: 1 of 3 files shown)


packages/inngest/src/components/triggers/triggers.ts
```diff
@@ -253,6 +253,18 @@ export function eventType<
   });
 }
 
+export function staticSchema<
+  TSchema extends Record<string, unknown>,
+>(): StandardSchemaV1<TSchema> {
+  return {
```
amh4r, line 265: Can you add a comment that explains why we're return this hardcoded standard schema? Just to make it easier to grok in the future

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -649,3 +649,45 @@ describe("mixed triggers", () => {
...
+  const et = eventType("user/created", {
+    schema: staticSchema<Payload>(),
+  });
+
+  test("returns a valid StandardSchemaV1", () => {
+    const schema = staticSchema<Payload>();
+    expect(schema["~standard"].version).toBe(1);
+    expect(schema["~standard"].vendor).toBe("inngest");
+    expectTypeOf(schema).toExtend<StandardSchemaV1<Payload>>();
+  });
```
amh4r: Unnecessary since TypeScript already catches this (via the `staticSchema` return type)

packages/inngest/src/components/triggers/trigger.test.ts
```diff
@@ -649,3 +649,45 @@ describe("mixed triggers", () => {
...
+
+  test("validate is a no-op passthrough", async () => {
+    const schema = staticSchema<Payload>();
+    const input = { userId: "123" };
+    const result = schema["~standard"].validate(input);
+    // Not a promise for staticSchema
+    expect(result).toEqual({ value: input });
+    expect((result as { issues?: unknown }).issues).toBeUndefined();
```
amh4r: This is feeling kinda implementation detaily. Could be this user-facing code instead:

```suggestion
    // Invalid data passes validation, since the schema is just for static types

    // @ts-expect-error - Intentionally invalid data
    const event = et.create({ userId: 123 });

    // Won't throw
    await event.validate();
```

Issue #1296, opened 2026-02-06
[BUG] `extendedTracesMiddleware`: behaviour "auto" registers duplicate instrumentations when existing OTel provider is present


ptts:
## Describe the bug

When using `extendedTracesMiddleware` with the default `behaviour: "auto"` alongside an existing OpenTelemetry SDK setup (e.g. `NodeSDK`), the middleware fails to extend the existing provider and falls back to `createProvider()`. This fallback calls `getNodeAutoInstrumentations()`, which registers **~40+ duplicate instrumentations** (including `HttpInstrumentation`, `UndiciInstrumentation`, etc.) on top of the user's own instrumentations.

This causes **duplicate spans for every HTTP request** — both incoming and outgoing.

### Root cause

[`extendProvider()`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/execution/otel/util.ts#L49-L82) checks for `addSpanProcessor` on the result of `trace.getTracerProvider()`:

```typescript
const existingProvider = trace.getTracerProvider();
// ...
if (
  !("addSpanProcessor" in existingProvider) ||
  typeof existingProvider.addSpanProcessor !== "function"
) {
  return { success: false };
}
```

However, `trace.getTracerProvider()` in `@opentelemetry/api` always returns a [`ProxyTracerProvider`](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/ProxyTracerProvider.ts) wrapper — **not** the actual `NodeTracerProvider`/`BasicTracerProvider`. `ProxyTracerProvider` only exposes `getTracer()`, `getDelegate()`, and `setDelegate()` — it does **not** have `addSpanProcessor`.

So `extendProvider()` always fails when an existing OTel SDK is initialized, regardless of the actual provider type.

Then in `"auto"` mode, [`middleware.ts`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/execution/otel/middleware.ts#L86-L97) falls through to `createProvider()`:

```typescript
case "auto": {
  const extended = extendProvider(behaviour);
  if (extended.success) { break; }         // ← always fails
  const created = createProvider(behaviour, instrumentations);  // ← always runs
  // ...
}
```

And [`createProvider()`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/execution/otel/util.ts#L15-L41) unconditionally registers all auto-instrumentations:

```typescript
registerInstrumentations({
  instrumentations: [
[...]

PR #1325, opened 2026-02-20
fix(otel): support OTel SDK v2 in extendProvider()

### LINKED ISSUE #1324: [BUG] `extendProvider()` fails with OTel SDK v2 — `addSpanProcessor` was removed
# [BUG] `extendProvider()` fails with OTel SDK v2 — `addSpanProcessor` was removed

## Describe the bug

`extendedTracesMiddleware` with `behaviour: "extendProvider"` (or `"auto"`) fails when the host app uses **OpenTelemetry JS SDK v2** (`@opentelemetry/sdk-trace-base >= 2.0.0`).

### What happens

[`extendProvider()`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/execution/otel/util.ts#L46-L90) duck-types for `addSpanProcessor` on the resolved provider:

```typescript
// packages/inngest/src/components/execution/otel/util.ts — extendProvider()
if (
  !existingProvider ||
  !("addSpanProcessor" in existingProvider) ||    // ← false in v2
  typeof existingProvider.addSpanProcessor !== "function"
) {
  // prints warning, returns { success: false }
}
```

[`BasicTracerProvider.addSpanProcessor()`](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts) was **removed in OTel SDK v2.0.0** (released 2025-03-17). The check always fails → warning → no Inngest traces.

### What this is NOT

This is a **separate issue** from #1296 / PR #1297. That fix correctly unwrapped `ProxyTracerProvider` via `getDelegate()` ✅. But even after unwrapping, the delegate `NodeTracerProvider` (extends `BasicTracerProvider`) no longer has `addSpanProcessor` in v2.

### Partial v2 migration already done

Inngest's own `@opentelemetry/sdk-trace-base` dependency range is `>=2.0.0 <3.0.0`, and [`createProvider()`](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/execution/otel/util.ts#L15-L40) already uses the v2 constructor pattern:

```typescript
const p = new BasicTracerProvider({
  spanProcessors: [processor],  // ← v2 pattern ✅
});
```

`extendProvider()` was missed in the migration.

## Impact

| `behaviour` | Result |
|---|---|
| `"extendProvider"` | Always fails, logs warning, **no Inngest traces** |
| `"auto"` | Fails extend → falls through to `createProvider()` → registers **~40 duplicate instrumentations** |

No Inngest-specific trace data (function names, step spans) reaches the user's trace backend. Only HTTP-level spans from the user's own instrumentations are captured.

## OTel SDK v2 context

`BasicTracerProvider.addSpanProcessor()` was deprecated in v1 ([PR #5134](https://github.com/open-telemetry/opentelemetry-js/pull/5134)) and removed in v2.0.0. Span processors are now constructor-only:

```typescript
// v1 (removed in v2):
const provider = new BasicTracerProvider();
provider.addSpanProcessor(new MyProcessor());

// v2:
const provider = new BasicTracerProvider({
  spanProcessors: [new MyProcessor()],
});
```

There is **no official v2 API** to add a span processor after construction ([opentelemetry-js#5299](https://github.com/open-telemetry/opentelemetry-js/issues/5299) — closed as "not planned").

However, the internal structure is still accessible at runtime (TypeScript `private`, not ES `#private`):

- [`BasicTracerProvider._activeSpanProcessor`](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L44) — a `MultiSpanProcessor` instance
- [`MultiSpanProcessor._spanProcessors`](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/MultiSpanProcessor.ts#L29) — the underlying `SpanProcessor[]` array

## To reproduce

1. Set up an OTel `NodeSDK` with any instrumentation (`@opentelemetry/sdk-trace-base >= 2.0.0`)
2. Create an Inngest client with `extendedTracesMiddleware({ behaviour: "extendProvider" })`
3. Trigger an Inngest function
4. Observe: warning logged, no Inngest spans in trace backend

```typescript
// instrumentation.ts
import { NodeSDK } from "@opentelemetry/sdk-node"; // sdk-node 0.200+
const sdk = new NodeSDK({ /* ... */ });
sdk.start();

// inngest-client.ts
import { extendedTracesMiddleware } from "inngest/experimental";

export const inngest = new Inngest({
  id: "my-app",
  middleware: [
    extendedTracesMiddleware({ behaviour: "extendProvider" }),
    // ⚠️ Logs: "Existing OTel provider is not a BasicTracerProvider..."
    // No Inngest spans exported
  ],
});
```

## Suggested fix (immediate)

Try `addSpanProcessor()` first (v1 compat), then fall back to pushing into the internal `_spanProcessors` array (v2). The helper is wrapped in try/catch so it can never crash the host app — worst case it returns `undefined` and the middleware logs a warning. I can open a PR with the implementation and tests if this approach works for you.

### Why this approach

1. **v1 still works** — tries `addSpanProcessor()` first, no breakage for users on OTel SDK v1
2. **v2 works** — falls back to internal array push, the only viable approach
3. **Safe** — helper validates every field before accessing it and is wrapped in try/catch; can never crash
4. **No new dependencies** — purely runtime duck-typing, same pattern already used in the function

## Longer-term: avoid reaching into provider internals

The `extendProvider` path has now broken twice — first with `ProxyTracerProvider` wrapping (#1296), now with the v2 `addSpanProcessor` removal. Both stem from the same root problem: **the middleware tries to mutate a provider it doesn't own**, which is inherently fragile and couples Inngest to OTel implementation details.

OTel v2's design makes this explicit — providers are immutable after construction. Rather than working around that with increasingly creative hacks, consider giving users a way to wire the `InngestSpanProcessor` into their own provider at construction time. For example:

```typescript
// Option A: export the processor, let users add it to their SDK
import { InngestSpanProcessor } from "inngest/otel";

const sdk = new NodeSDK({
  spanProcessors: [new InngestSpanProcessor()],
});
```

```typescript
// Option B: factory that returns both middleware and processor
const { middleware, processor } = createInngestTracing();

const sdk = new NodeSDK({ spanProcessors: [processor] });
const inngest = new Inngest({ middleware: [middleware] });
```

Either approach would:
- Work with any OTel version without accessing internals
- Make the integration explicit and debuggable
- Let the `createProvider` path (for users without their own OTel setup) stay as-is

The internal-array-push fix is pragmatic for now, but exposing the processor would make `extendProvider` unnecessary long-term.

## System info

- **inngest:** 3.52.0 (also verified on 3.52.2 — not fixed)
- **@opentelemetry/sdk-trace-base:** 2.2.0 (host app), 2.5.0 (bundled in inngest)
- **@opentelemetry/sdk-node:** 0.212.0
- **@opentelemetry/api:** 1.9.0

## Related

- #1296 — original bug report for `ProxyTracerProvider` wrapping
- PR #1297 — added `getDelegate()` unwrapping (merged 2026-02-09)
- [OTel JS v2 migration guide](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md) — "removed `BasicTracerProvider#addSpanProcessor(...)` -> use constructor options"
- [OTel Issue #5299](https://github.com/open-telemetry/opentelemetry-js/issues/5299) — no public API to add processors post-construction in v2 (closed "not planned")

ptts:
## Summary

`extendProvider()` fails when the host app uses OTel SDK v2 because `BasicTracerProvider.addSpanProcessor()` was removed in v2.0.0. The duck-type check always fails → warning → no Inngest traces.

This PR:
- Tries `addSpanProcessor()` first (v1 compat)
- Falls back to pushing into the internal `_activeSpanProcessor._spanProcessors` array (v2)
- Helper is wrapped in try/catch — can never crash the host app
- Adds tests for the v2 path, unknown providers, and warning behavior

## Checklist

- _N/A — internal bug fix, no docs change needed_

## Related

- Fixes #1324
- Follow-up to #1296 / PR #1297 (ProxyTracerProvider unwrapping)

### DIFF
--- a/.changeset/fix-otel-v2-extend-provider.md
+++ b/.changeset/fix-otel-v2-extend-provider.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix `extendProvider()` for OTel SDK v2 where `addSpanProcessor()` was removed
--- a/packages/inngest/src/components/execution/otel/util.test.ts
+++ b/packages/inngest/src/components/execution/otel/util.test.ts
@@ -1,4 +1,4 @@
-import { trace } from "@opentelemetry/api";
+import { type TracerProvider, trace } from "@opentelemetry/api";
 import { BasicTracerProvider } from "@opentelemetry/sdk-trace-base";
 import { InngestSpanProcessor } from "./processor.ts";
 import { extendProvider } from "./util.ts";
@@ -60,7 +60,7 @@ describe("extendProvider", () => {
     warnSpy.mockRestore();
   });
 
-  test("should call addSpanProcessor on the underlying provider", () => {
+  test("should call addSpanProcessor on the underlying provider (v1 path)", () => {
     const { provider, addSpanProcessor } = createProviderWithAddSpanProcessor();
     trace.setGlobalTracerProvider(provider);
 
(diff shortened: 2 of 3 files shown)


Linell:
Closing as this was moving into https://github.com/inngest/inngest-js/pull/1366

PR #1333, opened 2026-02-23
Add internalLogger option

amh4r:
## Summary
Add an `internalLogger` option to the client. This logger is used for internal (i.e. "not in Inngest function") logging.

A separate field was chosen because it gives the user more control. They may want to use a child logger or something else.

## Motivation
We want to give the user the ability to prevent cluttering their app logs with our internal logs. Our internal logs are often valuable, but serve a different purpose than app logs. Also, users may want different levels for app logs and internal logs

### DIFF
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -6,7 +6,6 @@ import {
   dummyEventKey,
   envKeys,
   headerKeys,
-  logPrefix,
 } from "../helpers/consts.ts";
 import { createEntropy } from "../helpers/crypto.ts";
 import {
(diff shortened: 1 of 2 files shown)


packages/inngest/src/components/Inngest.ts
```diff
@@ -113,6 +120,14 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
 
   private readonly _logger: Logger;
 
+  /**
+   * Logger for SDK internal messages. Falls back to the user's `logger` if
+   * `internalLogger` is not provided in client options.
+   *
+   * @internal
+   */
+  readonly [internalLoggerSymbol]: Logger;
```
jpwilliams, line 128:
```suggestion
  private readonly [internalLoggerSymbol]: Logger;
```
Could we make it `private` too? Should mean that `Symbol` doesn't appear in the autocomplete when a user writes `inngest.` (it will if they write `inngest[`, but that's used for accessing private/protected methods).

PR #1334, opened 2026-02-23
Migrate encryption middleware to v4

amh4r:
Migrate `packages/middleware-encryption` to TS SDK v4 middleware

### DIFF
--- a/packages/middleware-encryption/package.json
+++ b/packages/middleware-encryption/package.json
@@ -53,7 +53,7 @@
     "@types/crypto-js": "^4.2.1",
     "@types/libsodium-wrappers": "^0.7.14",
     "cross-fetch": "^4.0.0",
-    "inngest": ">=4.0.0-alpha.5 <5.0.0",
+    "inngest": ">=4.0.0-beta.2 <5.0.0",
     "typescript": "~5.5.2",
     "vitest": "^3.0.0"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -329,8 +329,8 @@ importers:
         specifier: ^4.0.0
         version: 4.0.0
       inngest:
-        specifier: '>=4.0.0-alpha.5 <5.0.0'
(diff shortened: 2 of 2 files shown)


.changeset/wicked-vans-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inngest/middleware-encryption": major
+---
+
+Support TypeScript SDK v4
```
jpwilliams:
```suggestion
- Support TypeScript SDK v4
- Drop support for TypeScript SDK <4
```

packages/middleware-encryption/src/manual.ts
```diff
@@ -1,36 +0,0 @@
...
-  getEncryptionStages,
-} from "./stages";
-
-/**
- * Encrypts and decrypts data sent to and from Inngest.
- *
- * Returns two separate middlewares: one for encrypting data, and one for
- * decrypting data, used in special circumstances pre-v4.
- */
-export const manualEncryptionMiddleware = (
```
jpwilliams, line 13: For posterity, this was used for v3's lack of onion-layered middleware; we no longer need it.

packages/middleware-encryption/src/middleware.ts
```diff
@@ -69,24 +69,281 @@ export const encryptionMiddleware = (
...
+    override async wrapStepHandler({ next }: Middleware.WrapStepHandlerArgs) {
+      const output = await next();
+
+      if (opts.decryptOnly) {
+        return output;
+      }
+
+      if (output != null) {
+        return encryptValue(output);
+      }
```
jpwilliams: Should this be `if (output !== undefined)`?

`null` is an empty return value, but should that still be encrypted? It's sent back as `"null"`.

amh4r: Good catch noticing that differences! I felt that that encrypting/decrypting `null` was unnecessary overhead, since it inherently does not contain sensitive info.

I think this is a backwards compatible change since the old code treated `null` as "not encrypted":
https://github.com/inngest/inngest-js/pull/1334/changes#diff-deb7202ed513b7396425f5acdb7654cc19d42cb54ed2e4064f1715dbf6d09823L242

amh4r: The loose inequality check (`!=`) is confusing tho, since the vast majority of JS engineers use strict inequality checks (`!==`), so I'll change that to remove the footgun

jpwilliams: Aight cool, we'll lave as-is!

PR #1336, opened 2026-02-24
Migrate Sentry middleware to v4

amh4r:
- Migrate `packages/middleware-sentry` to TS SDK v4 middleware.
- Add integration tests using real Dev Server.
- Fix rogue vitest process bug (local dev only problem)


.changeset/dull-dryers-allow.md
Linell, line 1: Is one of these changesets a dupe?

PR #1337, opened 2026-02-24
Replace functionInfo with fn (v4 middleware)

amh4r:
Replace `functionInfo` with `fn`. We need more function info within middleware hooks


packages/inngest/src/components/middleware/middleware.ts
```diff
@@ -569,7 +566,7 @@ export namespace Middleware {
 
 type DeepReadonly<T> = T extends (infer U)[]
   ? readonly DeepReadonly<U>[]
-  : T extends Function
+  : T extends Date | Function
```
Linell: What's up with the addition of `Date` here?

PR #1342, opened 2026-02-26
EXE-1377: v4 + realtime

jacobheric:
## Summary

Introduces the new **first-class realtime implementation** in `packages/inngest` with a simpler DX, while keeping `@inngest/realtime` deprecated during the transition.

- **Spec:** https://www.notion.so/inngest/Realtime-SDK-DX-Improvements-305b64753bbd80a5a0bfff2000f4a9d7
- **Docs PR:** https://github.com/inngest/website/pull/1451
- **Examples using the new implementation:** https://github.com/inngest/inngest-js/pull/1367

---

### 1. Move non-durable publish to the client API

- **Non-durable publish** is now client-level: `inngest.publish(ref, data)`
- **Durable publish** remains step-level: `step.realtime.publish(id, ref, data)`

```ts
// Non-durable
await inngest.publish(chat.status, { message: "Streaming..." });

// Durable
await step.realtime.publish("final-status", chat.status, { message: "Done" });
```

### 2. Runtime validation on both sides

- **Publish-time validation**
  - Runs when a topic schema exists.
  - Applies to both `inngest.publish` and `step.realtime.publish`.

- **Subscribe-time validation**
  - Runs client-side when a schema exists.
  - `validate` option is supported (default: `true`).
```ts
const stream = await inngest.realtime.subscribe({
  channel: chat,
  topics: ["status"],
  validate: true,
});
```

**Validation example:**  
https://github.com/inngest/inngest-js/pull/1367/changes#diff-8dbca5b6232bb4bc72c74fe8309bfd4cf5ecfc3b2f04a76f7bad473c151231aa

### 3. Typing model on client subscriptions

- **Fully typed payload inference** requires subscribing with the shared channel object/instance.
- **String-only channel/topic subscriptions** are still supported, but payload typing will be generic/untyped.
```ts
const topics = ["status", "tokens"] as const;
const { messages } = useRealtime({
  channel: contentPipeline({ runId }),
  topics,
  token: tokenFactory,
});

// Per-topic typed access
messages.byTopic.status?.data.message; // string

// Union narrowing by topic
for (const msg of messages.delta) {
  if (msg.kind === "run") continue;
  if (msg.topic === "tokens") {
    msg.data.token; // string
  }
}
```

### 4. `useRealtime` naming clarity

Went a bit off script here. Trying to make auto-complete nice while preserving some convenience stuff for the ui. I can totally flatten this back out and remove the last/delta stuff if you think that's distracting. 

Replaced ambiguous fields with a **message-first shape**:
```ts
const { messages } = useRealtime(...);

messages.byTopic.status?.data; // core map by topic
messages.all; // core bounded
messages.last; // convenience (single most resent message across all topics)
messages.delta; // convenience (since last flush/render window)
```

**Examples with additional debug/context:**  
https://github.com/inngest/inngest-js/pull/1367/changes#diff-da7df91f25f2f6fdc35936aab88cf18e4a10173b6a86585f198349066cab74b5

---

## Checklist

---

## Related

- EXE-1377

---

> [!NOTE]
> Integrates first-class realtime support into the core `inngest` package, migrating from the separate `@inngest/realtime` package. Adds typed channels/topics with Standard Schema validation, `inngest.realtime.publish()` for non-durable publishing, `step.realtime.publish()` for durable step publishing, server-side `inngest.realtime.subscribe()` and `inngest.realtime.token()`, and a `useRealtime` React hook. The `@inngest/realtime` package is kept but deprecated. Examples are updated to use the new built-in API.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 84c12178e2f6cab75ec8aa5c379db610f66307d9.</sup>

### DIFF
--- a/.changeset/heavy-clouds-vanish.md
+++ b/.changeset/heavy-clouds-vanish.md
@@ -1,0 +1,5 @@
+---
+"@inngest/test": major
+---
+
+Drop support for inngest SDK v3
--- a/packages/middleware-validation/package.json
+++ b/packages/middleware-validation/package.json
@@ -45,7 +45,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.7.0",
-    "@inngest/test": "workspace:^",
+    "@inngest/test": "0.1.9",
     "@types/eslint__js": "^8.42.3",
     "@types/jest": "^29.5.14",
     "eslint": "^8.30.0",
--- a/packages/test/CLAUDE.md
+++ b/packages/test/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Package Overview
 
-This is the `@inngest/test` package - a testing utility library that provides Jest-compatible mocking and testing tools for Inngest functions. It enables developers to unit test their Inngest functions with proper mocking of step tools, events, and execution state.
+This is the `@inngest/test` package - a testing utility library that provides mocking and testing tools for Inngest functions. It enables developers to unit test their Inngest functions with proper mocking of step tools, events, and execution state.
 
 ## Development Workflow
 
@@ -18,7 +18,7 @@ pnpm install
 ### Common Commands
 
 ```bash
-pnpm test             # Run Jest tests
+pnpm test             # Run Vitest tests
 pnpm build            # Clean build with TypeScript
 pnpm build:clean      # Remove dist directory
 pnpm build:tsc        # TypeScript compilation only
@@ -27,18 +27,18 @@ pnpm pack             # Create inngest-test.tgz for local testing
 
 ### Testing Strategy
 
-- **Unit tests**: Jest (`pnpm test`)
-- Uses Jest for internal testing but designed to be framework-agnostic
+- **Unit tests**: Vitest (`pnpm test`)
+- Designed to be framework-agnostic for end users
 - Supports Jest, Vitest, Bun test, Deno, and Chai
 
 ## Architecture
 
 ### Key Concepts
 - **InngestTestEngine**: Main testing class that provides function execution and mocking
-- **Jest Compatibility**: Designed to work with major testing frameworks
+- **Framework Compatibility**: Designed to work with major testing frameworks
 - **Step Mocking**: Mock individual steps within functions
 - **Event Mocking**: Mock incoming event data
-- **Spy Functions**: Jest-compatible spies for step tools
+- **Spy Functions**: Testing-framework-compatible spies for step tools
 
 ### Core API
 - `InngestTestEngine` - Primary testing interface
@@ -49,7 +49,7 @@ pnpm pack             # Create inngest-test.tgz for local testing
 ### Package Structure
 - `src/InngestTestEngine.ts` - Main testing engine
 - `src/InngestTestRun.ts` - Individual test run implementation
-- `src/spy.ts` - Jest-compatible spy utilities
+- `src/spy.ts` - Spy utilities (compatible with Jest/Vitest/etc.)
 - `src/util.ts` - Testing utilities
 
 ## Usage Patterns
@@ -62,6 +62,6 @@ This package is designed for testing Inngest functions with patterns like:
 
 ## Dependencies
 
-- Requires `inngest@>=3.22.12` as peer dependency
-- Uses `tinyspy` for Jest-compatible spying
+- Requires `inngest@^4.0.0` as peer dependency
+- Uses `tinyspy` for spy/mock functionality
 - Generates unique IDs with `ulid`
--- a/packages/test/jest.config.js
+++ b/packages/test/jest.config.js
@@ -1,8 +1,0 @@
-/** @type {import('ts-jest').JestConfigWithTsJest} **/
-module.exports = {
-  testEnvironment: "node",
-  transform: {
-    "^.+.tsx?$": ["ts-jest", {}],
-  },
-  roots: ["<rootDir>/src"],
-};
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -8,7 +8,7 @@
     "registry": "https://registry.npmjs.org"
   },
   "scripts": {
-    "test": "jest",
+    "test": "vitest run",
     "build": "pnpm run build:clean && pnpm run build:tsc",
     "build:clean": "rm -rf ./dist",
     "build:tsc": "tsc --project tsconfig.build.json",
@@ -18,9 +18,9 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "require": "./dist/index.js",
-      "import": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      "import": "./dist/index.js"
     }
   },
   "files": [
@@ -40,14 +40,15 @@
   "author": "Inngest Inc. <hello@inngest.com>",
   "license": "Apache-2.0",
   "dependencies": {
-    "inngest": "^3.42.3",
     "tinyspy": "^3.0.2",
     "ulid": "^2.3.0"
   },
   "devDependencies": {
-    "@types/jest": "^30.0.0",
-    "jest": "^29.5.0",
-    "prettier": "^3.1.0",
-    "typescript": "^5.9.2"
+    "inngest": "^4.0.0",
+    "typescript": "^5.9.2",
+    "vitest": "^3.0.0"
+  },
+  "peerDependencies": {
+    "inngest": "^4.0.0"
   }
 }
--- a/packages/test/src/InngestTestEngine.test.ts
+++ b/packages/test/src/InngestTestEngine.test.ts
@@ -1,5 +1,6 @@
 import { Inngest } from "inngest";
-import { InngestTestEngine } from "../";
+import { describe, expect, it } from "vitest";
+import { InngestTestEngine } from "./index";
 
 describe("InngestTestEngine", () => {
   describe("lazy execution of mocked steps", () => {
@@ -10,8 +11,7 @@ describe("InngestTestEngine", () => {
       const executionOrder: string[] = [];
 
       const fn = inngest.createFunction(
-        { id: "test-fn" },
-        { event: "test/event" },
+        { id: "test-fn", triggers: [{ event: "test/event" }] },
         async ({ step }) => {
           const step1Result = await step.run("step-1", () => {
             executionOrder.push("step-1-unmocked");
@@ -78,8 +78,7 @@ describe("InngestTestEngine", () => {
       const executionOrder: string[] = [];
 
       const fn = inngest.createFunction(
-        { id: "test-fn" },
-        { event: "test/event" },
+        { id: "test-fn", triggers: [{ event: "test/event" }] },
         async ({ step }) => {
           const step1Result = await step.run("step-1", () => {
             executionOrder.push("step-1-unmocked");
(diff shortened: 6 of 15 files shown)


jpwilliams (commented):
Should we keep `@inngest/realtime` around so that we can patch it if we need to? Also so we can mark it as deprecated in the `README.md` etc.

I think we've done the same with `@inngest/middleware-validation`, as that's also in the same boat.

jacobheric:
> Should we keep `@inngest/realtime` around so that we can patch it if we need to? Also so we can mark it as deprecated in the `README.md` etc.
> 
> I think we've done the same with `@inngest/middleware-validation`, as that's also in the same boat.

I can put it back and mark it as deprecated. Last time we chatted we decided to pull it altogether in since it was going in on top of v4.

jacobheric:
> > Should we keep `@inngest/realtime` around so that we can patch it if we need to? Also so we can mark it as deprecated in the `README.md` etc.
> > I think we've done the same with `@inngest/middleware-validation`, as that's also in the same boat.
> 
> I can put it back and mark it as deprecated. Last time we chatted we decided to pull it altogether in since it was going in on top of v4.

Done!

realtime-spec-updated.md
```diff
@@ -0,0 +1,819 @@
+# Realtime SDK DX Improvements [Draft]
```
jacobheric: I will remove this one before merging. Just keeping it here for ref for now.

packages/inngest/src/components/realtime/subscribe/index.ts
```diff
@@ -0,0 +1,138 @@
...
+export function subscribe<
+  const InputChannel extends Realtime.ChannelInput,
+  const InputTopics extends ChannelTopicsInput<InputChannel>,
+  const TToken extends Realtime.Subscribe.Token<InputChannel, InputTopics>,
+  const TOutput extends Realtime.Subscribe.StreamSubscription<TToken>,
+>(
+  token: SubscribeBaseArgs<InputChannel, InputTopics>,
+  callback?: Realtime.Subscribe.Callback<TToken>,
+): Promise<TOutput>;
+export async function subscribe<
```
jacobheric, line 60: Subscribe supports stream, callback or hybrid mode.

packages/inngest/src/components/realtime/subscribe/index.ts
```diff
@@ -0,0 +1,138 @@
+import type { Inngest } from "../../Inngest.ts";
+import type { Realtime } from "../types.ts";
+import { TokenSubscription } from "./TokenSubscription.ts";
+
+type ChannelTopicNames<InputChannel extends Realtime.ChannelInput> = Extract<
```
jacobheric, line 5: Topic keys are inferred from the channel types.

packages/inngest/src/components/realtime/subscribe/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  } else {
+    callbackStream.cancel("Not needed");
+  }
+
+  if (token.onMessage) {
+    retStream.cancel("Callback subscription doesn't expose a stream");
+    return extras;
+  }
+
+  return Object.assign(retStream, extras) as unknown as TOutput;
```
jacobheric, line 104: Lemme know if you want me to tighten up and try to make it properly generic over TToken. I see similar casting elsewhere, like here: https://github.com/inngest/inngest-js/blob/v4/packages/inngest/src/components/Inngest.ts#L619 so I didn't go to far down on this one.

packages/inngest/src/components/realtime/subscribe/StreamFanout.ts
```diff
@@ -0,0 +1,56 @@
+export class StreamFanout<TInput = unknown> {
```
jacobheric, line 1: This is pretty much the same as the old impl with the addition of a bit more defense on closing. Old for ref:https://github.com/inngest/inngest-js/blob/v4/packages/realtime/src/subscribe/StreamFanout.ts.

packages/inngest/src/components/realtime/subscribe/TokenSubscription.ts
```diff
@@ -0,0 +1,554 @@
+import type { StandardSchemaV1 } from "@standard-schema/spec";
```
jacobheric, line 1: Compare to old impl: https://github.com/inngest/inngest-js/blob/v4/packages/realtime/src/subscribe/TokenSubscription.ts. 

I'll point out some diffs below.

packages/inngest/src/components/realtime/subscribe/TokenSubscription.ts
```diff
@@ -0,0 +1,554 @@
...
+      );
+    }
+
+    url.protocol = url.protocol === "http:" ? "ws:" : "wss:";
+    url.searchParams.set("token", token);
+
+    return url;
+  }
+
+  private isExpectedChannel(channel: string): boolean {
```
jacobheric, line 108: Some additional guardrails here.

packages/inngest/src/components/realtime/subscribe/TokenSubscription.ts
```diff
@@ -0,0 +1,554 @@
...
+
+      if (!key) {
+        throw new Error(
+          "No subscription token key passed and failed to retrieve one automatically",
+        );
+      }
+    }
+
+    const ret = createDeferredPromise<void>();
+    let isConnectSettled = false;
```
jacobheric, line 146: Additional lifecycle checks here compared to before.

packages/inngest/src/components/realtime/subscribe/TokenSubscription.ts
```diff
@@ -0,0 +1,554 @@
...
+
+    //
+    // Fallback: try fetching directly using env-based signing keys.
+    // This path is used when no Inngest client is available.
+    throw new Error(
+      "No getSubscriptionToken handler provided. Pass an Inngest client or provide a token key.",
+    );
+  }
+
+  public close(reason = "Userland closed connection") {
```
jacobheric, line 477: More aggressive cleanup than before.

packages/inngest/src/components/realtime/channel.ts
```diff
@@ -0,0 +1,83 @@
...
+    accessors[topicName] = {
+      channel: channelName,
+      topic: topicName,
+      config,
+    };
+  }
+  return accessors;
+};
+
+export const channel = <
```
jacobheric, line 40: Big change here: declarative object instantiation instead of the old builder pattern. 

https://github.com/inngest/inngest-js/blob/v4/packages/realtime/src/channel.ts#L7

packages/inngest/src/components/realtime/channel.ts
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    Object.defineProperties(def, {
+      topics: { value: topics, enumerable: true },
+      $infer: { get: () => topics },
+      $params: { get: () => undefined },
+    });
+
+    // biome-ignore lint/suspicious/noExplicitAny: sacrifice for clean generics
+    return def as any;
```
jacobheric, line 67: Again shout if you want me to do these more properly.

packages/inngest/src/components/realtime/channel.ts
jacobheric: Also, implementation is quite a bit simpler here, less proxy/dynamic behavior and validation has moved elsewhere, e.g.: https://github.com/inngest/inngest-js/pull/1342/changes#diff-27229b025aa4e81ba7b2dbca6b99645a94323919e9cebf0f53ba16b3b8e0b10eR771

packages/inngest/src/react.ts
```diff
@@ -0,0 +1,653 @@
...
+  token: tokenInput,
+  key,
+  enabled = true,
+  bufferInterval = 0,
+  validate = true,
+  historyLimit = 100,
+  reconnect = true,
+  reconnectMinMs = 250,
+  reconnectMaxMs = 5_000,
+  pauseOnHidden = true,
```
jacobheric, line 279: Good citizen behavior, by default we pause on hidden so we don't stream in inactive tabs.

packages/inngest/src/react.ts
```diff
@@ -0,0 +1,653 @@
...
+  const [allMessages, setAllMessages] = useState<Realtime.Message[]>([]);
+  const [messageDelta, setMessageDelta] = useState<Realtime.Message[]>([]);
+  const [messagesByTopic, setMessagesByTopic] = useState<
+    Record<string, Realtime.Message | undefined>
+  >({});
+  const [lastMessage, setLastMessage] = useState<Realtime.Message | null>(null);
+  const [error, setError] = useState<Error | null>(null);
+  const [connectionStatus, setConnectionStatus] =
+    useState<UseRealtimeConnectionStatus>("idle");
+  const [pauseReason, setPauseReason] = useState<UseRealtimePauseReason>(null);
```
jacobheric, line 297: distinguishes hidden versus disabled.

packages/inngest/src/react.ts
```diff
@@ -0,0 +1,653 @@
...
+        setAllMessages((prev) =>
+          clampMessages(prev, [message], messageLimitRef.current),
+        );
+        setLastMessage(message);
+        return;
+      }
+
+      messageBufferRef.current.push(message);
+    };
+
```
jacobheric, line 529: main connection loop: read token, subscribe and read messages until close or cancel.

jacobheric: on disconnect/error, clean up and backoff + reconnect until cancelled.

jacobheric:
@mendral-app can you review this PR?

packages/inngest/src/components/realtime/subscribe/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    token: token as Realtime.Subscribe.Token,
+    apiBaseUrl: app?.apiBaseUrl,
+    getSubscriptionToken,
+    validate: token.validate,
+  });
+
+  const retStream = subscription.getJsonStream();
+  const callbackStream = subscription.getJsonStream();
+
+  await subscription.connect();
```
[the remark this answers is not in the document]
jacobheric: 1. getJsonStream() returns a ReadableStream that tees from the subscription's internal source stream. If connect() rejects, the underlying connection was never established, so the source stream either never existed or was already closed/errored.
 3. When connect() throws, the await on line 86 propagates the exception, and the function returns a rejected promise. The two ReadableStream objects (retStream, callbackStream) become unreachable local variables eligible for GC, there's no persistent reference holding them alive.
4. Even if you moved stream creation after connect(), you'd have the same situation in reverse -- if stream creation failed, you'd have an open connection to clean up.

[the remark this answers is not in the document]
jacobheric: 1. getJsonStream() returns a ReadableStream that tees from the subscription's internal source stream. If connect() rejects, the
  underlying connection was never established, so the source stream either never existed or was already closed/errored.
  2. When connect() throws, the await on line 86 propagates the exception, and the function returns a rejected promise. The two
  ReadableStream objects (retStream, callbackStream) become unreachable local variables eligible for GC -- there's no persistent
  reference holding them alive.
  3. Even if you moved stream creation after connect(), you'd have the same situation in reverse -- if stream creation failed, you'd
  have an open connection to clean up.

packages/inngest/src/react.ts
```diff
@@ -0,0 +1,649 @@
...
+  pauseReason: UseRealtimePauseReason;
+
+  messages: {
+    byTopic: MessagesByTopicMap<TChannel, TTopics>;
+    all: SubscribedMessage<TChannel, TTopics>[];
+    last: SubscribedMessage<TChannel, TTopics> | null;
+    delta: SubscribedMessage<TChannel, TTopics>[];
+  };
+
+  result: unknown;
```
jpwilliams, line 96: Is this the output of the function?

jacobheric: yes, and `inferRunLifecycleUpdate` will set it from output if output exists.

PR #1345, opened 2026-02-26
Make `INNGEST_CONNECT_ISOLATE_EXECUTION` opt-out

Linell:
## Summary

Make the connection isolation opt out, but provide a working way to provide the workers with the gateway rewrite. 

## Checklist

## Related
[EXE-1320](https://linear.app/inngest/issue/EXE-1320/make-inngest-connect-isolate-execution-opt-out)

### DIFF
--- a/examples/connect-example/index.ts
+++ b/examples/connect-example/index.ts
@@ -54,9 +54,6 @@ connect({
     },
   ],
   instanceId: "my-worker",
-  rewriteGatewayEndpoint: (endpoint) => {
-    return endpoint.replace("connect-gateway:8080", "localhost:8100");
-  },
 }).then(async (conn) => {
   console.log("Connected!");
 
--- a/packages/inngest/src/components/connect/index.ts
+++ b/packages/inngest/src/components/connect/index.ts
@@ -121,6 +121,21 @@ class WebSocketWorkerConnection implements WorkerConnection {
       }
     }
 
+    // Check for worker thread env var (opt-out: false/0 disables isolation)
+    if (options.isolateExecution === undefined) {
+      const envValue = env[envKeys.InngestConnectIsolateExecution];
+      if (envValue === "0" || envValue === "false") {
+        options.isolateExecution = false;
+      }
+    }
+
+    if (options.gatewayUrl === undefined) {
+      const envValue = env[envKeys.InngestConnectGatewayUrl];
+      if (envValue) {
+        options.gatewayUrl = envValue;
+      }
+    }
+
     return options;
   }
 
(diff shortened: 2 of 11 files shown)


packages/inngest/src/components/connect/index.ts
```diff
@@ -334,23 +501,822 @@ class WebSocketWorkerConnection implements WorkerConnection {
...
+      this.excludeGateways.add(startResp.gatewayGroup);
+      rejectWebsocketConnected?.(
+        new ReconnectError(`Connection ${connectionId} timed out`, attempt),
+      );
+    }, 10_000);
+
+    const finalEndpoint = resolveGatewayEndpoint(
+      startResp.gatewayEndpoint,
+      this.options.gatewayUrl,
+      allProcessEnv(),
```
amh4r: Should we handle this the "code option or env var" stuff in `applyDefaults`? That function looks like the preexisting place where we enrich options with env vars

packages/inngest/src/components/connect/types.ts
```diff
@@ -33,7 +33,17 @@ export interface ConnectHandlerOptions extends RegisterOptions {
    */
   handleShutdownSignals?: string[];
 
-  rewriteGatewayEndpoint?: (endpoint: string) => string;
+  /**
+   * Override the gateway WebSocket endpoint. When set, this URL is used
+   * instead of the endpoint returned by the Inngest API.
+   *
+   * Useful for local development with Docker where the gateway hostname
+   * needs to be rewritten (e.g. `ws://localhost:8100`).
```
amh4r: This might be a little misleading, since users may infer that they _need_ to set this for local dev. AFAICT, it isn't typically necessary for local dev because our REST API infers the Connect gateway origin via the request's host header.

Might want to ask Bruno if he can think of a good use case for this. Maybe the only use case is when there's a proxy between the worker and gateway?

packages/inngest/src/components/connect/types.ts
```diff
@@ -42,9 +52,9 @@ export interface ConnectHandlerOptions extends RegisterOptions {
    *
    * Only works in Node.js environments that support worker_threads.
```
amh4r:
```suggestion
   * Only works in environments that support worker_threads.
```

It should work in Deno and Bun. Can you fix this comment in your PR? I'm pretty sure I wrote it originally 😅

examples/connect-example/index.ts
```diff
@@ -54,9 +54,7 @@ connect({
     },
   ],
   instanceId: "my-worker",
-  rewriteGatewayEndpoint: (endpoint) => {
-    return endpoint.replace("connect-gateway:8080", "localhost:8100");
-  },
+  gatewayUrl: "ws://localhost:8100",
```
amh4r:
```suggestion
```

Let's remove this since it breaks the example

PR #1347, opened 2026-02-27
feat: default checkpointing on

Linell:
## Summary
Turn checkpointing on by default. 

Before/After, one `pnpm i` apart:
Screenshot 2026-02-26 at 4 29 07 PM


Linell:
... oh I just saw your #1346 @amh4r 😢 ☠️

PR #1350, opened 2026-02-27
Fix false NESTING_STEPS

amh4r:
# Summary
Fix false `NESTING_STEPS` warning when checkpointing is enabled and there are parallel steps.

# Context
The following function reproduces the issue:
```ts
inngest.createFunction(
  {
    id: "fn-1",
    retries: 0,
    checkpointing: true,
  },
  { event: "event-1" },
  async ({ step }) => {
    await step.run("step-1", () => "result-1");
    await step.run("step-2", () => "result-2");

    await Promise.all([
      step.run("parallel-a", () => "a"),
      step.run("parallel-b", () => "b"),
    ]);
  }
);
```

### DIFF
--- a/.changeset/few-mails-stare.md
+++ b/.changeset/few-mails-stare.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix false NESTING_STEPS error


jpwilliams:
Careful this doesn't cause a regression of #1318.

amh4r:
> Careful this doesn't cause a regression of #1318.

I confirmed that `step` is defined for steps in `transformOutput`

PR #1352, opened 2026-03-02
Experimentation Steps

AndyInternet:
## Summary
This PR introduces a native experimentation framework to the Inngest TypeScript SDK, allowing developers to run A/B tests and feature experiments directly within durable functions with guaranteed consistency and observability.

## 1. Core Tool: `group.experiment()`

The `group.experiment()` tool is the primary entry point for running experiments. It coordinates variant selection and execution while ensuring durability.

- **Durable Selection**: The selection process is wrapped in a memoized step. This ensures that once a variant is chosen for a specific run, the same variant is used across all replays, even if the selection logic involves randomness.
- **Flexible Execution**: Each variant is defined as a callback. The selected variant's callback is executed at the top level, allowing its internal `step.*` calls to participate in normal step discovery.
- **Result Handling**: By default, it returns the result of the selected variant's callback. Using `withVariant: true` allows the caller to receive both the result and the name of the selected variant.

### Example Usage:
```ts
const result = await group.experiment("checkout-flow", {
  variants: {
    control: () => step.run("old-flow", () => oldLogic()),
    treatment: () => step.run("new-flow", () => newLogic()),
  },
  select: experiment.weighted({ control: 50, treatment: 50 }),
});
```

## 2. Selection Strategies (`experiment.*`)

The PR provides a set of pre-built strategies for variant selection:

- **`experiment.fixed(variantName)`**: Always selects the specified variant. Useful for manual overrides or testing.
- **`experiment.weighted(weights)`**: Performs weighted random selection. It is deterministic per run because it is seeded with the Inngest `runId`.
- **`experiment.bucket(value, options?)`**: Uses consistent hashing (SHA-256) to map a value (like a `userId`) to a variant. The same input value will always result in the same variant for a given set of weights.
- **`experiment.custom(fn)`**: Allows developers to provide their own selection logic (e.g., fetching a flag from an external provider). The result is still memoized durably by Inngest.

## 3. Observability and Metadata

The framework is deeply integrated with the Inngest platform for better visibility:

- **Automatic Metadata**: The selection step automatically carries `inngest.experiment` metadata, including:
  - Experiment name
  - Selected variant
  - Selection strategy used
  - Available variants and their weights
- **Context Propagation**: Using `AsyncLocalStorage`, the experiment context (experiment ID and variant name) is propagated to all steps executed within the selected variant's callback. This allows the Inngest UI to group and attribute these steps to the experiment.
- **Dashboard Warnings**: Includes built-in warnings, such as when `experiment.bucket()` receives a null/undefined value.

## 4. Reliability and Safety

- **Zero-Step Guard**: To ensure durability, the SDK detects if a variant callback completes without invoking any Inngest step tools. Since non-step logic re-runs on every replay, the SDK throws a `NonRetriableError` to force developers to wrap their variant logic in `step.run()`.
- **Validation**:
  - Ensures at least one variant is defined.
  - Validates that weights are non-negative and have a positive total.
  - Verifies that `custom` or `bucket` strategies return a valid variant name defined in the experiment.
- **Runtime Support**: Requires `AsyncLocalStorage` for context propagation and zero-step detection, falling back gracefully where possible in restricted environments.

## Checklist

- [x] Added a [docs PR](https://github.com/inngest/website) that references this PR - [docs here](https://github.com/inngest/website/pull/1444)

## Related

- INN-

### DIFF
--- a/packages/inngest/src/components/InngestStepTools.ts
+++ b/packages/inngest/src/components/InngestStepTools.ts
@@ -37,7 +37,6 @@ import {
 } from "./Inngest.ts";
 import { InngestFunction } from "./InngestFunction.ts";
 import { InngestFunctionReference } from "./InngestFunctionReference.ts";
-import { NonRetriableError } from "./NonRetriableError.ts";
 import type { GroupTools } from "./InngestGroupTools.ts";
 import {
   type MetadataBuilder,
@@ -46,6 +45,7 @@ import {
   UnscopedMetadataBuilder,
 } from "./InngestMetadata.ts";
 import type { Middleware } from "./middleware/index.ts";
+import { NonRetriableError } from "./NonRetriableError.ts";
 import type { Realtime } from "./realtime/types.ts";
 import type { EventType } from "./triggers/triggers.ts";
 


packages/inngest/src/components/experiment.test.ts
```diff
@@ -0,0 +1,934 @@
...
+ * Install our test ALS as the global singleton *before* any production
+ * module is imported.  The als.ts module lazily initialises on first
+ * access, so overwriting the cache here guarantees every call to
+ * `getAsyncCtxSync()` / `getAsyncLocalStorage()` resolves to our instance.
+ */
+(globalThis as Record<symbol, unknown>)[alsSymbol] = {
+  promise: Promise.resolve(als),
+  resolved: als,
+  isFallback: false,
+};
```
jakobevangelista, line 28: since this overwrites the global ALS singleton at module load time, this will override any other test that land in the same worker thread, which could lead to unexpected behavior in tests that depend on `als.ts` state

the other tests scope their overrides or use `vi.mock()` which vitest isolates the behavior automatically

should we introduce some sort of cleanup or scoping of behavior to this?

packages/inngest/src/components/InngestGroupTools.ts
```diff
@@ -121,13 +225,199 @@ export interface GroupTools {
...
+        `group.experiment("${stepOpts.id}"): variant "${selectedVariant}" ` +
+          `was selected but is not defined. Available variants: ${variantNames.join(", ")}`,
+      );
+    }
+
+    // Propagate experiment context via ALS so variant sub-steps include
+    // experiment fields in their OutgoingOp.opts. Also track whether any
+    // step tool is invoked to detect zero-step variants.
+    //
+    // TODO: On replay, experimentStepHashedId is undefined because it's
```
jakobevangelista, line 383: noice

just gotta remember to remove this after 👍

packages/inngest/src/components/ExperimentStrategies.ts
```diff
@@ -0,0 +1,200 @@
...
+  weighted(weights: Record<string, number>): ExperimentSelectFn {
+    validateWeights(weights);
+
+    return createSelectFn(
+      () => {
+        const runId =
+          getAsyncCtxSync()?.execution?.ctx.runId ?? crypto.randomUUID();
+        return selectByWeight(hashToFloat(runId), weights);
+      },
+      { strategy: "weighted", weights },
```
jakobevangelista: this might be a nit but i'm thinking of an edge case where since weights are captured by reference here, it's possible that someone could mutate `weights` here after calling `experiment.weighted(weights)` and the runtime behaviour would silently change without revalidation

should we pass by value here instead?

packages/inngest/src/components/ExperimentStrategies.ts
```diff
@@ -0,0 +1,204 @@
...
+    if (options?.weights) {
+      validateWeights(options.weights);
+    }
+
+    const str = value == null ? "" : String(value);
+
+    return createSelectFn(
+      (variantNames?: string[]) => {
+        const weights =
+          options?.weights ??
```
jakobevangelista, line 169: `weighted()` above uses the shallow copy / snapshot

should this also use the snapshot to prevent re-assigning at runtime as well?

jakobevangelista (approved):
awesome! left a comment but the rest lgtm 👍

PR #1359, opened 2026-03-05
fix: parent userland spans under correct step during checkpointing

Linell:
## Summary

Userland OTel spans were orphaned across checkpoint boundaries because the SDK had no way to reference the step spans the executor creates server-side.

- Compute deterministic span IDs matching Go's ChaCha8-based generation _(thank you Claude 🙃)_
- Track active step context to attach inngest.step.parentSpanId to userland spans
- Filter infrastructure spans between checkpoint steps

## Checklist

## Related
* https://linear.app/inngest/issue/EXE-1408/extended-traces-not-working
* https://github.com/inngest/inngest/pull/3804

### DIFF
--- a/packages/inngest/src/helpers/deterministicId.ts
+++ b/packages/inngest/src/helpers/deterministicId.ts
@@ -1,0 +1,357 @@
+/**
+ * Produces span IDs identical to the Go executor's
+ * `DeterministicSpanConfig(seed).SpanID` in `pkg/tracing/tracer.go`.
+ *
+ * Algorithm: SHA-256 the seed, interpret the 32-byte digest as a ChaCha8
+ * key, then read 24 bytes from Go's `chacha8rand` PRNG
+ * (https://c2sp.org/chacha8rand) — 16 bytes for TraceID (discarded) plus
+ * 8 bytes for SpanID (returned).  This is the 3rd uint64 in the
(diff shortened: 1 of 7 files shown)


packages/inngest/src/helpers/deterministicId.ts
```diff
@@ -0,0 +1,357 @@
...
+ * Compute blocks 0 and 1 (counters 0,1) and return the 3rd uint64 (buf[2])
+ * of the interleaved output buffer.  This matches `DeterministicSpanConfig.SpanID`
+ * which reads 16 bytes (TraceID = buf[0..1]) then 8 bytes (SpanID = buf[2]).
+ *
+ * The chacha8rand interleaved layout is:
+ *   buf[2*i]   = block0[i] | block1[i] << 32
+ *   buf[2*i+1] = block2[i] | block3[i] << 32
+ * So buf[2] = buf[2*1] = block0[1] | block1[1] << 32 = s1 from both blocks.
+ */
+function chacha8randThirdUint64(key: number[]): bigint {
```
rhino1998, line 44: Is there a library for this we could/should use?

rhino1998: I looked around and it seems like not really

Linell: ☠️  🐉

PR #1386, opened 2026-03-18
Durable Endpoint Streaming DX 

Linell:
## Summary

Updates the way users can interact with streaming from durable endpoints.

## Checklist

## Related

- INN-

---

> [!NOTE]
> Refactors the Durable Endpoint Streaming DX: replaces the manual ReadableStream + tee approach with `stream.pipe()` / `stream.push()` API supporting ReadableStream, AsyncIterable, and generator factories. Adds per-step channel tagging on SSE frames for client-side rollback, ships a new `inngest/experimental/durable-endpoints` client package with `streamRun`/`subscribeToRun`, migrates the demo app to use `correlationId` instead of `runId` for human-in-the-loop signaling, fixes parallel step execution in `runToCompletion` mode, and removes the `fn_id`/`qi_id` header dependency by gating `shouldAsyncCheckpoint` on `forceExecution`.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit db6d83d8853802138674dfc255707ff3920a4cba.</sup>

### DIFF
--- a/packages/inngest/src/components/execution/engine.ts
+++ b/packages/inngest/src/components/execution/engine.ts
@@ -426,9 +426,7 @@ class InngestExecutionEngine
       if (stepError && this.isFinalAttempt(stepError)) {
         // Permanent failure — close with a failed frame so the client
         // gets onFunctionFailed instead of silence.
-        this.streamTools.closeFailed(
-          errorMessage(stepError),
-        );
+        this.streamTools.closeFailed(errorMessage(stepError));
       } else {
         // End stream without a result frame — client uses redirect_info
         // to reconnect.
@@ -851,9 +849,7 @@ class InngestExecutionEngine
 
         if (this.streamTools.activated) {
           if (isFinal) {
-            this.streamTools.closeFailed(
-              errorMessage(checkpoint.error),
-            );
+            this.streamTools.closeFailed(errorMessage(checkpoint.error));
           } else {
             // Retryable error — suppress the result frame; the run will retry.
             this.streamTools.end();
@@ -1000,9 +996,7 @@ class InngestExecutionEngine
         const isFinal = this.isFinalAttempt(checkpoint.error);
 
         if (isFinal) {
-          this.streamTools.closeFailed(
-            errorMessage(checkpoint.error),
-          );
+          this.streamTools.closeFailed(errorMessage(checkpoint.error));
         } else {
           // Retryable error — suppress the result frame; the run will retry.
           this.streamTools.end();
--- a/packages/inngest/src/stream.test.ts
+++ b/packages/inngest/src/stream.test.ts
@@ -58,7 +58,11 @@ describe("streamRun", () => {
     rs._fromSource(
       framesFrom([
         { type: "stream", data: "chunk" },
-        { type: "inngest.result", status: "failed", error: "Dog Speak is Much Too Hard to Translate" },
+        {
+          type: "inngest.result",
+          status: "failed",
+          error: "Dog Speak is Much Too Hard to Translate",
+        },
       ]),
     );
 
@@ -203,7 +207,11 @@ describe("streamRun", () => {
     // didn't close the connection). The stream should stop after result.
     async function* neverEnding(): AsyncGenerator<SSEFrame> {
       yield { type: "stream", data: "a" } as SSEFrame;
-      yield { type: "inngest.result", status: "succeeded", data: "done" } as SSEFrame;
+      yield {
+        type: "inngest.result",
+        status: "succeeded",
+        data: "done",
+      } as SSEFrame;
       // These should never be reached:
       yield { type: "stream", data: "SHOULD NOT APPEAR" } as SSEFrame;
     }
--- a/packages/inngest/src/test/integration/durableEndpoints/streaming.test.ts
+++ b/packages/inngest/src/test/integration/durableEndpoints/streaming.test.ts
@@ -138,7 +138,10 @@ test(
       (e) => e.event === "inngest.result",
     );
     expect(resultEvents.length).toBe(1);
-    expect(JSON.parse(resultEvents[0]!.data)).toEqual({ status: "succeeded", data: "All done" });
+    expect(JSON.parse(resultEvents[0]!.data)).toEqual({
+      status: "succeeded",
+      data: "All done",
+    });
 
     await state.waitForRunComplete();
   },
@@ -185,7 +188,10 @@ describe("header negotiation", () => {
 
           const results = events.filter((e) => e.event === "inngest.result");
           expect(results.length).toBe(1);
-          expect(JSON.parse(results[0]!.data)).toEqual({ status: "succeeded", data: "done" });
+          expect(JSON.parse(results[0]!.data)).toEqual({
+            status: "succeeded",
+            data: "done",
+          });
         },
       );
 
@@ -253,7 +259,10 @@ describe("header negotiation", () => {
 
           const results = events.filter((e) => e.event === "inngest.result");
           expect(results.length).toBe(1);
-          expect(JSON.parse(results[0]!.data)).toEqual({ status: "succeeded", data: "computed" });
+          expect(JSON.parse(results[0]!.data)).toEqual({
+            status: "succeeded",
+            data: "computed",
+          });
 
           const streamFrames = events.filter((e) => e.event === "stream");
           expect(streamFrames.length).toBe(0);
@@ -408,7 +417,10 @@ describe("streaming functionality", () => {
       // Result is the concatenated pipe output
       const results = events.filter((e) => e.event === "inngest.result");
       expect(results.length).toBe(1);
-      expect(JSON.parse(results[0]!.data)).toEqual({ status: "succeeded", data: "token1token2" });
+      expect(JSON.parse(results[0]!.data)).toEqual({
+        status: "succeeded",
+        data: "token1token2",
+      });
     },
   );
 
@@ -479,7 +491,10 @@ describe("streaming functionality", () => {
       // Result frame
       const results = events.filter((e) => e.event === "inngest.result");
       expect(results.length).toBe(1);
-      expect(JSON.parse(results[0]!.data)).toEqual({ status: "succeeded", data: "finished" });
+      expect(JSON.parse(results[0]!.data)).toEqual({
+        status: "succeeded",
+        data: "finished",
+      });
 
       // redirect_info is always sent once checkpoint creates the run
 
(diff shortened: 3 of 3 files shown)


packages/inngest/src/components/execution/streaming.ts
```diff
@@ -190,3 +272,143 @@ export async function drainStreamWithTimeout(
...
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw.data);
+  } catch {
+    parsed = raw.data;
+  }
+
+  switch (raw.event) {
+    case "inngest.metadata": {
+      const obj = parsed as Record<string, unknown>;
```
amh4r: We should always avoid type casts, since they create an opportunity for static types to deviate from runtime types. Type guards are safer (e.g. `isRecord()`), but the cleanest solution might be to define the SSE event types (e.g. `SSEMetadataFrame`) as zod schemas instead of static-only interfaces

packages/inngest/src/components/InngestStreamTools.ts
```diff
@@ -80,67 +107,118 @@ export class InngestStream {
...
+      } catch {
+        continue;
+      }
+
+      this.writeChain = this.writeChain
+        .then(() => this.writer.write(this.encoder.encode(frame)))
+        .catch(() => {
+          // Writer errored — swallow to keep the read loop draining
+          // so we still collect all chunks for the return value.
+        });
```
amh4r: Should this be an `enqueue()` call instead? Seems to do the same thing

packages/inngest/src/components/execution/streaming.ts
```diff
@@ -190,3 +272,143 @@ export async function drainStreamWithTimeout(
...
+  let buffer = "";
+
+  try {
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+
+      buffer += decoder.decode(value, { stream: true });
+
+      const parts = buffer.split("\n\n");
```
amh4r, line 372: What does `\n\n` mean?

packages/inngest/src/components/execution/streaming.ts
```diff
@@ -190,3 +272,143 @@ export async function drainStreamWithTimeout(
...
+
+// ---------------------------------------------------------------------------
+// SSE line parser (async generator)
+// ---------------------------------------------------------------------------
+
+/**
+ * Parses a `ReadableStream<Uint8Array>` as an SSE byte stream, yielding
+ * `RawSSEEvent` objects for each complete event.
+ */
+export async function* iterSSE(
```
amh4r, line 356:
```suggestion
export async function* iterSse(
```

This is gonna sound annoying, but can we tree initialisms as words in casing? That's idiomatic in TS (unlike Go)

packages/inngest/src/components/execution/engine.ts
```diff
@@ -557,19 +555,20 @@ class InngestExecutionEngine
       return;
     }
 
-    if (!this.state.checkpointedRun?.realtimeToken) {
+    const run = this.state.checkpointedRun;
+    if (!run?.realtimeToken) {
```
amh4r, line 557: (Preexisting issue, but we're changing this line so it's an opportunity to improve)

This should always be a set. We can confirm that later, but in the meantime let's log an internal logger error and add a `TODO` comment about whether we can make `realtimeToken` a required field

packages/inngest/src/components/execution/engine.ts
```diff
@@ -857,9 +851,26 @@ class InngestExecutionEngine
...
         // If stream was already activated, close with an error frame and
-        // terminate the loop. The HTTP response was already sent.
+        // terminate the loop. The HTTP response was already sent via
+        // earlyStreamResponse, so the return value here just ends the
+        // core loop.
         if (this.streamTools.activated) {
           this.streamTools.close({ error: String(checkpoint.error) });
+
+          // Mirror the non-stream path: if not on the final attempt,
+          // inform the IS so it can schedule a retry.
```
amh4r:
```suggestion
          // inform the Inngest Server so it can schedule a retry.
```

Claude probably thought "IS" was a well known acronym, but it isn't. Just a thing to make the spec a little more succinct

packages/inngest/src/components/execution/engine.ts
```diff
@@ -1065,6 +1082,13 @@ class InngestExecutionEngine
             i,
           );
           if (output?.type === "function-resolved") {
+            // When running to completion (DE re-entry), return the
+            // function result directly instead of wrapping it in a
+            // steps-found checkpoint for the IS.
```
amh4r:
```suggestion
            // When running to completion (DurableEndpoint re-entry), return the
            // function result directly instead of wrapping it in a
            // steps-found checkpoint for the Inngest Server.
```

DE is also not a well known acronym

packages/inngest/src/components/execution/engine.ts
```diff
@@ -1186,6 +1266,20 @@ class InngestExecutionEngine
         },
 
         "checkpointing-buffer-interval-reached": () => {
+          // During runToCompletion, fire-and-forget the checkpoint for
+          // observability but don't block the core loop.
+          if (this.options.runToCompletion) {
+            if (this.state.checkpointingStepBuffer.length) {
+              void this.checkpoint(this.state.checkpointingStepBuffer).catch(
+                () => {
+                  // Best-effort: don't crash if the IS rejects this.
```
amh4r: Should we log in the internal logger?

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -999,14 +999,21 @@ export class InngestCommHandler<
         });
       },
       "function-resolved": ({ data }) => {
-        // If the execution returned a Response with a ReadableStream body
-        // (SSE streaming), pipe it through as a streaming response.
-        if (data instanceof Response && data.body instanceof ReadableStream) {
+        // If the execution returned a Response (SSE from the engine or a
+        // user-constructed Response), pass it through directly.
```
amh4r: Which situation would a user not return a `Response` object? Do some frameworks allow that?

packages/inngest/src/components/InngestCommHandler.ts
```diff
@@ -1596,6 +1603,14 @@ export class InngestCommHandler<
...
                 headerKeys.InngestRunId,
               ),
+              fn_id: await actions.headers(
+                "getting function ID for forced execution",
+                headerKeys.InngestFnId,
+              ),
+              qi_id: await actions.headers(
+                "getting queue item ref for forced execution",
+                headerKeys.InngestQueueItemRef,
+              ),
```
amh4r: What's this stuff for?

packages/inngest/src/components/InngestStreamTools.ts
```diff
@@ -80,67 +107,118 @@ export class InngestStream {
...
+      });
+  }
+
+  /**
+   * Emit a step lifecycle SSE frame (`step:running`, `step:completed`,
+   * `step:errored`). Internal use only — called by the execution engine.
+   */
+  stepLifecycle(stepId: string, status: "running"): void;
+  stepLifecycle(stepId: string, status: "completed"): void;
+  stepLifecycle(stepId: string, status: "errored", data: StepErrorData): void;
```
amh4r: Is there a way to avoid overloads? Their ergonomics aren't great in TS

packages/inngest/src/components/InngestStreamTools.ts
```diff
@@ -80,67 +107,118 @@ export class InngestStream {
...
+    // Resolve the source into an AsyncIterable<string>.
+    // Check ReadableStream first — on Node 18+ ReadableStream implements
+    // Symbol.asyncIterator, so the instanceof check must come before
+    // the async-iterable duck-type test.
+    const iterable: AsyncIterable<string> =
+      source instanceof ReadableStream
+        ? this.readableToAsyncIterable(source)
+        : typeof source === "function"
+          ? source()
+          : source;
```
amh4r: Maybe a personal preference, but I find nested ternaries difficult to read

packages/inngest/src/components/InngestStreamTools.ts
```diff
@@ -2,9 +2,24 @@ import { getAsyncCtx, getAsyncCtxSync } from "./execution/als.ts";
...
+  type SSEStepFrame,
+  type StepErrorData,
 } from "./execution/streaming.ts";
 
+/**
+ * Accepted source types for `stream.pipe()`.
+ *
+ * - `ReadableStream` — piped directly
+ * - `AsyncIterable<string>` — iterated; each yielded value becomes a chunk
+ * - `() => AsyncIterable<string>` — factory invoked lazily, then iterated
```
amh4r, line 17: Is the factory to support a user-facing use case? If so, what is it?

packages/inngest/src/components/InngestStreamTools.ts
```diff
@@ -150,14 +228,12 @@ export class InngestStream {
...
-  sendRedirectInfo(data: { run_id: string; token: string; url?: string }): void {
-    const frame = buildSSERedirectFrame(data);
-
-    this.writeChain = this.writeChain
-      .then(() => this.writer.write(this.encoder.encode(frame)))
-      .catch(() => {
-        // Writer already errored/closed — nothing to do.
-      });
+  sendRedirectInfo(data: {
+    run_id: string;
```
amh4r, line 241: Should we use camelCase instead of snake_case? Snake_case isn't idiomatic TS, so it reads wonky in business logic. But also, maybe the SSE event should also have camelCase too?

packages/inngest/src/helpers/consts.ts
```diff
@@ -151,6 +151,8 @@ export enum headerKeys {
   InngestRunId = "x-run-id",
   InngestStepId = "x-inngest-step-id",
   InngestForceStepPlan = "x-inngest-force-step-plan",
+  InngestFnId = "x-inngest-fn-id",
+  InngestQueueItemRef = "x-inngest-queue-item-ref",
```
amh4r: What are these for? Do they exist in the Inngest server?

packages/inngest/src/index.ts
```diff
@@ -53,7 +53,10 @@ export type { InngestFunction } from "./components/InngestFunction";
 export type { InngestFunctionReference } from "./components/InngestFunctionReference";
 export { referenceFunction } from "./components/InngestFunctionReference.ts";
 export { group, step } from "./components/InngestStepTools.ts";
-export type { StreamTools } from "./components/InngestStreamTools.ts";
+export type {
+  PipeSource,
+  StreamTools,
```
amh4r: Which user-facing code needs `PipeSource` and `StreamTools`?

packages/inngest/src/experimental/durable-endpoints.ts
```diff
@@ -0,0 +1,35 @@
+/**
```
amh4r, line 1: Let's export this stuff under `inngest/experimental/durable-endpoints` for now. That gives us flexibility to make breaking changes without releasing v5

packages/inngest/src/components/execution/streaming.ts
Linell: Theoretically that's the separator for different SSE messages and it _does_ work, but it may be smart and even simplify some other pieces if we used something like [eventsource-parser](https://npmx.dev/package/eventsource-parser)

packages/inngest/src/components/InngestCommHandler.ts
Linell: I mentioned it in [Slack](https://inngest.slack.com/archives/C096Q3ZANRZ/p1774032288865259), but it fixes an issue where when we re-enter a durp function after going async, we need the  fn_id and qi_id to correctly go into AsyncCheckpointingMode and create the spans. Here's the corresponding [backend PR](https://github.com/inngest/inngest/pull/3859).

packages/inngest/src/components/InngestStreamTools.ts
Linell: [This one:](https://github.com/inngest/inngest-js/pull/1386/changes#diff-ea1a5d59a93cdcb9b8cfa767597488af92138f21180339e6127724ff95d6d7e7R25-R37)

```ts
  // Option B: Pipe an async generator. Works with any provider that
  // exposes an async iterable of events.
  //
  // return stream.pipe(async function* () {
  //   for await (const event of response) {
  //     if (
  //       event.type === "content_block_delta" &&
  //       event.delta.type === "text_delta"
  //     ) {
  //       yield event.delta.text;
  //     }
  //   }
  // });
```

packages/inngest/src/components/execution/streaming.ts
amh4r: No issues with using `\n\n` like that! But we should at least add a comment explaining why

Linell: That's fair!! It _does_ look a bit cleaner using that library: https://github.com/inngest/inngest-js/pull/1402

packages/inngest/src/components/execution/streaming.ts
Linell: Good idea! Added!

packages/inngest/src/api/api.ts
```diff
@@ -526,13 +526,13 @@ export class InngestApi {
...
     runId: string;
-    fnId: string;
+    fnId?: string;
     queueItemId: string;
     steps: OutgoingOp[];
   }): Promise<void> {
     const body = JSON.stringify({
       run_id: args.runId,
-      fn_id: args.fnId,
+      ...(args.fnId && { fn_id: args.fnId }),
```
amh4r, line 535: Revert this line?

packages/inngest/src/api/api.ts
```diff
@@ -526,13 +526,13 @@ export class InngestApi {
    */
   async checkpointStepsAsync(args: {
     runId: string;
-    fnId: string;
+    fnId?: string;
```
amh4r, line 529:
```suggestion
    fnId: string | undefined;
```

Basically the same thing, but it avoids a footgun where we forget to set `fnId` to something (which can be `undefined`)

packages/inngest/src/components/execution/als.ts
```diff
@@ -36,7 +36,7 @@ export interface AsyncContext {
      * callback. Useful to understand whether we are in the context of a step
      * execution or within the main function body.
      */
-    executingStep?: StepOptions;
+    executingStep?: StepOptions & { hashedId?: string };
```
amh4r, line 39: Revert?

PR #1402, opened 2026-03-23
refactor: clean up SSE parsing

Linell:
Use the [eventsource-parser](https://npmx.dev/package/eventsource-parser) library to clean up the way that we're handling parsing SSEs. Also refactors to use the more idiomatic `fooSse` instead of `fooSSE` that I'd used previously.

---

> [!NOTE]
> Replaces manual SSE parsing with the `eventsource-parser` library across `streaming.ts`, `InngestStreamTools.ts`, `engine.ts`, and the integration test helpers. Also renames all `SSE`-prefixed identifiers to the more idiomatic `Sse` casing, removes the now-redundant `RawSSEEvent` type, and makes `parseSseFrame` private. New unit tests cover chunk-boundary splitting, multi-line data, and unknown event filtering.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit f98ad611cd6fda8b9593bd1829364db2bfc2b0d9.</sup>

### DIFF
--- a/packages/inngest/package.json
+++ b/packages/inngest/package.json
@@ -297,6 +297,7 @@
     "canonicalize": "^1.0.8",
     "cross-fetch": "^4.0.0",
     "debug": "^4.3.4",
+    "eventsource-parser": "^3.0.6",
     "hash.js": "^1.1.7",
     "json-stringify-safe": "^5.0.1",
     "ms": "^2.1.3",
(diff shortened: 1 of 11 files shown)


amh4r:
How necessary is `eventsource-parser`? It's good to avoid adding dependencies unless there's a compelling reason, since every dependency we add becomes a transitive dependency for our users

Linell:
Closing -- was just an exploration and I don't think it's needed right now!

PR #1423, opened 2026-03-27
fix: export JsonifyObject type from inngest/types

### LINKED ISSUE #1273: Export JsonifyObject type from inngest/types
## Description

When using `step.run()` with a return type that contains a recursive/nested object structure stored in a typed JSON column (via Drizzle ORM's `json().$type<T>()`), TypeScript emits error TS4058:

```
TS4058: Return type of exported function has or is using name 'JsonifyObject' from external module "inngest/helpers/jsonify" but cannot be named.
```

This happens because `Jsonify<T>` is exported from `inngest/types`, but `JsonifyObject<T>` (which `Jsonify` uses internally for complex object types) is not exported.

## Minimal Reproduction

```typescript
// schema.ts (Drizzle ORM)
import { pgTable, text, json } from "drizzle-orm/pg-core";

// A recursive type for tree-structured data
interface TreeNode {
  value: string;
  children?: TreeNode[];
}

export const documents = pgTable("documents", {
  id: text("id").primaryKey(),
  tree: json("tree").$type<TreeNode>(),
});

// my-function.ts
import { inngest } from "./client";
import { db } from "./db";
import { documents } from "./schema";
import { eq } from "drizzle-orm";

export const myFunction = inngest.createFunction(
  { id: "my-function" },
  { event: "test/run" },
  async ({ step }) => {
    // This causes TS4058 because the return type includes TreeNode,
    // which gets wrapped in JsonifyObject (not exported)
    const doc = await step.run("fetch-document", async () => {
      return db.query.documents.findFirst({
        where: eq(documents.id, "doc-1"),
      });
    });

    return { doc };
  }
);
```

## Expected Behavior

`JsonifyObject` should be exported alongside `Jsonify` from `inngest/types`, allowing TypeScript to properly emit declaration files for functions that return complex nested types.

## Current Workaround

We have to manually construct the type using `Jsonify` for the base fields and bypass the problematic field:

```typescript
import { Jsonify } from "inngest/types";

type JsonifiedDocument = Jsonify<Omit<typeof documents.$inferSelect, "tree">> & {
  tree: (typeof documents.$inferSelect)["tree"];
};

// Then cast: return doc as JsonifiedDocument;
```

This is fragile and requires manual maintenance whenever the schema changes.

## Suggested Fix

Export `JsonifyObject` from `inngest/types`:

```typescript
// inngest/types.ts
export type { JsonifyObject } from "./helpers/jsonify";
```

claygeo:
## Problem

When `step.run()` returns a type with recursive or nested object structures (e.g., Drizzle ORM `json().$type<TreeNode>()`), TypeScript emits **TS4058**:

```
error TS4058: Return type of exported function has or is using name 'JsonifyObject'
from external module "inngest/types" but cannot be named.
```

This happens because `Jsonify<T>` is exported from `inngest/types`, but `JsonifyObject<T>` (which `Jsonify` resolves to internally for object types) is not. TypeScript needs all types in the resolution chain to be exported when generating declaration files.

### Reproduction

```typescript

type TreeNode = {
  id: string;
  children: TreeNode[]; // recursive
};

const inngest = new Inngest({ id: "app" });

  { id: "test" },
  { event: "test/run" },
  async ({ step }) => {
    const result = await step.run("get-tree", (): TreeNode => ({
      id: "root",
      children: [],
    }));
    return result; // TS4058: cannot name 'JsonifyObject'
  }
);
```

## Solution

Export `JsonifyObject` from `helpers/jsonify.ts` and re-export it from `types.ts` alongside the existing `Jsonify` export.

## Changes

| File | Change |
|------|--------|
| `packages/inngest/src/helpers/jsonify.ts` | Added `export` to `JsonifyObject` type (line 47) |
| `packages/inngest/src/types.ts` | Added `JsonifyObject` to the re-export from `./helpers/jsonify.ts` |

## Notes

- This is backwards-compatible. Adding an export does not affect existing code.
- The type constraint `JsonifyObject<T extends object>` is unchanged.
- Previous PR #972 attempted a broader change. This PR is scoped to just the missing export.
- Tracked as EXE-1211 in Linear.

Closes #1273

---

> [!NOTE]
> Exports `JsonifyObject` from `helpers/jsonify.ts` and re-exports it from `types.ts` to fix TS4058 errors when `step.run()` returns recursive/nested object types. The type was already used internally but not publicly accessible, causing TypeScript to fail when generating declaration files.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 7c0ff64d33a07944724e2bf050ed797ba5d24164.</sup>

### DIFF
--- a/packages/inngest/src/helpers/jsonify.ts
+++ b/packages/inngest/src/helpers/jsonify.ts
@@ -44,7 +44,7 @@ type FilterJsonableKeys<T extends object> = {
 /**
 JSON serialize objects (not including arrays) and classes.
 */
-type JsonifyObject<T extends object> = {
+export type JsonifyObject<T extends object> = {
   [Key in keyof Pick<T, FilterJsonableKeys<T>>]: Jsonify<T[Key]>;
 };
 
--- a/packages/inngest/src/types.ts
+++ b/packages/inngest/src/types.ts
@@ -35,7 +35,7 @@ import type {
 } from "./helpers/types.ts";
 import type { Logger } from "./middleware/logger.ts";
 
-export type { Jsonify } from "./helpers/jsonify.ts";
+export type { Jsonify, JsonifyObject } from "./helpers/jsonify.ts";
 export type { SimplifyDeep } from "./helpers/types.ts";
 
 const baseJsonErrorSchema = z.object({


jpwilliams:
Fixed in:
- #1432 
- #1433

PR #1480, opened 2026-04-24
[SYS-815] Connect - Log Remaining In Flight and Remove Phantom Leases

mar-inngest:
## Checklist

## Related

- INN-

---

> [!NOTE]
> Adds verbose shutdown diagnostics to the Connect SDK: per-request debug logs during drain via `dumpInFlightForShutdown`, a 60s periodic interval timer, structured `WakeReason` logging at every reconcile-loop wake site, `leaseAcquiredAt`/`leaseLastExtendedAt` timestamps on `InFlightRequest`, and a fix for phantom lease resurrection by ignoring late extend-lease ACKs for already-cleaned-up requests. Also wakes the reconcile loop on lease-lost during shutdown so `close()` doesn't hang. The final commit adds a `closeOnce` idempotency guard and a `pendingWakeReasons` accumulator to coalesce concurrent wakes.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 88c21228402c29d743b25f9e230299de1b18f2ed.</sup>


packages/inngest/src/components/connect/strategies/core/requestProcessor.ts
```diff
@@ -180,7 +182,14 @@ export class RequestProcessor {
       };
 
       this.logger.debug(
-        { connectionId: latestConn.id, leaseId: currentLeaseId },
```
mar-inngest, line 183: adding more logging information here

BrunoScheufler (approved):

packages/inngest/src/components/connect/strategies/core/connection.ts
```diff
@@ -269,7 +279,14 @@ export class ConnectionCore {
...
       "Shutting down, waiting for in-flight requests",
     );
+    // Flip the shutdown flag before starting any timers so the periodic
+    // dump guard (`if (!this._shutdownRequested) return`) cannot observe a
+    // stale `false` on its first tick.
     this._shutdownRequested = true;
+    // Verbose per-request dump (debug-only) at drain start so operators can
+    // immediately see which runs are holding the shutdown.
+    this.dumpInFlightForShutdown("drain-start");
+    this.startShutdownInFlightDumpTimer();
```
amh4r, line 297: Is this stuff idempotent? Just in case, it seems like we should make this whole `close()` method idempotent

packages/inngest/src/components/connect/strategies/core/connection.ts
```diff
@@ -315,7 +332,8 @@ export class ConnectionCore {
     this.wakeSignal = { promise, resolve: resolve! };
   }
 
-  private wake(): void {
+  private wake(reason: WakeReason = WAKE_REASON.Unknown): void {
+    this.pendingWakeReasons.push(reason);
     this.wakeSignal.resolve();
     this.resetWakeSignal();
```
amh4r: Is this problematic? We reset the wake signal immediately after resolving it, so doesn't that mean `await this.wakeSignal.promise` can get stuck?

For example, we may be somewhere above `await this.wakeSignal.promise` in the reconcile loop (e.g. establishing the connection) when `wake()` is called. Then when we get down to `await this.wakeSignal.promise` we're stuck waiting for the _next_ wake

amh4r: If this is an issue, it's preexisting. However, this PR adds more `this.wake()` calls, which makes the issue more likely (if it's truly an issue)

packages/inngest/src/components/connect/strategies/core/connection.ts
mar-inngest: Good call! Lemme make `close()` idempotent

packages/inngest/src/components/connect/strategies/core/connection.ts
mar-inngest: updated, and added some guard rails around it

PR #1481, opened 2026-04-24
fix: HMAC signature verification tweak

Linell:
## Summary

Fix timing side-channel in HMAC signature verification and harden our signature timestamp verification.

## Checklist

## Related

[EXE-1652: Timing attack in HMAC signature comparison (InngestCommHandler)](https://linear.app/inngest/issue/EXE-1652/timing-attack-in-hmac-signature-comparison-inngestcommhandler)

---

> [!NOTE]
> Fixes a timing side-channel in HMAC signature verification by replacing `===` with a constant-time XOR accumulator (`timingSafeEqual`), and hardens timestamp validation with a radix-10 guard on `parseInt`, a `Number.isFinite` check for unparseable values, and `Math.abs` clamping to reject future-dated timestamps. Also fixes the `signDataWithKey` call site and test helper to pass seconds (not milliseconds) as the timestamp.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit e14963712b01539c3848baf0326d7ed82c9e19df.</sup>

### DIFF
--- a/.changeset/timing-safe-signature-compare.md
+++ b/.changeset/timing-safe-signature-compare.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Improves HMAC signature verification by using a constant-time comparison, which mitigates a potential timing-based signature-recovery attack against the request signature. Also improves handling of timestamps in signatures, including malformed or future-dated values.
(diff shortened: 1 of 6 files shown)


.changeset/timing-safe-signature-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"inngest": minor
```
amh4r: Isn't this a patch?

amh4r (approved):

PR #1439, opened 2026-04-03
Export missing types from main entry point (MaybePromise, GroupExperiment, ParallelOptions)

nicksrandall:
## Summary

Several types used in the public API surface are not re-exported from the main package entry (`src/index.ts`):

- **`MaybePromise`** — used in return types of `dependencyInjectionMiddleware`, `Inngest` class methods, and `createFunction`
- **`GroupExperiment`** — used in `GroupTools.experiment`
- **`ParallelOptions`** — used in `GroupTools.parallel`

This causes **TS2883** errors during declaration emit when consumers use TypeScript's composite projects (`tsgo` or `tsc` with `declaration: true`), because these types cannot be portably referenced from their internal module paths (which are not in the package `exports` map).

## Error reproduced

```
error TS2883: The inferred type of 'inngest' cannot be named without a reference to
'MaybePromise' from 'inngest/helpers/types'. This is likely not portable.
A type annotation is necessary.
```

## Fix

Add the three missing types to the re-exports in `index.ts`:
- `MaybePromise` from `./helpers/types` (alongside existing `ExclusiveKeys`, `IsStringLiteral`, etc.)
- `GroupExperiment` and `ParallelOptions` from `./components/InngestGroupTools` (alongside existing `GroupTools`, `GroupToolsDeps`)

## Test plan

- [x] Verify types appear in the generated `dist/index.d.ts` after build
- [x] Confirm no TS2883 errors in a consumer project using `declaration: true` with `tsgo`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> Exports `MaybePromise`, `GroupExperiment`, and `ParallelOptions` from the main `src/index.ts` entry point to fix TS2883 declaration emit errors in consumer projects using TypeScript composite projects with `declaration: true`.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit a51f772722a6ecdd622185ae91ce06fac10c2e6c.</sup>

PR #1520, opened 2026-05-11
fix(types): re-export InngestApi, MaybePromise, GroupExperiment, ParallelOptions from main entry

### LINKED ISSUE #1510: [BUG] step.sendSignal return type leaks internal InngestApi namespace and causes error TS2742 for consumers
**Describe the bug**

Starting in `inngest@4.2.5`, any consumer project that compiles with `declaration: true` (or `composite: true`) and uses `inngest.createFunction(...)` fails to emit `.d.ts` with Typescript error TS2742:

```
error TS2742: The inferred type of '<name>' cannot be named without a reference to
'../../node_modules/inngest/api/api'. This is likely not portable.
A type annotation is necessary.
```

The leak comes from `step.sendSignal()`'s return type, which was changed from `Promise<null>` to `Promise<InngestApi.SendSignalResponse>` in #1479. `InngestApi` is defined in `packages/inngest/src/api/api.ts`, which is **not** in the package's public `exports` map, so TypeScript can't write a portable module specifier for it when emitting consumer declaration files. It falls back to a relative `node_modules/inngest/api/api` reference, which trips TS2742.

This is a clear regression since the same consumer code emits fine on `4.2.4` and earlier.

**To Reproduce**

This issue can be demonstrated in one file (plus a `tsconfig.json`):

`src/index.ts`:
```ts

const inngest = new Inngest({ id: "demo" });

  { id: "my-fn", triggers: [{ event: "foo" }] },
  async (ctx) => ctx,
);
```

`tsconfig.json`:
```jsonc
{
  "compilerOptions": {
    "declaration": true,
    "outDir": "dist",
    "module": "Preserve",
    "moduleResolution": "Bundler",
    "strict": true,
    "target": "ES2022"
  },
  "include": ["src"]
}
```

```bash
$ npm i inngest@4.2.6
$ npx tsc

src/index.ts:5:14 - error TS2742: The inferred type of 'fn' cannot be named
without a reference to '../../node_modules/inngest/api/api'. This is likely
not portable. A type annotation is necessary.
```

The function does not even need to call `step.sendSignal` since `step` is part of the inferred handler type returned by `createFunction`.  So the leak cascades into every consumer's `.d.ts` regardless of whether they use the signal API.

**Expected behavior**

`tsc` emits the consumer's `.d.ts` cleanly, as it did in `4.2.4` and earlier (when `step.sendSignal` returned `Promise<null>`). The TypeScript compiler should be able to resolve any type referenced from the SDK's publicly exported API.

**TL;DR**

Just need a one-line addition to `packages/inngest/src/index.ts`:

```diff
 export * from "@inngest/ai";
 export type { StandardSchemaV1 } from "@standard-schema/spec";
+export type { InngestApi } from "./api/api.ts";
 export { experiment } from "./components/ExperimentStrategies.ts";
```

**Verified leak path**

In the published `4.2.5+` `.d.ts` tree, `InngestApi` reaches the public surface at exactly two points:

- `node_modules/inngest/components/InngestStepTools.d.ts:216`
  ```ts
  sendSignal: (idOrOptions: StepOptionsOrId, opts: SendSignalOpts) => Promise<InngestApi.SendSignalResponse>;
  ```
- `node_modules/inngest/components/Inngest.d.ts:178`
  ```ts
  }): Promise<InngestApi.SendSignalResponse>;
  ```

Both files import `InngestApi` from `../api/api.js`, but `./api/api` (or `./api`) is not in `package.json`'s `exports` map, so consumer-side declaration emit can't write a portable specifier for `InngestApi.SendSignalResponse`.

```
$ git log --all --oneline -S "Promise<InngestApi.SendSignalResponse>" --reverse
6b9769ca EXE-266: Update sendSignal's return type (#1479)

$ git tag --contains 6b9769ca
inngest@4.2.5
inngest@4.2.6
```

Regression commit: **#1479** ("EXE-266: Update sendSignal's return type"), first released in **`inngest@4.2.5`**. The PR's checklist notes "Added unit/integration tests: N/A additional testing not needed for typing change" — the consumer declaration-emit path wasn't covered, which is why it slipped through.

(For context, the existing `pnpm test:composite` does call `inngest.sendSignal(...)` but discards the result with `void`, so TypeScript never has to *emit* `InngestApi.SendSignalResponse` to a portable `.d.ts`. The leak would have been caught if the result were exported, or if the test exercised `step.sendSignal` from inside a `createFunction` handler.)

**Some Workarounds I Tried**

The simplest approach was a single annotation:

```ts
const fn: InngestFunction.Any = inngest.createFunction(/* ... */);
```

However, it turns out this does _not_ work because the annotation back-propagates as the contextual type into the `createFunction` call and collapses the handler generic to `Handler.Any`. Inside the handler, `step.run("id", () => ({ a: 1 }))` then returns `unknown` instead of `{ a: 1 }`, downstream `.filter(i => ...)` / `.map(i => ...)` lose their parameter types, and consumers see a cascade of secondary TS7006 errors. Als note that `as InngestFunction.Any` casting had the same back-propagation behavior (and is forbidden by my strict lint rules anyway).

The only workaround I could find that preserves handler inference was a two-variable indirection at every call site (which was not fun to deal with in a few dozen places across our codebase):

```ts
const _x = inngest.createFunction(opts, handler); // full handler inference
const x: InngestFunction.Any = _x;                // portable outer type
export { x };
```

This is verbose, has to be applied at every `createFunction` call site, and is the kind of thing that probably shouldn't be required of SDK consumers in the first place.

**Simplest fix (as mentioned above)**

A one-line addition to `packages/inngest/src/index.ts`:

```diff
 export * from "@inngest/ai";
 export type { StandardSchemaV1 } from "@standard-schema/spec";
+export type { InngestApi } from "./api/api.ts";
 export { experiment } from "./components/ExperimentStrategies.ts";
```
I made this change locally and installed the package directly from my local inngest-js fork and I can confirm all TypeScript inference errors across our monorepo were resolved. 

Once `InngestApi` is reachable through the public package entry, TypeScript writes `import("inngest").InngestApi.SendSignalResponse` into consumer `.d.ts` files instead of the non-portable relative reference and then the TS2742 error goes away.

Also note that `export type` (rather than `export`) was intentional.  That ensures only the namespace member types (`SendSignalResponse`, etc.) need to be reachable for inference.  The `InngestApi` class itself is internal and isn't used by consumers. `export type` erases at runtime, so there's no bundle-size impact either.

This is the same pattern already used for similar cases. PR #1432 ("Fix composite project type exports", `inngest@4.1.1`) added these missing type re-exports for the same reasons:

- `StepFetch`, `GetFunctionOutputRaw`, `GroupTools`, `GroupToolsDeps`, `Realtime`, `HandlerWithTriggers`, `DurationLike`, `InstantLike`, `ZonedDateTimeLike`, `ExclusiveKeys`, `JsonifyObject`

`InngestApi` was simply missed when #1479 introduced a new public reference to it.

**System info**

- npm package version: `inngest@4.2.5` and `inngest@4.2.6` (current)
- Last working version: `inngest@4.2.4`
- TypeScript: 5.9.3
- Node: 24.15.0
- Framework: framework-agnostic (reproduces with the bare SDK)
- Platform: any (this is a build-time TypeScript declaration emit issue, unrelated to runtime)

### LINKED ISSUE #1514: [BUG] Missing types from main entry point
Several types used in the public API surface are not re-exported from the main package entry (`src/index.ts`):

- **`MaybePromise`** — used in return types of `dependencyInjectionMiddleware`, `Inngest` class methods, and `createFunction`
- **`GroupExperiment`** — used in `GroupTools.experiment`
- **`ParallelOptions`** — used in `GroupTools.parallel`

This causes **TS2883** errors during declaration emit when consumers use TypeScript's composite projects (`tsgo` or `tsc` with `declaration: true`), because these types cannot be portably referenced from their internal module paths (which are not in the package `exports` map).

## Error reproduced

```
error TS2883: The inferred type of 'inngest' cannot be named without a reference to
'MaybePromise' from 'inngest/helpers/types'. This is likely not portable.
A type annotation is necessary.
```

**Expected behavior**
I expect exposed types to be portable. 

- OS: Mac
- npm package Version 4.2.6
- Framework Tanstack Start
- Platform Google cloud run

Related PR: #1439

brone1323:
## Problem

Two open issues report `TS2742`/`TS2883` errors for consumers using TypeScript `composite` projects or `declaration: true` emit:

- **#1510** — `step.sendSignal()` return type leaks `InngestApi.SendSignalResponse` from the internal `./api/api` module path, which is not in the package `exports` map. TypeScript can't generate a portable declaration reference and throws TS2742 on any function that uses `step`.
- **#1514** — `MaybePromise`, `GroupExperiment`, and `ParallelOptions` appear in public API signatures but are not re-exported, causing TS2883 during declaration emit.

## Fix

Add four `export type` statements to `packages/inngest/src/index.ts`:

```ts
export type { InngestApi } from "./api/api";          // fixes #1510
export type { GroupExperiment, ParallelOptions } from "./components/InngestGroupTools";  // fixes #1514
export type { MaybePromise } from "./helpers/types";   // fixes #1514
```

All are `export type` — zero runtime/bundle impact. The pattern follows the approach used in PR #1432.

## Checklist

- [x] Types-only change, no runtime diff
- [x] Follows existing export style in `src/index.ts`
- [x] Closes #1510, #1514

---

> [!NOTE]
> Adds four `export type` re-exports (`InngestApi`, `GroupExperiment`, `ParallelOptions`, `MaybePromise`) to the main package entry point to fix TS2742/TS2883 errors in consumer TypeScript projects using `composite` or `declaration: true` settings.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 88baf85474b83b599c834d8c0b92b91139710a5e.</sup>

### DIFF
--- a/packages/inngest/src/index.ts
+++ b/packages/inngest/src/index.ts
@@ -39,6 +39,7 @@
 
 export * from "@inngest/ai";
 export type { StandardSchemaV1 } from "@standard-schema/spec";
+export type { InngestApi } from "./api/api";
 export { experiment } from "./components/ExperimentStrategies.ts";
 export type { StepFetch } from "./components/Fetch";
 export { fetch } from "./components/Fetch.ts";
@@ -56,8 +57,10 @@ export type { InngestFunction } from "./components/InngestFunction";
 export type { InngestFunctionReference } from "./components/InngestFunctionReference";
 export { referenceFunction } from "./components/InngestFunctionReference.ts";
 export type {
+  GroupExperiment,
   GroupTools,
   GroupToolsDeps,
+  ParallelOptions,
 } from "./components/InngestGroupTools";
 export { group, step } from "./components/InngestStepTools.ts";
 export { Middleware } from "./components/middleware/index.ts";
@@ -91,6 +94,7 @@ export type {
 export type {
   ExclusiveKeys,
   IsStringLiteral,
+  MaybePromise,
   SendEventPayload,
   StrictUnion,
   StrictUnionHelper,


Linell:
Thanks for taking a swing at this! It ended up fixed by #1674 and #1439, both in inngest@4.16.0, so I'm going to close this one out.

PR #1521, opened 2026-05-12
Phase 1 Scoring

jakobevangelista:
## Summary

Scoring phase 1

OSS PR: https://github.com/inngest/inngest/pull/4170

## Checklist

- [x] Added a [docs PR](https://github.com/inngest/website/pull/1644) that references this PR

---

> [!NOTE]
> Adds Phase 1 scoring primitives: `inngest.score()` for raw API-level score writes and `step.score()` gated behind `scoreMiddleware()`. Core additions include `InngestScore.ts` with validation/normalization, `canUseCurrentStepBatch` in `InngestMetadata.ts` to route writes through the batch opcode when targeting the current step, `userlandId` on `executingStep` in ALS for userland step ID matching, and a comprehensive test suite.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit ede9a8f17106e9160dc4b9aeb8c0b6804112eab4.</sup>


packages/inngest/src/experimental.ts
```diff
@@ -12,3 +12,10 @@ export { extendedTracesMiddleware } from "./components/execution/otel/middleware
 export { PublicInngestSpanProcessor as InngestSpanProcessor } from "./components/execution/otel/processor.ts";
 // Step Metadata
 export { metadataMiddleware } from "./components/InngestMetadata.ts";
+// Scoring
+export type {
```
amh4r: Is there a use case for these exports?

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -419,6 +425,24 @@ export const createStepTools = <
...
   };
 
+  const createStepScoreWrapper: ScoreStepTool = async (
+    memoizationId,
+    options,
+  ) => {
+    if (!client["experimentalScoreEnabled"]) {
+      throw new Error(
+        'step.score() is experimental. Enable it by adding scoreMiddleware() from "inngest/experimental" to your client middleware.',
+      );
```
amh4r: Should we switch to `NonRetriableError`? If someone is using `step.score()` but `experimentalScoreEnabled` is disabled then retries we keep hitting the same error

packages/inngest/src/components/InngestScore.ts
```diff
@@ -0,0 +1,167 @@
...
+};
+
+export type ScoreStepTool = (
+  memoizationId: string,
+  options: StepScoreOptions,
+) => Promise<void>;
+
+export const scoreSymbol = Symbol.for("inngest.step.score");
+
+function validateScoreFields(
```
amh4r, line 78: This is tough to read. What are your thoughts on something like this:

```ts
function validateIdField({
  value,
  field,
  required,
}: {
  value: unknown;
  field: string;
  required: boolean;
}): void {
  if (!required && value === undefined) {
    return;
  }

  const isValidString = typeof value === "string" && value.trim().length > 0;
  if (!isValidString) {
    throw new Error(`${field} must be a non-empty string`);
  }
}

function validateScoreFields(
  options: unknown,
  requiredTargetIds: readonly ("runId" | "stepId")[],
): asserts options is {
  runId?: unknown;
  stepId?: unknown;
  name: unknown;
  value: unknown;
} {
  if (!isRecord(options)) {
    throw new Error("score options must be an object");
  }

  for (const field of ["runId", "stepId"] as const) {
    validateIdField({
      value: options[field],
      field,
      required: requiredTargetIds.includes(field),
    });
  }

  if (typeof options.name !== "string" || !scoreNameRegex.test(options.name)) {
    throw new Error(
      `invalid score name "${String(options.name)}"; must match ${scoreNameRegex.source}`,
    );
  }

  if (typeof options.value !== "boolean" && !isFiniteNumber(options.value)) {
    throw new Error("score value must be a finite number or boolean");
  }
}
```

With this in `helpers/types.ts`:
```ts
export function isFiniteNumber(value: unknown): value is number {
  return typeof value === "number" && Number.isFinite(value);
}
```

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,15 +278,65 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+    return true;
+  }
+
+  if (config.stepId === executingStep.id) {
+    return true;
+  }
+
+  if (config.stepId !== executingStep.userlandId) {
+    return false;
+  }
```
amh4r: I'm confused here. Is `config.stepId` the hashed step ID or userland ID?

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,15 +278,65 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+): BuilderConfig {
+  const executingStep = ctx?.execution?.executingStep;
+  if (
+    !executingStep ||
+    config.spanId ||
+    !executingStep.userlandId ||
+    config.stepId !== executingStep.userlandId
+  ) {
+    return config;
+  }
```
amh4r: Can we clarify this condition more? Is each condition representing a distinct reason to return the config as-is? Or does their combination represent a higher-level state?

If each is a distinct reason, then it'd be more understandable with multiple if conditions returning the config as-is, each with their own comment explaining why

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,15 +278,65 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+
+  const currentIndex = executingStep.userlandIndex ?? 0;
+  return {
+    ...config,
+    ...(currentIndex > 0 ? { stepIndex: currentIndex } : {}),
+  };
+}
```
amh4r: Is this actually another case where we want to return the config as-is? If `currentIndex <= 0` then we're spreading an empty objects, which means we're returning config as-is

packages/inngest/src/experimental.ts
jakobevangelista: https://github.com/inngest/inngest-js/pull/1521/commits/849f349c25cedb0ecfe7bfc4e361b951cc93e56d

packages/inngest/src/components/InngestStepTools.ts
jakobevangelista: https://github.com/inngest/inngest-js/pull/1521/commits/9124309d2ba9a0a2a6c01379a2924cbb83e2b5ed

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,15 +278,83 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+
+  if (targetStepId === undefined) {
+    return true;
+  }
+
+  if (targetStepId === null) {
+    return true;
+  }
+
+  // Callers may target the runtime step ID or original userland ID.
```
amh4r: Is "runtime step ID" the hashed step ID? And Is "caller" the user? If it is the hashed step ID, what's the use case for targeting that? It seems like users should only be able to target userland IDs

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,15 +278,83 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+  return (
+    targetStepId === currentStepId || targetStepId === currentUserlandStepId
+  );
+}
+
+/**
+ * REST fallback needs step_index when duplicate userland IDs target this step;
```
amh4r: What's this about targeting duplicate userland IDs? I thought we weren't supporting that with scoring

packages/inngest/src/components/InngestStepTools.ts
jakobevangelista: this made me wonder if we should do the same for deterministic score validation errors too, like invalid score names, NaN / infinity, or empty ids whcih also won’t be fixed by a retry

kept this change on the missing middleware/config case for now, but curious if you think validation errors should be `NonRetriableError` too

packages/inngest/src/components/InngestScore.ts
jakobevangelista: https://github.com/inngest/inngest-js/pull/1521/commits/207d212a8e69a7122dc22dbca3a124bce0b030a9

packages/inngest/src/components/InngestMetadata.ts
jakobevangelista: fixed by not handling duplicate stepids https://github.com/inngest/inngest-js/pull/1521/commits/12cd83db4b13e03ec855e9d569e47026b818fcf9

packages/inngest/src/components/InngestMetadata.ts
jakobevangelista: also fixed with https://github.com/inngest/inngest-js/pull/1521/commits/12cd83db4b13e03ec855e9d569e47026b818fcf9

packages/inngest/src/components/InngestMetadata.ts
jakobevangelista: https://github.com/inngest/inngest-js/commit/207d212a8e69a7122dc22dbca3a124bce0b030a9

packages/inngest/src/components/InngestMetadata.ts
jakobevangelista: https://github.com/inngest/inngest-js/pull/1521/commits/12cd83db4b13e03ec855e9d569e47026b818fcf9

packages/inngest/src/components/InngestMetadata.ts
jakobevangelista: now we only batch explicit step targets when they match the current userland `stepId`

if no `stepId` is provided, that’s run scoped and still batches through the current opcode when possible

packages/inngest/src/components/InngestStepTools.ts
amh4r: `NonRetriableError` might make sense in other places too. Really any time there's an error that a retry won't fix. But the other places aren't a blocker for this PR

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,7 +278,34 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
   return "step";
 }
 
-async function performOp(
+/**
+ * Keep in-step metadata durable by batching only targets that resolve to this
+ * step opcode; other targets must go through REST.
+ */
+function canUseCurrentStepBatch(
```
amh4r: This function is confusing. Its purpose isn't entirely clear and it has redundant checks that its caller makes. For example, we're already checking `config.spanId` [here](https://github.com/inngest/inngest-js/pull/1521/changes#diff-8c9b2dd9ae88d3d28616e449d1b1a75cc61edde5027fbaabfb99c504abf17301R334)

jakobevangelista: cleared it up with https://github.com/inngest/inngest-js/pull/1521/commits/ec4ddd33e01d310c3ee4a1f8edd6002484779b52

packages/inngest/src/components/InngestScore.ts
```diff
@@ -0,0 +1,173 @@
...
+  name: string;
+  value: ScoreValue;
+};
+
+export type StepScoreOptions = {
+  runId?: string;
+  stepId?: string;
+  name: string;
+  value: ScoreValue;
+};
```
AndyInternet: `SendScoreOptions` and `StepScoreOptions` are identical. They might diverge in the future, but until then, we should use one type.

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -36,7 +37,7 @@ export type MetadataUpdate = {
 
 export type MetadataValues = Record<string, unknown>;
 
-interface BuilderConfig {
+export interface BuilderConfig {
```
AndyInternet, line 45: nit: `BuilderConfig` and `performOp` got promoted from internal to export here so `InngestScore.ts` can call them. That's fine, but anything reachable via `inngest/components/InngestMetadata.ts` is now a soft API contract for deep imports. We should add an `@internal` JSDoc comment to them so it's clear that it's an internal use structure.

packages/inngest/src/components/InngestMetadata.ts
```diff
@@ -277,7 +278,30 @@ function getBatchScope(config: BuilderConfig): MetadataScope {
...
+  }
+
+  if (targetStepId === null) {
+    return true;
+  }
+
+  return targetStepId === currentUserlandStepId;
+}
+
+export async function performOp(
```
AndyInternet, line 321: See above nit

AndyInternet (approved):
Let's go!

PR #1547, opened 2026-06-01
Sessions

jakobevangelista:
## Summary

sessions let you group related runs under shared identifiers (eg. all runs in a conversation). this adds the SDK surface: an optional `sessions` field on event payloads

```ts
await inngest.send({
  name: "chat/message.sent",
  data: { ... },
  meta: { 
    sessions: {
      conversation_id: "conv_123"
    } 
  },
});
```

a session is a key (`conversation_id`) + an ID (`conv_123`). values can be string/number/boolean and are normalized to strings client-side; received events are always typed `Record<string, string>`

surfaces, one commit each:
- `send` / `step.sendEvent` — attach sessions to any event
- `step.invoke` — explicit `sessions` option
- `step.waitForEvent` — sessions never participate in matching, the matched event's sessions are just piped through on the result
- lifecycle events (`function.failed`/`finished`/`cancelled`) carry the triggering event's sessions, so `onFailure` runs stay in the same session — typed on the payloads

server side: inngest/inngest#4292, cloud: inngest/monorepo#7355

convo [here](https://inngest.slack.com/archives/C05QSAYPEG0/p1781721674384799?thread_ts=1781717694.974469&cid=C05QSAYPEG0) about the `meta` changes

## Checklist

## Related
- EXE-1902


packages/inngest/src/helpers/sessions.ts
jakobevangelista, line 1: validation is intentionally minimal

only things that can't be delegated to the server like non-finite numbers and type/empty checks fail fast at the call site

count and length limits are enforced on the server

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -1161,6 +1177,15 @@ type InvocationTargetOpts<TFunction extends InvokeTargetFunctionDefinition> = {
...
   InvocationTargetOpts<TFunction> &
     Omit<TriggerEventFromFunction<TFunction>, "id"> & {
+      /**
+       * Session metadata used to group the invoked run with other runs.
+       *
+       * Sessions are not inherited from the calling run; pass them explicitly
+       * to group the invoked run. Values are normalized to strings before
+       * sending, as with `inngest.send()`.
+       */
+      sessions?: EventSessions;
```
jakobevangelista: explicit-only was deliberate over inheriting the calling run's sessions

packages/inngest/src/components/InngestStepTools.ts
```diff
@@ -1259,7 +1284,14 @@ type WaitForEventResult<TOpts> =
...
     >;
   }
-    ? { name: TName; data: TData; id: string; ts: number; v?: string } | null
+    ? {
+        name: TName;
+        data: TData;
+        id: string;
+        ts: number;
+        v?: string;
+        sessions?: Record<string, string>;
```
jakobevangelista: sessions are not matched

waits match on event name + match/if expression only, and the matched event's sessions are piped through on the result, already normalized to strings

Linell (approved):

PR #1550, opened 2026-06-03
EXE-1547: Long Running Steps split into two OpCodes

scottnuma:
## Summary

- We would like to display in-progress steps. 
- In this PR, for checkpointing steps that runs for < 1s, we will emit a StepPlanned Opcode after 1s that communicates the step names and starting time. Users can then see their step in progress in the dashboard. Upon the completion of the step, we emit the StepRun OpCode with the rest of the information. The step will transition to completed in the dashboard.
- In the trailing StepRun OpCode, we specifically omit the `startedAt` and `completedAt` attributes to prevent overwriting those attributes from the leading StepPlanned OpCode.
- https://github.com/inngest/inngest/pull/4313 must merge and deploy before this is deployed

I diagram the control flow here

image

Full diagram

image

## Checklist

## Related

- INN-


packages/inngest/src/api/api.ts
```diff
@@ -601,6 +601,40 @@ export class InngestApi {
...
+    fnId: string;
+    queueItemId: string;
+    steps: OutgoingOp[];
+  }): Promise<{ ok: boolean; status?: number }> {
+    const body = JSON.stringify({
+      run_id: args.runId,
+      fn_id: args.fnId,
+      qi_id: args.queueItemId,
+      steps: args.steps,
+      ts: new Date().valueOf(),
```
rhino1998, line 623: If you use the same timestamp as the one that becomes the trailing step's `timing` field then the backend handling can be much simpler. (Actually, why not reuse the `timing` field here w/ a zero duration)

so emit planned with queued/started = t_0 and then emit run w/ queued/started=t_0 and ended=t_1 (we actually emit start&duration rather than start&end, but same idea)

packages/inngest/src/components/execution/engine.ts
```diff
@@ -1731,10 +1760,164 @@ class InngestExecutionEngine
...
+    }
+
+    if (state.kind === "pending") {
+      // Step finished before the timer fired. Emit the usual
+      // StepRun with both the actual start and end time
+      return op;
+    }
+
+    if (state.kind === "failed") {
+      // Leading edge never made it — push a `StepPlanned` ahead of the
```
rhino1998, line 1893: If we do the above, we can skip re-sending the planned step span because the final step span would have all the info we need.

packages/inngest/src/components/execution/engine.ts
```diff
@@ -1731,10 +1760,164 @@ class InngestExecutionEngine
...
+    // completion arm of an earlier `StepPlanned`, so it should write only
+    // `EndedAt`/`DynamicStatus` (not `QueuedAt`/`StartedAt`) — otherwise
+    // the maxMap-by-start_time aggregation would clobber the leading's
+    // queued/started timestamps and make the step appear to take 0ms.
+    const endNs =
+      (op.timing?.a ?? record.startMs * 1_000_000) + (op.timing?.b ?? 0);
+    const collapsedTrailing: OutgoingOp = {
+      ...op,
+      opts: { ...(op.opts ?? {}), _paired_trailing: true },
+      timing: { a: endNs, b: 0 },
```
rhino1998, line 1917: We could send the original start time here with the real duration rather than a 0-duration timing

scottnuma:
Two paths forward that reduce the complexity of this feature:
1 - send the StepPlanned with a -1ms start to make the StepRun's completed status win
2 - send the StepRunCompleted span's timestamp as a 0 duration span (t_end, t_end) to make the StpeRun's completed status win without changing the t_start and t_end presented to users

1 looks like it can have a smaller footprint, going to look into pursuing that one else use 2

Either way, we can reduce logic here by consistently sending the same StepRun opcode and reduce the backend changes.

PR #1552, opened 2026-06-04
internal: VS Code biome config

amh4r:
- Config VS Code to use biome
- Add lint script to test harness package


packages/test-harness/src/testApp.ts
```diff
@@ -1,3 +1,5 @@
+// biome-ignore-all lint/suspicious/noExplicitAny: it's cool
```
Linell, line 1: 😎

Linell (approved):

jpwilliams (approved):

PR #1553, opened 2026-06-04
feat: Emit a StepPlanned opcode for long running steps (EXE-1547)

scottnuma:
## Summary

- We would like to display in-progress steps
- In this PR, for steps that take > 1s, we emit a StepPlanned opcode that enables the dashboard to display that a named step is in progress
- On our StepPlanned opcode, we set the started at to be the actual start time - 1ms. This means when we aggregate the StepPlanned opcode with the completion opcode, the actual start time and completed status will win. This also means that the start time displayed while the step is in progress is 1ms earlier for this long step.

## Checklist

### DIFF
--- a/.changeset/long-steps-report.md
+++ b/.changeset/long-steps-report.md
@@ -1,0 +1,5 @@
+---
+"inngest": feature
+---
+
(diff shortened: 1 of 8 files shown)


scottnuma:
I'm investigating some strange behavior for long running steps with multiple steps. In this example, we have a long running step that emits the StepPlanned OpCode, but then fails. We then see a phantom flaky step at the bottom of this graph. I think because of the difference in parentage between the StepPlanned OpCode and the error opcode. 

Screenshot 2026-06-04 at 10 02 09 AM

PR #1562, opened 2026-06-08
EXE-1938: Parse basic AI Metadata from OTel Attributes

scottnuma:
## Summary

- Parse OTel attributes emitted by common LLM OTel instrumentation libraries and SDKs (Otel, Open Inference, Vercel). We begin with two simple attributes: model and input tokens
- We include a simple method to aggregate
- We include a large test data set from OSS. This is makes for the bulk of the 10,000 lines added. If that many lines added is problematic, we could point to them in OSS or add the test apps that generated them instead.
- These functions are currently unused, but we'll put them to work in future PRs.

## Checklist


packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -0,0 +1,144 @@
...
+ * AI metadata extracted from a span's attributes.
+ *
+ * A field is only present when a matching attribute was found on the span;
+ * absent fields are omitted rather than zero-valued.
+ */
+export interface AIMetadata {
+  /** The requested model, e.g. `gpt-4.1-nano`. */
+  model?: string;
+  /** The number of input (prompt) tokens consumed by the request. */
+  inputTokens?: number;
```
rhino1998, line 13: ~Did we need any of the other fields that the Go version of this extracts? (A bit out of the loop on what was decided)~ I need to read PR descriptions before diffs

rhino1998 (approved):
Seems good

PR #1569, opened 2026-06-11
fix(middleware): map step.ai.wrap planned op to ai.wrap step type

### LINKED ISSUE #1555: step.ai.wrap emits "Unknown step type" warning in middleware step type mapper
## Summary

`step.ai.wrap()` appears to emit a planned step with `opts.type === "step.ai.wrap"`, but the middleware step-type mapper only recognizes `"step.ai.wrap"` under the `AIGateway` opcode. As a result, the SDK logs an `Unknown step type` warning for normal `step.ai.wrap()` usage.

This does not appear to block function execution or third-party OpenTelemetry export in my local reproduction, but it pollutes logs and may label middleware step metadata as `unknown`.

## Versions checked

- Observed locally with `inngest@4.4.0`
- `inngest@4.5.0` appears to have the same mapper shape in the published npm tarball

## Minimal repro

```ts
  { id: "repro" },
  { event: "test/repro" },
  async ({ step }) => {
    return step.ai.wrap("classify", async () => "ok");
  }
);
```

## Actual behavior

The function runs, but the SDK logs:

```txt
Unknown step type
{ op: 'StepPlanned', type: 'step.ai.wrap' }
```

## Expected behavior

No warning for normal `step.ai.wrap()` usage. Either:

- `StepPlanned` with `opts.type === "step.ai.wrap"` maps to `ai.wrap`, or
- `step.ai.wrap()` emits an `AIGateway` op if that is the intended opcode for AI steps.

## Source-level notes

In the published package:

- `components/InngestStepTools.js` defines `wrap` via `createStepRun("step.ai.wrap")`, which produces a `StepPlanned` op.
- `components/middleware/utils.js` maps `opts.type === "step.ai.wrap"` only inside the `StepOpCode.AiGateway` branch.
- The `StepPlanned` branch maps other planned step subtypes such as `step.sendEvent`, `step.realtime.publish`, and `group.experiment`, but not `step.ai.wrap`.

That combination seems to make `step.ai.wrap()` fall through to `unknown` in the middleware step-type mapper.

tsushanth:
## Summary

`step.ai.wrap()` is implemented via `createStepRun("step.ai.wrap")` ([InngestStepTools.ts:764](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/InngestStepTools.ts#L764)), which emits a `StepPlanned` op with `opts.type === \"step.ai.wrap\"`. The `stepTypeFromOpCode` mapper at [middleware/utils.ts:87-126](https://github.com/inngest/inngest-js/blob/main/packages/inngest/src/components/middleware/utils.ts#L87-L126) only recognised `step.ai.wrap` inside the `AiGateway` branch — which is not the opcode actually emitted at runtime — so normal `step.ai.wrap()` usage fell through to `\"unknown\"` and logged `\"Unknown step type\"` on every call.

This PR adds the missing case to the `StepPlanned` branch so middleware step-type metadata is correct and the warning no longer fires for valid usage. The existing `AiGateway` mapping is preserved (it's harmless and already test-covered, so removing it could be a separate cleanup if it's known-unreachable).

## What was wrong

In `packages/inngest/src/components/InngestStepTools.ts`:
- `step.ai.infer` → `StepOpCode.AiGateway` with `opts.type = \"step.ai.infer\"` (line 739)
- `step.ai.wrap` → `createStepRun(\"step.ai.wrap\")` → `StepOpCode.StepPlanned` with `opts.type = \"step.ai.wrap\"` (line 363, 764)

The mapper only handled `\"step.ai.wrap\"` under `StepOpCode.AiGateway`, so the actual `StepPlanned + \"step.ai.wrap\"` path returned `\"unknown\"`.

## Change

```diff
 } else if (op === StepOpCode.StepPlanned) {
   if (opts?.type === undefined) {
     return \"run\";
   }
   if (opts?.type === \"step.sendEvent\") {
     return \"sendEvent\";
   }
   if (opts?.type === \"step.realtime.publish\") {
     return \"realtime.publish\";
   }
   if (opts?.type === \"group.experiment\") {
     return \"group.experiment\";
   }
+  if (opts?.type === \"step.ai.wrap\") {
+    return \"ai.wrap\";
+  }
 }
```

Plus a unit test mirroring the existing `AiGateway + step.ai.wrap` case for the `StepPlanned` branch, and a `patch` changeset.

## Test plan

- [x] `pnpm vitest run src/components/middleware/utils.test.ts` — 44/44 pass, including the new `StepPlanned + step.ai.wrap` case
- [x] `pnpm lint` (Biome) — clean
- [x] Existing `AiGateway + step.ai.wrap` test still passes (mapping preserved)

Closes #1555

### DIFF
--- a/.changeset/fix-step-ai-wrap-unknown-type.md
+++ b/.changeset/fix-step-ai-wrap-unknown-type.md
@@ -1,0 +1,14 @@
+---
+"inngest": patch
+---
+
+fix(middleware): map `step.ai.wrap` planned op to `ai.wrap` step type
+
+`step.ai.wrap()` emits a `StepPlanned` op with `opts.type === "step.ai.wrap"`,
+but `stepTypeFromOpCode` only recognised `step.ai.wrap` under the
+`AiGateway` opcode branch. Normal `step.ai.wrap()` usage therefore fell
+through to `unknown` and logged an "Unknown step type" warning.
+
+Adds the missing `step.ai.wrap` case to the `StepPlanned` branch so
+middleware step-type metadata is correct and the warning no longer fires.
+The existing `AiGateway` mapping is preserved.
--- a/packages/inngest/src/components/middleware/utils.test.ts
+++ b/packages/inngest/src/components/middleware/utils.test.ts
@@ -43,6 +43,16 @@ describe("stepTypeFromOpCode", () => {
     ).toBe("realtime.publish");
   });
 
+  test("StepPlanned with type 'step.ai.wrap' returns 'ai.wrap'", () => {
+    expect(
+      stepTypeFromOpCode(
+        StepOpCode.StepPlanned,
+        { type: "step.ai.wrap" },
+        logger,
+      ),
+    ).toBe("ai.wrap");
+  });
+
   test("StepPlanned with unknown type returns 'unknown'", () => {
     expect(
       stepTypeFromOpCode(
--- a/packages/inngest/src/components/middleware/utils.ts
+++ b/packages/inngest/src/components/middleware/utils.ts
@@ -113,6 +113,9 @@ export function stepTypeFromOpCode(
     if (opts?.type === "group.experiment") {
       return "group.experiment";
     }
+    if (opts?.type === "step.ai.wrap") {
+      return "ai.wrap";
+    }
   } else if (op === StepOpCode.Sleep) {
     return "sleep";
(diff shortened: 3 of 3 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #1571, opened 2026-06-11
Don't double count Vercel SDK AI spans (EXE-1837)

scottnuma:
## Summary

- The Vercel SDK AI emits multiple spans per LLM call
- The parent is a roll up that has aggregated data about the LLM call(s)
- We perform our own aggregation, so parsing both leads to double counting. Let's ignore the roll up span until we want to extract data only found there and not in the children.

## Checklist


rhino1998 (approved):

scottnuma:
Closing in favor approach used in https://github.com/inngest/inngest/pull/4420

PR #1573, opened 2026-06-12
Create `@inngest/otel` package

amh4r:
## Summary

Create `@inngest/otel` package. This package is focused on being an optional turnkey solution for setting up OpenTelemetry for our SDK.

## Motivation

Move OpenTelemetry provider creation and instrumentation out of our main SDK package (`inngest`). Our primary motivation was reducing dependencies in our main `inngest` package. The benefits of that are:
- More freedom to add/update dependencies without affecting our SDK.
- Reduce the chance of our users getting spurious security warnings for unused transitive dependencies.


jpwilliams (approved):

Linell (approved):

packages/otel/docs/support.md
```diff
@@ -0,0 +1,65 @@
...
+- ensures there's a process-global OpenTelemetry tracer provider
+- installs an async hooks context manager when this package creates the provider
+- doesn't install Inngest span processors or exporters
+- no-ops when loaded more than once in the same process
+
+If an OpenTelemetry tracer provider already exists, this package leaves it in
+place. If no provider exists, this package creates a basic provider.
+
+The main `inngest` package is responsible for adding Inngest span processors.
+User-managed OpenTelemetry setups can add their own processors and exporters.
```
scottnuma: I think in this version, users can't add their own processors. Seems good for early versions, but maybe also worth calling this out in these docs to make sure we don't forget and mislead users later.

packages/otel/docs/support.md
```diff
@@ -0,0 +1,65 @@
...
+
+Instrumentation should be loaded before application imports. Libraries imported
```
scottnuma:
```suggestion
Instrumentation should be loaded before importing this package. Libraries imported
```

I think this makes it a little more explicit what they need to import before

packages/otel/docs/support.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Not supported
+
+This package doesn't currently support:
+
+- `node --require @inngest/otel/node`
+- `require("@inngest/otel/node")`
+- `import "@inngest/otel"`
+- browser, edge, worker, or non-Node runtimes
+- configuration of exporters, sampling, resources, or instrumentation lists
```
scottnuma: maybe processors in here too for now

packages/otel/docs/support.md
```diff
@@ -0,0 +1,65 @@
...
+
+The package doesn't expose a root `@inngest/otel` export.
+
+## Runtime behavior
+
+When `@inngest/otel/node` is preloaded, it:
+
+- registers the OpenTelemetry ESM instrumentation hook
+- registers supported Node auto-instrumentations
+- registers Anthropic instrumentation
```
scottnuma: Given that we call out Anthropic, and that our user case is AI, I think we should call out that OpenAI is also instrumented - just inside of Node auto-instrumentations.

I definitely didn't realize this until I looked closely at extended traces.

packages/otel/src/instrument.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    context.setGlobalContextManager(
+      new AsyncLocalStorageContextManager().enable(),
+    );
+  } catch (e) {
+    const err = toError(e);
+    debug("failed to initialize provider:", err);
+    return toError(err);
+  }
```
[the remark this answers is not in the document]
scottnuma: in case you follow the advice, make sure to watch out for the ProxyProvider that getTracerProvider will return with newer versions.

packages/otel/src/instrument.ts
```diff
@@ -0,0 +1,102 @@
...
+    isTraceInstrumentationStarted = false;
+    debug("unable to register trace instrumentations", instrumented);
+    return;
+  }
+
+  const provider = ensureTraceProvider();
+  if (provider instanceof Error) {
+    isTraceInstrumentationStarted = false;
+    debug("unable to initialize provider", provider);
+  }
```
scottnuma, line 44: It's slightly unexpected to me that we create the provider inside of this instrumentTracing fn - might be worth pulling out to be top level, or calling it out in the docstring

scottnuma (approved):
nice! nothing blocking - many of these changes would probably have a better home in a differnt pr

packages/otel/docs/support.md
amh4r: This saying that instrumentation should be done before the application imports anything

packages/otel/src/instrument.ts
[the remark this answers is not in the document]
amh4r: Not gonna implement these suggestions at this time

packages/otel/docs/support.md
amh4r: ✅

packages/otel/docs/support.md
amh4r: ✅

packages/otel/docs/support.md
amh4r: ✅

packages/otel/src/instrument.ts
amh4r: I added a function comment to help clarify

PR #1578, opened 2026-06-15
Bring JS AI Metadata fields to closer to parity with OSS (EXE-1938)

scottnuma:
## Summary

- Add 6 or so new fields to AI Metadata that we parsed in OSS but did not previously in JS

## Checklist

### DIFF
--- a/packages/inngest/src/components/execution/otel/aiExtractor.ts
+++ b/packages/inngest/src/components/execution/otel/aiExtractor.ts
@@ -9,6 +9,11 @@ import type { Attributes, AttributeValue } from "@opentelemetry/api";
 export interface AIMetadata {
   /** The requested model, e.g. `gpt-4.1-nano`. */
   model?: string;
+  /**
+   * The model that actually served the request. This is often a dated snapshot
+   * of the requested {@link AIMetadata.model} (e.g. `gpt-4.1-nano-2025-04-14`).
+   */
+  responseModel?: string;
   /** The number of input (prompt) tokens consumed by the request. */
   inputTokens?: number;
   /** The number of output (completion) tokens produced by the response. */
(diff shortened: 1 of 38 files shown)


packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -193,26 +325,66 @@ export const extractAIMetadataFromAttributes = (
     metadata.model = model;
   }
 
-  const inputTokens = candidates.inputTokens?.value;
+  const responseModel = candidates.responseModel?.value;
+  if (typeof responseModel === "string" && responseModel !== "") {
```
amh4r:
```suggestion
  if (responseModel) {
```

Nit: you can check truthiness instead
```
> Boolean("")
false
> Boolean(undefined)
false
```

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -63,23 +102,28 @@ const expandLangfuseUsageDetails = (
...
   }
 
-  const input = (counts as Record<string, unknown>).input;
-  if (typeof input !== "number") {
-    return {};
+  const counts = parsed as LangfuseUsageDetails;
+
+  const out: Record<string, AttributeValue> = {};
+  for (const key of langfuseUsageKeys) {
+    const count = counts[key];
```
amh4r:
```suggestion
  if (!isRecord(parsed)) {
    return {};
  }

  const out: Record<string, AttributeValue> = {};
  for (const key of langfuseUsageKeys) {
    const count = parsed[key];
```

We can use `isRecord` to avoid the type cast

packages/inngest/src/components/execution/otel/aiExtractor.ts
scottnuma: Switched in https://github.com/inngest/inngest-js/pull/1578/changes/3cc834ff2eb84d60262970228b52e2dbe02550f5

packages/inngest/src/components/execution/otel/aiExtractor.ts
scottnuma: Switched in https://github.com/inngest/inngest-js/pull/1578/changes/3cc834ff2eb84d60262970228b52e2dbe02550f5

scottnuma:
The last commit also adds to CLAUDE.md to prevent this from repeating again on the Claude side.

PR #1587, opened 2026-06-17
fix: correctly handle timeouts w/custom fetch 

Linell:
## Summary

Fix an issue, noticed by a customer using a customer `fetch()` with a timeout to prevent runaway Vercel function timeouts, that would cause us to throw errors that are actually gracefully handled.

## Checklist


packages/inngest/src/helpers/env.ts
```diff
@@ -413,13 +413,24 @@ export const getFetch = (
...
-          !err.message?.startsWith("fetch failed")
-        ) {
+        const name =
+          typeof err === "object" && err !== null && "name" in err
+            ? err.name
+            : undefined;
+
+        const isExpectedFetchError =
+          name === "AbortError" ||
+          name === "TimeoutError" ||
```
amh4r, line 430: Would `instanceof` checks work?

packages/inngest/src/helpers/env.ts
```diff
@@ -413,13 +413,24 @@ export const getFetch = (
...
+          (err instanceof Error && err.message?.startsWith("fetch failed"));
```
amh4r, line 431: What's the case where we hit this instead of `AbortError` or `TimeoutError`?

packages/inngest/src/helpers/env.ts
Linell: `AbortError` and `TimeoutError` are both just `DomException`s, which feels too broad. We've seen [other cases where `instanceof` can be unreliable](https://github.com/inngest/inngest-js/pull/1242/changes) too, so I think the name check is the right path here.

packages/inngest/src/helpers/env.ts
Linell: You'd hit it on an actual transport failure as opposed to the `DOMException` - it's the OG case.

PR #1594, opened 2026-06-18
Parse more OTel GenAI attributes for AI Metadata

scottnuma:
## Summary

- Capture a broader of Open Telemetry GenAI attributes
- For tokens and cost, we sum when there are multiple spans per step. For everything else, last span wins.
- This removes the previous metadata extraction for Open Inference, Vercel AI SDK-specific, and Langfuse-specific span attributes.

## Checklist


packages/inngest/package.json
```diff
@@ -333,6 +333,7 @@
     "@actions/core": "^1.10.0",
     "@actions/exec": "^1.1.1",
     "@arethetypeswrong/cli": "^0.18.2",
+    "@arizeai/openinference-semantic-conventions": "^2.5.0",
```
amh4r: Should be in `dependencies` instead?

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -1,257 +1,170 @@
...
-    field: "inputTokens",
-    convention: Convention.Vercel,
-  },
-  // Embeddings spans emit only a single `ai.usage.tokens` count (no
-  // input/output split); map it to inputTokens to match the semconv embeddings
-  // case.
-  "ai.usage.tokens": { field: "inputTokens", convention: Convention.Vercel },
+const isOpenInferenceKey = (key: string): boolean => {
+  const dot = key.indexOf(".");
+  const root = dot === -1 ? key : key.slice(0, dot);
```
amh4r:
```suggestion
  const root = key.split(".")[0];
  return openInferenceRootSet.has(root);
```

scottnuma: wiped out during later commits

packages/inngest/package.json
scottnuma: I chose to only use these in tests to avoid including it in our package. We still do validation with it in our testing.

Pros: lighter package for users 
Cons: we write the string constants ourselves

Can easily flip too!

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -1,256 +1,335 @@
...
+    if (spec.combine === "text") {
+      const text = typeof value === "string" ? value : String(value);
+      if (text !== "") {
+        setField(metadata, spec.field, text);
+      }
+    } else {
+      const num = Number(value);
+      if (!Number.isNaN(num)) {
+        setField(metadata, spec.field, num);
       }
```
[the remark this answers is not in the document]
scottnuma: A 0 value in an attribute is useful data. We will report it. It would be buggy if there were no attribute and we reported a 0, but that's not the case.

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -1,256 +1,335 @@
...
+    }
 
-  if (a.inputTokens !== undefined || b.inputTokens !== undefined) {
-    metadata.inputTokens = (a.inputTokens ?? 0) + (b.inputTokens ?? 0);
+    if (spec.combine === "sum") {
+      const aValue = out[spec.field];
+      setField(
+        out,
+        spec.field,
+        (typeof aValue === "number" ? aValue : 0) + (bValue as number),
```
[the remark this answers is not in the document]
scottnuma: Accepted fix

.changeset/open-inference.md
```diff
@@ -0,0 +1,5 @@
+---
+"inngest": feature
```
amh4r:
```suggestion
"inngest": minor
```

.changeset/open-inference.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Adds support for many Open Inference attributes when parsing OTel spans for metadata. Removes support for the keys of OTel semantic conventions (gen_ai.*), Vercel AI SDK, and Langfuse.
```
amh4r:
```suggestion
Replace OpenTelemetry semantic conventions (`gen_ai.*`) with OpenInference. This only affects experimental AI metadata extraction.
```

Nit: Maybe something that reads a little more changeloggy (since that's where this text will appear).

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -280,29 +243,24 @@ export const aggregate = (a: AIMetadata, b: AIMetadata): AIMetadata => {
   const out: AIMetadata = { ...a };
 
   for (const spec of FIELD_SPECS) {
-    if (spec.merge === "sum") {
+    if (spec.valueType === "number") {
```
scottnuma, line 246: inverting the if statements to do typing on the outside is a little nicer for typing

packages/inngest/src/components/execution/otel/aiExtractor.ts
```diff
@@ -1,68 +1,69 @@
...
- * For each category, we describe how we handle aggregation of multiple
- * AIMetadata.
+ * For each category we describe how multiple AIMetadata values are aggregated
+ * (see {@link aggregate}).
  */
 export interface AIMetadata {
   // Identity & classification (last-write-wins).
-  spanKind?: string;
-  model?: string;
+
```
scottnuma, line 17: There are more fields we could parse, but this steals from [the approach we were using before OpenInference](https://github.com/inngest/inngest-js/pull/1591/changes) as a starting base. We'll add more in a separate PR.

amh4r (approved):

PR #1595, opened 2026-06-18
feature: add AI metadata extraction opt-out

amh4r:
## Summary

Add option for opting out of AI metadata extraction.

### DIFF
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -142,6 +142,13 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
   private readonly _logger: Logger;
 
   /**
+   * Whether this client should collect AI metadata from OpenTelemetry spans.
+   *
+   * @internal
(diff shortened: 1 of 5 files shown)


packages/inngest/src/components/Inngest.ts
```diff
@@ -330,6 +337,7 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
     }
 
     this.id = id;
+    this.aiMetadataEnabled = this.options.aiMetadata !== false;
```
scottnuma, line 340: In case people put junk in this field, like `0`, would it be more robust to condition on `this.options.aiMetadata == true`

scottnuma (approved):

packages/inngest/src/components/Inngest.ts
amh4r: It's opt out, so the only value we need to react to is `false`. If someone puts `0` despite the type error then that's kinda on them 😄

PR #1613, opened 2026-06-29
fix: preserve step.run output types with multiple middleware

VihaanAgarwal:
## Summary

When a client has two or more middleware, `step.run`'s output type can collapse to `{}` if the returned value contains an object with an optional property. Reading any property off it then fails to type-check, even though the value is correct at runtime.

Root cause: each middleware's default `stepOutputTransform` is `Jsonify<In>`. With N middleware those nest as `Jsonify<Jsonify<...>>`. An optional property pushes `Jsonify` down its `UndefinedToOptional` path, and the extra nesting crosses TypeScript's instantiation-depth guard. TS does not raise TS2589 here; it silently resolves the inner type to `JsonifyObject<{}>`. A single middleware stays under the limit, which is why the bug only appears at two or more.

Fix: pass inputs that are already JSON values through the default transform unchanged. `Jsonify` is idempotent on those types so the result type is identical, but the second and later transforms stop re-instantiating `Jsonify`, so the depth stays bounded no matter how many middleware are stacked. Custom `stepOutputTransform`s are unaffected.

Added a type test covering two no-op middleware plus a returned element with an optional property, which fails on `main` and passes with this change.

## Checklist

- [ ] ~~Added a docs PR~~ N/A Type-inference fix with no doc-facing surface

## Related

- inngest/inngest#4483

### DIFF
--- a/.changeset/tame-ducks-cheer.md
+++ b/.changeset/tame-ducks-cheer.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix `step.run` output types collapsing to `{}` when two or more middleware are used and a returned object has an optional property
--- a/packages/inngest/src/components/middleware/middleware.test.ts
+++ b/packages/inngest/src/components/middleware/middleware.test.ts
@@ -33,3 +33,39 @@ test("stepOutputTransform does not affect step.invoke return type", () => {
     },
   );
 });
+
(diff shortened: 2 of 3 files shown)


VihaanAgarwal:
Closing this out, #1618 fixed the same collapse at the Jsonify level which covers more cases than this did. Thanks!

PR #1614, opened 2026-07-01
fix(stream): guard heartbeat and finalize against closed ReadableStream controller (fixes #1475)

### LINKED ISSUE #1475: [BUG] serve({ streaming: true }) crashes Node with uncaught TypeError: Invalid state: Controller is already closed from heartbeat
**Describe the bug**

`createStream` in `src/helpers/stream.ts` (used when `serve({ streaming: true })` is enabled) sets up a `setInterval` heartbeat that calls `controller.enqueue(...)` on a `ReadableStream` controller, and a `finalize` function that asynchronously enqueues + closes inside a `Promise.resolve(data).then(...)`. Neither path is guarded:

- The heartbeat `setInterval` callback does not check whether the controller is still open, and is not wrapped in try/catch.
- The `finalize` `.then(...)` has no `.catch(...)`, and also does not check controller state before enqueueing.

If the downstream consumer closes/cancels the stream before the run finishes — which happens routinely on Vercel Fluid Compute when the proxy resets the socket, or on an Inngest executor timeout — the next heartbeat tick calls `controller.enqueue` on a closed controller and throws synchronously. Because the throw is inside a bare timer callback with no handler, Node surfaces it as an **uncaught exception** and the serverless instance crashes (`mechanism: auto.node.onuncaughtexception`, `level: fatal`). Any in-flight work on that instance is lost.

**To Reproduce**

1. Create a Next.js App Router route handler that serves Inngest with streaming:
   ```ts
   // app/api/inngest/route.ts
   import { serve } from "inngest/next";
   import { inngest } from "@/inngest/client";
   import { myLongRunningFunction } from "@/inngest/functions";

   export const runtime = "nodejs";
   export const { GET, POST, PUT } = serve({
     client: inngest,
     functions: [myLongRunningFunction],
     streaming: true,
   });
   ```
2. Deploy to Vercel with Fluid Compute enabled (or any platform where an upstream proxy may reset long-lived connections).
3. Trigger a function run whose execution window exceeds the proxy idle/reset threshold, or manually abort the executor's request mid-flight.
4. The heartbeat `setInterval` fires after the controller has been closed → `controller.enqueue` throws → Node process crashes with an uncaught exception.

**Expected behavior**

When the stream's controller is already closed/cancelled, heartbeat enqueues should be silently skipped (or the interval should self-cancel). An upstream disconnect is a normal condition for a long-lived streaming response and must not crash the Node process.

Offending code in `src/helpers/stream.ts` (compiled output `helpers/stream.js` shown, same shape in TS source):

```js
new ReadableStream({ start(controller) {
  const encoder = new TextEncoder();
  const heartbeat = setInterval(() => {
    controller.enqueue(encoder.encode(value));   // throws if controller closed
  }, interval);
  const finalize = (data) => {
    clearInterval(heartbeat);
    Promise.resolve(data).then((resolvedData) => {
      controller.enqueue(encoder.encode(stringify(resolvedData))); // no .catch
      controller.close();
    });
  };
  passFinalize(finalize);
}})
```

Production stack from Sentry:

```
TypeError: Invalid state: Controller is already closed
    at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:1077:13)
    at inngest/src/helpers/stream.ts:47:31 (<object>.stream.start)
    at listOnTimeout (node:internal/timers:585:17)
    at process.processTimers (node:internal/timers:521:7)

mechanism: auto.node.onuncaughtexception
level: fatal
handled: no
transaction: POST /api/inngest
```

**Suggested fix**

Guard both enqueue sites and self-cancel the interval on the first failure:

```ts
const heartbeat = setInterval(() => {
  try {
    controller.enqueue(encoder.encode(value));
  } catch {
    clearInterval(heartbeat); // consumer went away; stop heartbeating
  }
}, interval);

const finalize = (data) => {
  clearInterval(heartbeat);
  Promise.resolve(data)
    .then((resolvedData) => {
      try {
        controller.enqueue(encoder.encode(stringify(resolvedData)));
        controller.close();
      } catch {
        // controller already closed by consumer; nothing to flush
      }
    })
    .catch(() => {
      // swallow; the run result is already persisted server-side
    });
};
```

- OS: Linux (Vercel runtime, `5.10.252-285.992.amzn2.x86_64`)
- npm package Version: `inngest@4.2.4`
- Framework: Next.js (App Router, `runtime = "nodejs"`)
- Platform: Vercel (Fluid Compute, region `iad1`), Node `v22.22.2`

**Additional context**

- Only occurs with `serve({ streaming: true })`; disabling streaming avoids the code path entirely but sacrifices the heartbeat that prevents "Unexpected ending response" errors on Fluid Compute for long checkpointed runs.
- Because the crash is an uncaught exception on a serverless instance, it can take down unrelated in-flight requests on the same warm instance, not just the stream that was cancelled.
- Happy to open a PR with the guard above if that's useful.

tsushanth:
## Root cause

`createStream` in `src/helpers/stream.ts` creates a `ReadableStream` with a heartbeat `setInterval` and a `finalize` callback. Neither path guards against the controller already being closed:

- **Heartbeat**: `controller.enqueue(...)` is called unconditionally every `interval` ms. If an upstream proxy (Vercel Fluid Compute, etc.) disconnects and the platform cancels the stream, the next tick throws inside a bare `setInterval` callback → Node surfaces it as an **uncaught exception** → serverless instance crashes.
- **Finalize**: `Promise.resolve(data).then(...)` has no `.catch()`, and `controller.enqueue` + `controller.close()` inside it are unguarded. A post-cancel resolve produces an unhandled rejection.
- **`cancel` callback**: missing entirely, so there's no hook to self-cancel the interval when the consumer closes the stream.

## Fix

```diff
-     const stream = new ReadableStream({
-       start(controller) {
-         const encoder = new TextEncoder();
-
-         const heartbeat = setInterval(() => {
-           controller.enqueue(encoder.encode(value));
-         }, interval);
-
-         const finalize = (data: unknown) => {
-           clearInterval(heartbeat);
-           void Promise.resolve(data).then((resolvedData) => {
-             controller.enqueue(encoder.encode(stringify(resolvedData)));
-             controller.close();
-           });
-         };
+     let closed = false;
+     let heartbeatTimer: ReturnType<typeof setInterval> | undefined;
+
+     const stream = new ReadableStream({
+       start(controller) {
+         const encoder = new TextEncoder();
+
+         heartbeatTimer = setInterval(() => {
+           if (closed) return;
+           try {
+             controller.enqueue(encoder.encode(value));
+           } catch {
+             closed = true;
+             clearInterval(heartbeatTimer);
+           }
+         }, interval);
+
+         const finalize = (data: unknown) => {
+           clearInterval(heartbeatTimer);
+           void Promise.resolve(data)
+             .then((resolvedData) => {
+               if (closed) return;
+               closed = true;
+               controller.enqueue(encoder.encode(stringify(resolvedData)));
+               controller.close();
+             })
+             .catch(() => {});
+         };
+       },
+       cancel() {
+         closed = true;
+         clearInterval(heartbeatTimer);
+       },
```

- `closed` and `heartbeatTimer` are hoisted outside the `ReadableStream` callbacks so `cancel` can share them with `start`.
- The heartbeat skips enqueue when `closed` and catches the TOCTOU race where the controller closes between the guard and the actual enqueue.
- `finalize` skips enqueue/close when `closed` and adds `.catch()` to handle post-cancel resolves.
- `cancel` sets `closed = true` and clears the interval immediately on consumer disconnect.

Fixes #1475

### DIFF
--- a/.changeset/fix-stream-closed-controller.md
+++ b/.changeset/fix-stream-closed-controller.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+fix(stream): guard heartbeat and finalize against closed ReadableStream controller
--- a/packages/inngest/src/helpers/stream.ts
+++ b/packages/inngest/src/helpers/stream.ts
@@ -42,27 +42,49 @@ export const createStream = (opts?: {
 
   return new Promise(async (resolve, reject) => {
     try {
+      let closed = false;
+      let heartbeatTimer: ReturnType<typeof setInterval> | undefined;
+
       const stream = new ReadableStream({
         start(controller) {
           const encoder = new TextEncoder();
 
-          const heartbeat = setInterval(() => {
-            controller.enqueue(encoder.encode(value));
+          heartbeatTimer = setInterval(() => {
+            if (closed) return;
+            try {
+              controller.enqueue(encoder.encode(value));
+            } catch {
+              // Controller was closed between the guard and the enqueue; ignore.
+              closed = true;
+              clearInterval(heartbeatTimer);
+            }
           }, interval);
 
           const finalize = (data: unknown) => {
-            clearInterval(heartbeat);
+            clearInterval(heartbeatTimer);
 
             // `data` may be a `Promise`. If it is, we need to wait for it to
             // resolve before sending it. To support this elegantly we'll always
             // assume it's a promise and handle that case.
-            void Promise.resolve(data).then((resolvedData) => {
-              controller.enqueue(encoder.encode(stringify(resolvedData)));
-              controller.close();
-            });
+            void Promise.resolve(data)
+              .then((resolvedData) => {
+                if (closed) return;
+                closed = true;
+                controller.enqueue(encoder.encode(stringify(resolvedData)));
+                controller.close();
+              })
+              .catch(() => {
+                // Stream was already closed by the time the final value resolved.
+              });
           };
 
           passFinalize(finalize);
+        },
+        cancel() {
+          // Consumer closed the stream (e.g. upstream proxy disconnect).
+          // Mark closed so the heartbeat self-cancels without throwing.
+          closed = true;
+          clearInterval(heartbeatTimer);
         },
       });
 


Linell:
Hey @tsushanth! Thank you for you contribution, it looks great!! I actually _just missed_ your PR and opened one of my own here with similar changes, and it's been merged: https://github.com/inngest/inngest-js/pull/1615

I'm going to close this one for now, but the contribution is much appreciated!

PR #1616, opened 2026-07-02
fix: restore OTel context captured at step.run() when executing step callback

tsushanth:
## What

Fixes #1436: third-party OTel context entries (e.g. Langfuse's `propagateAttributes` user/session IDs) set via `context.with()` before calling `step.run()` are lost inside the step callback.

## Root cause

Inngest functions are deterministic replay machines. When a step is first encountered, the engine's `stepHandler` registers the step and returns a pending promise to userland. The actual step callback (`fn`) is executed later by `executeStep`, which is called from the engine's **core loop** — not from within the user's `context.with()` scope.

By the time `executeStep` runs the step callback, the `context.with()` scope that `propagateAttributes` (or any other OTel-context-setting code) set up around `step.run()` has already exited. The OTel `AsyncLocalStorage` no longer holds those context entries, so spans created inside the step callback are missing them.

## Fix

Two small changes:

**`InngestStepTools.ts`** — add `otelCtx?: OtelContext` to `FoundStep` to carry the captured context.

**`engine.ts`** — in `stepHandler`, when a new step with a handler is registered, snapshot `otelContext.active()` and store it in `step.otelCtx`. Then in `executeStep`, wrap the `fn` invocation with `otelContext.with(capturedCtx, ...)` to restore the context the user had active when `step.run()` was called.

```ts
// engine.ts — stepHandler (step registration)
otelCtx: opts?.fn ? otelContext.active() : undefined,

// engine.ts — executeStep (step execution)
const capturedOtelCtx = foundStep.otelCtx ?? otelContext.active();
const actualHandler = () =>
  otelContext.with(capturedOtelCtx, () => runAsPromise(fn));
```

This is the standard OTel pattern for carrying context across async boundaries that don't flow naturally through a continuous `await` chain.

## Testing

Reproducer: wrap `step.run("s", callback)` in `context.with(ctx, ...)` where `ctx` carries a custom OTel attribute. Without this fix, the attribute is absent on spans created inside `callback`. With this fix, it is present.

Fixes #1436

### DIFF
--- a/packages/inngest/src/components/InngestStepTools.ts
+++ b/packages/inngest/src/components/InngestStepTools.ts
@@ -1,3 +1,4 @@
+import type { Context as OtelContext } from "@opentelemetry/api";
 import { type AiAdapter, models } from "@inngest/ai";
 import type { StandardSchemaV1 } from "@standard-schema/spec";
 import { z } from "zod/v3";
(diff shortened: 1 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #1625, opened 2026-07-13
feat: support for aborting defers

Linell:
## Summary

First PR in a stack: #1625 → #1628 → #1629.

Adds `defer()` abort support: `defer(...)` now returns a handle whose `abort()` cancels the deferred run via a `DeferAbort` opcode. Aborts are sync, fire-and-forget, and idempotent; a co-batched add+abort pair is sanitized at the wire boundary so the pair can't race in the executor.

Deliberately ships without the previously proposed `RunError` opcode ([Slack](https://inngest.slack.com/archives/C05QSAYPEG0/p1783638766004439)) to avoid version skew issues. Instead, a rejection with buffered defer ops rides the existing `[DeferAdd..., StepError|StepFailed]` batch, with two follow-up fixes stacked on this PR to close the gaps that opcode would have covered:

- #1628 — `RetryAfterError` timing rides a `Retry-After` header on the 206 — ignored by current executors (default backoff, status quo), honored once the backend supports it. No backend changes required to ship.
- #1629 — The executor's error-discovery re-invocation after a terminal failure is short-circuited (the SDK detects the memoized terminal op and re-reports the rejection without re-running user code or its side effects).

## Checklist

## Related

- INN-


packages/inngest/src/components/execution/engine.ts
```diff
@@ -2653,33 +2682,156 @@ class InngestExecutionEngine
...
+        return;
+      }
+
+      // Cancel a still-buffered add locally, but ship the abort regardless:
+      // the executor can dispatch a request with a stale `defers` map, so
+      // "the add never shipped" is not knowable locally. Unknown-target
+      // aborts are tolerated by the backend.
+      if (!this.state.lazyOps.remove(targetHashedId)) {
+        const prior = this.state.priorDefers[targetHashedId];
+        if (isRecord(prior) && prior.abortable === false) {
```
amh4r: Needing these assertions is a smell for the preexisting `priorDefers` type, which is `Record<string, unknown>`. Since we expect `abortable`, it should be `Record<string, { abortable: boolean }>`. That change will probably necessitate a few other wiring points, including the Zod schema where we validate the incoming SDK request body

packages/inngest/src/components/execution/engine.ts
```diff
@@ -130,6 +130,13 @@ function errorMessage(error: unknown): string {
  */
 const RUN_COMPLETE_STEP_ID = "complete";
 
+/**
+ * Reserved step ID for the synthetic run-level error op. Shipped raw (never
+ * hashed): user step IDs always reach the wire as sha1 hex, so a non-hex
+ * literal can never collide with a user step (e.g. one named "complete").
+ */
+const RUN_ERROR_STEP_ID = "~inngest:run-error";
```
amh4r: Is this necessary? I thought we were removing the new `RunError` opcode in the immediate defer abort work

Linell: Removed. The reserved `~inngest:run-error` ID was part of the `RunError`-era machinery _but_ it was still letting the SDK recognize the executor's error-discovery replay so it doesn't re-run user code. Now it works via a `__inngestRunRejection` marker in `error.data`, only added when a rejection ships alongside buffered defer ops to keep the blast radius super tight.

Linell: All of that is in #1629 now 👍

PR #1633, opened 2026-07-16
internal: add internal sessionPropagation toggle

scottnuma:
## Summary

- We're building out session propagation. Eventually this will be a configurable field.
- Let's wire up this configuration. We'll default it to off. Our future fields can honor the toggle, and we can add public access to it upon release
- We add INNGEST_SESSION_PROPAGATION as an env var to also serve as conifg. We still default to off.

## Checklist

## Related

- INN-

---

### DIFF
--- a/packages/inngest/src/components/Inngest.test.ts
+++ b/packages/inngest/src/components/Inngest.test.ts
@@ -1403,32 +1403,101 @@ describe("inngest.realtime.publish", () => {
 describe("sessionPropagation toggle", () => {
   // `sessionPropagation` is an internal, undocumented option that is
   // intentionally kept off the public `ClientOptions` type, so it's set via a
-  // narrow cast — the same way the SDK reads it internally.
-  const createWithSessionPropagation = (
-    sessionPropagation?: boolean,
-  ): Inngest.Any => {
+  // narrow cast — the same way the SDK reads it internally. `env` stubs
+  // `process.env` for the duration of construction, restoring it afterwards.
+  const createWithSessionPropagation = ({
+    option,
+    env,
+  }: {
+    option?: boolean;
+    env?: Record<string, string>;
+  } = {}): Inngest.Any => {
     const opts = { id: "test" } as ConstructorParameters<typeof Inngest>[0];
 
-    if (sessionPropagation !== undefined) {
-      (opts as { sessionPropagation?: boolean }).sessionPropagation =
-        sessionPropagation;
+    if (option !== undefined) {
+      (opts as { sessionPropagation?: boolean }).sessionPropagation = option;
     }
 
-    return new Inngest(opts);
+    let ogKeys: Record<string, string | undefined> = {};
+    if (env) {
+      ogKeys = Object.keys(env).reduce<Record<string, string | undefined>>(
+        (acc, key) => {
+          acc[key] = process.env[key];
+          process.env[key] = env[key];
+          return acc;
+        },
+        {},
+      );
+    }
+
+    try {
+      return new Inngest(opts);
+    } finally {
+      if (env) {
+        // biome-ignore lint/complexity/noForEach: intentional
+        Object.keys(ogKeys).forEach((key) => {
+          process.env[key] = ogKeys[key];
+        });
+      }
+    }
   };
 
-  test("defaults to false (dark-launch OFF) when the option is omitted", () => {
+  test("defaults to false (dark-launch OFF) when neither option nor env is set", () => {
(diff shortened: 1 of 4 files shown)


packages/inngest/src/components/Inngest.ts
```diff
@@ -378,6 +397,15 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
     this._logger = logger ?? new ConsoleLogger();
     this[internalLoggerSymbol] = this.options.internalLogger ?? this._logger;
 
+    // `sessionPropagation` is an internal, undocumented option intentionally
+    // kept off the public `ClientOptions` type, so it's read via a narrow cast.
+    const sessionPropagationOption = (
+      this.options as TClientOpts & { sessionPropagation?: boolean }
+    ).sessionPropagation;
```
amh4r: Wanna start with an undocumented env var (e.g. `INNGEST_SESSION_PROPAGATION`)? It's something we'll need anyway, and it avoids the need for throwaway code while still giving us "secret setting until we're ready to expose it"

scottnuma: added

packages/inngest/src/components/Inngest.ts
```diff
@@ -378,6 +398,29 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
     this._logger = logger ?? new ConsoleLogger();
     this[internalLoggerSymbol] = this.options.internalLogger ?? this._logger;
 
+    // Session propagation is resolved with the precedence
+    // `sessionPropagation` option > `INNGEST_SESSION_PROPAGATION` env var >
+    // default, matching the SDK's `dev`-mode resolution order.
+    let sessionPropagation = SESSION_PROPAGATION_DEFAULT_ENABLED;
+
+    const sessionPropagationEnv = parseAsBoolean(
+      this._env[envKeys.InngestSessionPropagation],
```
amh4r: We can't accurately check env vars at client construction time due to the nature of some serverless runtimes. This is why we have the `setEnvVars()` shenanigans.

In Node serverless runtimes, `process.env` is usually available. But in edge/non-Node runtimes, env vars may be exposed as request/runtime bindings instead of global process state. Examples include Cloudflare Workers-style APIs where the handler receives something like env per request, and there may be no real process object at all.

packages/inngest/src/components/Inngest.test.ts
```diff
@@ -1398,3 +1399,105 @@ describe("inngest.realtime.publish", () => {
...
+    option,
+    env,
+  }: {
+    option?: boolean;
+    env?: Record<string, string>;
+  } = {}): Inngest.Any => {
+    const opts = { id: "test" } as ConstructorParameters<typeof Inngest>[0];
+
+    if (option !== undefined) {
+      (opts as { sessionPropagation?: boolean }).sessionPropagation = option;
```
amh4r: Is this still needed? I'm wondering if it's a vestigial thing with the client constructor option

packages/inngest/src/components/Inngest.test.ts
```diff
@@ -1398,3 +1399,105 @@ describe("inngest.realtime.publish", () => {
...
+    }
+
+    let ogKeys: Record<string, string | undefined> = {};
+    if (env) {
+      ogKeys = Object.keys(env).reduce<Record<string, string | undefined>>(
+        (acc, key) => {
+          acc[key] = process.env[key];
```
amh4r: Use `vi.stubEnv()` for env vars. The code probably works as-is, but:
- Messing with globals increases the likelihood of bleed between tests
- `vi.stubEnv()` is cleaner

scottnuma: added, thanks!

packages/inngest/src/components/Inngest.test.ts
scottnuma: nope - the env vars are the clear method for setting, committed

packages/inngest/src/components/Inngest.ts
scottnuma: adjusted to match our existing patterns. 

Cloudflare workers is seeming like a scarier and scarier runtime. I noted in the spec updates that `inngest.send` will not have session propagation on CF workers too.

scottnuma: I've added a Claude.md commit as well to try and steer away from this problem in the future. Can remove / pull it into a different PR to workshop some more.

amh4r (approved):

PR #1634, opened 2026-07-16
internal: send propagated sessions with step.sendEvent

scottnuma:
## Summary

- step.sendEvent derives propagated sessions from its triggering events and sends them along during step.sendEvent
- Gated behind the internal session propagation toggle

## Checklist

## Related

- https://linear.app/inngest/issue/EXE-2033/stepsendevent-propagates-sessions

---

### DIFF
--- a/packages/inngest/src/components/Inngest.ts
+++ b/packages/inngest/src/components/Inngest.ts
@@ -460,8 +460,9 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
 
   /**
    * Whether session propagation is enabled for this client. Resolved with the
-   * precedence `INNGEST_SESSION_PROPAGATION` env var > {@link
-   * SESSION_PROPAGATION_DEFAULT_ENABLED}.
+   * precedence explicit `sessionPropagation` option > `INNGEST_SESSION_PROPAGATION`
+   * env var > {@link SESSION_PROPAGATION_DEFAULT_ENABLED}, mirroring how
+   * {@link mode} resolves the `isDev` option ahead of the env var.
    *
    * Resolved lazily on every access (mirroring {@link mode}) so that env vars
    * populated after construction via {@link setEnvVars} — as happens in
@@ -470,19 +471,25 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
    * @internal
    */
   get [sessionPropagationSymbol](): boolean {
-    // Session propagation is resolved with the precedence
-    // `INNGEST_SESSION_PROPAGATION` env var > default, matching the SDK's
-    // `dev`-mode resolution order.
-    let sessionPropagation = SESSION_PROPAGATION_DEFAULT_ENABLED;
+    // An explicit boolean option wins outright (like `isDev` in `mode`); it is
+    // internal/undocumented, so it lives off the public `ClientOptions` type
+    // and is read via a cast.
+    const sessionPropagationOption = (
+      this.options as { sessionPropagation?: unknown }
+    ).sessionPropagation;
+    if (typeof sessionPropagationOption === "boolean") {
+      return sessionPropagationOption;
+    }
 
+    // Otherwise fall back to the env var, then the default.
     const sessionPropagationEnv = parseAsBoolean(
       this._env[envKeys.InngestSessionPropagation],
     );
     if (sessionPropagationEnv !== undefined) {
-      sessionPropagation = sessionPropagationEnv;
+      return sessionPropagationEnv;
     }
 
-    return sessionPropagation;
+    return SESSION_PROPAGATION_DEFAULT_ENABLED;
   }
 
   get mode(): Mode {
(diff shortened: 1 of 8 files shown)


scottnuma:
is internal the right classification for this PR?

packages/inngest/src/components/Inngest.ts
```diff
@@ -452,19 +453,25 @@ export class Inngest<const TClientOpts extends ClientOptions = ClientOptions>
...
   get [sessionPropagationSymbol](): boolean {
-    // Session propagation is resolved with the precedence
-    // `INNGEST_SESSION_PROPAGATION` env var > default, matching the SDK's
-    // `dev`-mode resolution order.
-    let sessionPropagation = SESSION_PROPAGATION_DEFAULT_ENABLED;
+    // An explicit boolean option wins outright (like `isDev` in `mode`); it is
+    // internal/undocumented, so it lives off the public `ClientOptions` type
+    // and is read via a cast.
+    const sessionPropagationOption = (
+      this.options as { sessionPropagation?: unknown }
```
Linell: The cast makes the value read as `unknown`, so the write sites accept either `true` or `"true"` without an error. The read and write sites also assert the shape independently, which could be better.

Why not declare it on `ClientOptions` a la `isDev`? We can use the same `EXPERIMENTAL` banner as `defer`. Then you could make this bit work like `mode`.

packages/inngest/src/types.ts
```diff
@@ -761,6 +769,17 @@ export type EventMeta = {
...
   sessions?: EventSessions;
+
+  /**
+   * Sessions propagated from the run that emitted this event. Stamped
+   * automatically from `ctx.sessions`; carried as a separate layer from manual
+   * {@link EventMeta.sessions} and merged server-side.
+   *
+   * Not intended to be set by hand.
+   *
+   * @internal
```
Linell: This could be handled with the `EXPERIMENTAL` comment [here](https://github.com/inngest/inngest-js/pull/1634/changes#r3648987422).

packages/inngest/src/types.ts
```diff
@@ -579,6 +579,14 @@ export type BaseContext<TClient extends Inngest.Any> = {
   event: Simplify<EventPayload>;
   events: AsTuple<Simplify<EventPayload>>;
 
+  /**
```
Linell, line 582: "By default these are propagated" is easy to read as propagation being on by default, but the toggle defaults to off and this doc doesn't mention the toggle at all.

Could we describe it as what happens when propagation is enabled? Worth keeping the `step.sendEvent` scoping, since `inngest.send()` and `step.invoke` don't propagate.

packages/inngest/src/components/execution/engine.ts
```diff
@@ -2156,6 +2157,22 @@ class InngestExecutionEngine
       defer,
     } as unknown as Context.Any;
 
+    // Aggregate the run's triggering events into its propagated sessions in
+    // `ctx.sessions`
+    //
+    // Gated by the client-level session-propagation toggle so that,
+    // while disabled, `ctx.sessions` stays unset and no new surface appears
+    // on the handler context.
+    if (this.options.client[sessionPropagationSymbol]) {
```
Linell, line 2166: `reduceEventsToPropagatedSessions` returns `{}` for a run with no sessions, so `ctx.sessions` ends up truthy but empty on an optional field. No functional impact, since `stampPropagatedSessions` guards on empty, but `normalizeEventSessions` and `normalizeEventMeta` both collapse empty to `undefined`, so it's a little inconsistent. A guard on the assignment would keep the property absent.

Also `fnArg.sessions = undefined` in the catch creates the property, so it shows up in `Object.keys` and spreads. I think an empty catch would fix that?

Linell (changes requested):

packages/inngest/src/components/execution/engine.ts
scottnuma: Addressed this over in b203871784d2f43cb83229d0c9f7e1a47c36427d.

packages/inngest/src/components/Inngest.ts
scottnuma: Added in c3c93225b5a7f2dffc878ebe91a2a3fb2fe70cda

This pulls forward a few changes from the later PR https://github.com/inngest/inngest-js/pull/1645, where I originally added more of the public client config.

packages/inngest/src/types.ts
scottnuma: Added in ab2c6a3647326f865e687b65542e4fdbf2f89aa6

packages/inngest/src/types.ts
scottnuma: Added in 60220e7f5e1f2db1386e960a9a7b80bf82bb4af2

scottnuma:
Accepted your comments and re-requested review! Let me know if anything still differs.

amh4r (approved):

Linell (approved):
🍰

PR #1636, opened 2026-07-17
internal: invoke propagates sessions

scottnuma:
## Summary

- step.invoke should also propagate sessions

## Checklist

## Related

- EXE-2034

---

### DIFF
--- a/packages/inngest/src/components/execution/engine.ts
+++ b/packages/inngest/src/components/execution/engine.ts
@@ -2407,7 +2407,7 @@ class InngestExecutionEngine
         opId.opts?.payload
       ) {
         opId.opts.payload = stampPropagatedSessionsOnInvoke(
-          opId.opts.payload as MinimalEventPayload,
+          opId.opts.payload,
           this.fnArg.sessions,
         );
       }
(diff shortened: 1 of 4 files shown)


packages/inngest/src/components/execution/engine.ts
```diff
@@ -2384,6 +2386,29 @@ class InngestExecutionEngine
...
+      // where `this.fnArg` is in scope, avoiding an ALS read on the send path.
+      //
+      // Gated by the client-level session-propagation toggle.
+      if (
+        opId.op === StepOpCode.InvokeFunction &&
+        this.options.client[sessionPropagationSymbol] &&
+        opId.opts?.payload
+      ) {
+        opId.opts.payload = stampPropagatedSessionsOnInvoke(
+          opId.opts.payload as MinimalEventPayload,
```
scottnuma: I looked at this a few different ways (type guard, isRecord, stronger typing upstream) to see if we could avoid this cast, but none of them felt high enough on the gains to effort ratio. Open to suggestions here

amh4r: Is the cast necessary? I don't see a type error when I remove it

```suggestion
          opId.opts.payload,
```

scottnuma: Hmm yeah, we don't actually need anything here.

amh4r (approved):

PR #1637, opened 2026-07-17
internal: support null tombstones in sessions

scottnuma:
## Summary

- Users should be able to override propagated sessions
- Setting an individual key will clear the propagated session
- Setting the entire sessions object to null will clear all propagated sessions

## Checklist

## Related

- EXE-2034

---

### DIFF
--- a/packages/inngest/src/helpers/sessions.ts
+++ b/packages/inngest/src/helpers/sessions.ts
@@ -8,26 +8,20 @@ import type { EventMeta, EventSessions } from "../types.ts";
 type NormalizedSessions = Record<string, string | null>;
 
 /**
- * Validates event sessions and normalizes their values to strings, matching
- * the shape stored by the Inngest server.
- *
- * When `allowCut` is `true` (the manual layer), `null` values are preserved as
- * RFC 7386 tombstones rather than rejected, and a whole-field `null` is
- * preserved (as `null`) to mean "clear all inherited sessions". Tombstones are
- * consumed server-side and never count against the per-event session limit.
- * When `allowCut` is `false` (the propagated layer) `null` values are rejected
- * and a whole-field `null` is dropped, since it is machine-stamped and carries
- * no tombstones. The flag is required so each call site declares which side of
- * this trust boundary it is on rather than relying on a default.
+ * Validates event sessions and normalizes their values to strings, matching the
+ * shape stored by the Inngest server. Shared by the two layer-specific entry
+ * points below; `layer` selects how `null` is treated.
  *
  * Returns `undefined` if no sessions were given (an absent field, or an empty
  * object — an empty RFC 7386 patch leaves the inherited layer untouched);
  * returns `null` for a preserved "clear all" tombstone.
  */
-export const normalizeEventSessions = (
+const normalizeSessions = (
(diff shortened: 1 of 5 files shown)


packages/inngest/src/helpers/sessions.ts
```diff
@@ -1,17 +1,41 @@
...
+ *
+ * Returns `undefined` if no sessions were given (an absent field, or an empty
+ * object — an empty RFC 7386 patch leaves the inherited layer untouched);
+ * returns `null` for a preserved "clear all" tombstone.
  */
 export const normalizeEventSessions = (
   sessions: EventSessions | null | undefined,
-): Record<string, string> | undefined => {
-  if (sessions === undefined || sessions === null) {
+  allowCut: boolean,
```
amh4r: I have a couple thoughts on this.

## Is the `allowCut` parameter needed?

Is it true that a propagated session can't have a null value? If so, could we just add the "if null then cut" logic to this function and apply it to both sessions and propagated sessions?

In other words, if that null invariant is true then the "if null then cut" path would be unreachable for propagated sessions, allowing us to alleviate the `allowCut` burden from the consumer.

## If `allowCut`-like split is needed, can we have 2 different functions?

I'm a bit worried about how this boolean communicates its intent to callsites. For example, we see this elsewhere in the file:

```ts
  const sessions = normalizeEventSessions(meta.sessions, true);
  const propagatedSessions = normalizeEventSessions(
    meta.propagatedSessions,
    false,
  );
```

It's just a boolean with no implied meaning. Maybe it'd be better to have 2 functions: `normalizeManualSessions` and `normalizePropagatedSessions`?

packages/inngest/src/types.ts
```diff
@@ -744,14 +744,23 @@ export interface EventPayload<TData = any> extends MinimalEventPayload<TData> {
...
+ * `null` is a tombstone (RFC 7386 JSON Merge Patch), not a session id: on the
+ * manual {@link EventMeta.sessions} layer it cuts the inherited (propagated)
+ * session of the same key. It is consumed server-side and never counts against
+ * the per-event session limit. Received events never carry `null` — see
+ * {@link ReceivedEventMeta}.
+ *
  * @public
  */
-export type EventSessionValue = string | number;
+export type EventSessionValue = string | number | null;
```
amh4r, line 756: This change is fine, but should we create a dedicated type for propagated sessions? Since (IIUC), null values are invalid for them

```ts
export type PropagatedEventSessions = {
  [K in keyof EventSessions]: Exclude<EventSessions[K], null>;
};
```

packages/inngest/src/helpers/sessions.ts
scottnuma: > Is it true that a propagated session can't have a null value?

Propagated sessions shouldn't have a null value, though it's possible for users to mess with the values, so it's useful to catch and enforce here.

2 different functions sounds good and more readable than a bool. Splitting.

packages/inngest/src/types.ts
scottnuma: Adding a dedicated type

amh4r (approved):

PR #1639, opened 2026-07-17
internal: inngest.send supports session propagation

scottnuma:
## Summary
 
- `inngest.send` should support session propagation
- Unlike `step.sendEvent` and `step.Invoke`, `inngest.send` doesn't have ctx. We instead rely upon async local storage to determine when we're in the context of a run.
- Async local storage is not available in all runtimes, such as some of cloudflare workers. We will call this out in the docs.

## Checklist

~- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR~ docs later
~- [ ] Added changesets if applicable~ unreleased 

## Related

- EXE-2044

---

### DIFF
--- a/packages/inngest/src/components/Inngest.test.ts
+++ b/packages/inngest/src/components/Inngest.test.ts
@@ -14,6 +14,8 @@ import {
 import type { Logger } from "../middleware/logger.ts";
 import { createClient, nodeVersion } from "../test/helpers.ts";
 import type { SendEventResponse } from "../types.ts";
+import type { AsyncContext } from "./execution/als.ts";
+import { getAsyncLocalStorage } from "./execution/als.ts";
 import { sessionPropagationSymbol } from "./Inngest.ts";
 import type { createStepTools } from "./InngestStepTools.ts";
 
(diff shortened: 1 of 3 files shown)


amh4r:
Just something to note: cross-environment propagation seems possible. I don't necessarily think that's an issue in this PR, but something to be aware of

packages/inngest/src/components/Inngest.test.ts
```diff
@@ -1502,3 +1504,142 @@ describe("sessionPropagation toggle", () => {
...
+  });
+
+  const createClientWithPropagation = (
+    sessionPropagation: boolean,
+  ): Inngest.Any =>
+    createClient({
+      id: "test",
+      eventKey: testEventKey,
+      // Internal, undocumented option kept off the public `ClientOptions` type.
+      ...({ sessionPropagation } as { sessionPropagation: boolean }),
```
amh4r: Is this still necessary? IIRC we removed the client option for now

scottnuma: removed

amh4r (approved):

PR #1643, opened 2026-07-20
feat(middleware): add Middleware.isStepType type guard

scottnuma:
## Summary

- Middleware can be tricky to use due to the unknown types that vary based on the different step types
- We add a type guard to resolve typing based upon the step type
- We take special care to prevent this from causing breaking changes in the future

## Checklist

~- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR~ https://linear.app/inngest/issue/EXE-2052/add-type-guard-to-middleware-docs

## Related

- EXE-2036

### DIFF
--- a/packages/inngest/src/components/middleware/middleware.test.ts
+++ b/packages/inngest/src/components/middleware/middleware.test.ts
@@ -1,6 +1,45 @@
-import { test } from "vitest";
+import { expect, expectTypeOf, test } from "vitest";
 import { Inngest } from "../Inngest.ts";
 import { Middleware } from "./middleware.ts";
+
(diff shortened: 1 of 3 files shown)


packages/inngest/src/components/middleware/middleware.test.ts
```diff
@@ -1,7 +1,46 @@
-import { test } from "vitest";
+import { expect, expectTypeOf, test } from "vitest";
```
amh4r, line 1: Can you add a test to ensure the type narrowing is working for the input? IIUC, we want this helper for improved static safety in `transformStepInput`

scottnuma: added

PR #1646, opened 2026-07-21
feat(defer): propagate sessions through defer meta

scottnuma:
## Summary

- Adds support for sessions to deferred functions. Also adds support for propagated sessions to deferred runs
- We add additional server side support in oss and monorepo

Rebased onto `main` now that the rest of the stack has landed; this PR is no longer stacked. Note the run-level `ctx.sessions` is now the *intersection* of the triggering events' sessions (#1668), which the defer tests reflect.

## Checklist

~- [ ] Added a [docs PR](https://github.com/inngest/website) that references this PR~

## Related

- EXE-2035

### DIFF
--- a/packages/inngest/src/components/execution/engine.test.ts
+++ b/packages/inngest/src/components/execution/engine.test.ts
@@ -1021,3 +1021,116 @@ describe("step.invoke session propagation", () => {
     });
   });
 });
+
+describe("defer session propagation", () => {
+  // Drives a real execution whose handler calls `defer(...)`, then inspects the
+  // buffered `DeferAdd` op's `opts.meta`. Asserts the engine resolves the
+  // defer's session layers: the manual `meta.sessions` layer is normalized, and
+  // the run's `ctx.sessions` (aggregated from triggering events) is stamped as
+  // the separate `meta.propagated_sessions` layer, gated by the client toggle.
+  const runDefer = async (
+    enabled: boolean,
+    manualMeta?: Record<string, unknown>,
+  ) => {
+    const client = createClient({
+      id: "test",
+      isDev: true,
+      sessionPropagation: enabled,
+    });
+
+    const target = createDefer(
+      client,
+      { id: "target-defer" },
+      () => "target-result",
+    );
+
+    const fn = new InngestFunction(
+      client,
+      { id: "caller-fn", triggers: [{ event: "test/event" }] },
+      ({
+        defer,
+      }: Context.Any & { defer: (id: string, opts: unknown) => unknown }) => {
+        defer("d", {
+          function: target,
+          data: { foo: "bar" },
+          ...(manualMeta ? { meta: manualMeta } : {}),
+        });
+        return "done";
+      },
+    );
+
+    // Two triggering events carrying sessions — exercises the run-level
+    // aggregation that produces `ctx.sessions`.
+    const events = [
+      { name: "test/event", data: {}, meta: { sessions: { conv_id: "p1" } } },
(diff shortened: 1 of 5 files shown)


packages/inngest/src/components/execution/engine.ts
```diff
@@ -2758,6 +2761,26 @@ class InngestExecutionEngine
...
+        // run inherits. The manual `meta.sessions` layer (tombstones preserved)
+        // is always normalized; the run's propagated sessions (`ctx.sessions`)
+        // are stamped as the separate `meta.propagatedSessions` layer, gated by
+        // the client toggle, mirroring `step.invoke`/`step.sendEvent`. The
+        // server folds the two at finalize.
+        let deferMeta: ReturnType<typeof normalizeEventMeta>;
+        try {
+          deferMeta = normalizeEventMeta(meta);
+        } catch (err) {
+          log.error({ runId, err }, "defer skipped: invalid meta.sessions");
```
Linell: Nit: this also fires for non-object `meta` and for a bad `meta.propagatedSessions`, so maybe just `"invalid meta"`. The `err` is logged either way, so nothing's missing from the output.

packages/inngest/src/components/execution/engine.test.ts
```diff
@@ -1012,3 +1012,116 @@ describe("step.invoke session propagation", () => {
...
+  test("preserves a whole-field null (clear all inherited sessions)", async () => {
+    const meta = await runDefer(true, { sessions: null });
+    // `sessions: null` is the RFC 7386 whole-document tombstone: clear every
+    // inherited session. It must ship verbatim alongside the still-stamped
+    // propagated layer — the server folds/clears at finalize, not the SDK.
+    expect(meta).toBeDefined();
+    expect(meta?.sessions).toBeNull();
+    expect(meta?.propagatedSessions).toEqual({ conv_id: "p1", org_id: "42" });
+  });
+});
```
Linell, line 1161: Worth one test for the skip path? If `runDefer` were split so a variant returns the op without asserting it exists:

```ts
const { result, deferOp } = await startDefer(true, { sessions: { conv_id: {} } });
expect(deferOp).toBeUndefined();
expect(result).toMatchObject({
  steps: [expect.objectContaining({ op: "RunComplete", data: "done" })],
});
```

Ran it locally: passes, and fails if the catch continues instead of bailing. The `RunComplete` assertion is what pins "parent run still finishes".

Linell (approved):
Looks good. The two-layer split reads clearly, and the comment explaining why the SDK leaves the fold to the server is helpful for the next person in here.

Two small notes inline, neither blocking.

One thing for later, outside this PR's diff since it lands in the base branch: middleware can overwrite `ctx.sessions` via the `Object.assign` in `applyFunctionInputMutations`, and defer/invoke would then ship it unnormalized where `sendEvent` wouldn't. Freezing `ctx.sessions` after the reduce in `createFnArg` would cover both paths. Pre-existing with invoke, so no rush.

packages/inngest/src/components/execution/engine.ts
scottnuma: Changed to `invalid meta`

packages/inngest/src/components/execution/engine.test.ts
scottnuma: Added a test for this case

scottnuma:
Covering your comment with https://linear.app/inngest/issue/EXE-2135/ensure-were-covered-with-normalization-even-with-middleware

PR #1658, opened 2026-07-30
fix: wrapStep middleware hook sometimes skipped for parallel steps

### LINKED ISSUE #1656: [BUG] wrapStep middleware hook silently skipped for a sibling step when replaying parallel step.run() results
**Describe the bug**

`MiddlewareManager.activeWrapStep` (a `Set` used as a "middleware is currently inside its own `wrapStep`" recursion guard, `packages/inngest/src/components/middleware/manager.ts:81`) is scoped to the whole `MiddlewareManager` instance, not to an individual step. When two sibling steps that were originally run in parallel (e.g. via `Promise.all([step.run('a', ...), step.run('b', ...)])`) are **replayed together from memoized state in the same request** (which happens on every invocation once both steps have completed on the backend), their `wrapStep` calls land back-to-back in the same synchronous burst. If the middleware's `wrapStep` implementation does **any** async work before calling `next()` (e.g. `await` a key lookup before decrypting), the guard is still "held" by the first step when the second step's `wrapStep` call arrives, and the second step's `wrapStep` is silently skipped entirely: the caller receives the step's **raw, untransformed** value instead.

This is not cosmetic. For any `wrapStep`-based middleware that transforms step output (encryption, deserialization, etc.), this causes silent data corruption with no error at the point of failure. The failure only surfaces later, downstream, when calling code tries to use a value it assumed was already transformed.

**To Reproduce**

1. Create a function with a middleware whose `wrapStep` does async work before calling `next()`, and a handler that runs two steps in parallel via `Promise.all()`.
2. Let both steps complete, so the next invocation replays the function to completion using memoized step data (the normal, everyday terminal request for any function using parallel steps).
3. Observe that only the first of the two steps' values passed through `wrapStep`. The second is raw/untransformed.

This can be reproduced directly against the execution engine, no dev server or framework adapter needed. Add this file to the `inngest` package and run it with `pnpm test src/components/execution/parallel-repro.test.ts`:

```ts
// packages/inngest/src/components/execution/parallel-repro.test.ts
import { fromPartial } from "@total-typescript/shoehorn";
import { Middleware } from "../middleware/middleware.ts";
import { createClient } from "../../test/helpers.ts";
import { StepMode } from "../../types.ts";
import { InngestFunction } from "../InngestFunction.ts";
import { _internals } from "./engine.ts";

test("wrapStep is silently skipped for a sibling step on the memoized replay pass", async () => {
  const tagged: Record<string, unknown> = {};

  class TagMiddleware extends Middleware.BaseMiddleware {
    readonly id = "tag";
    override async wrapStep({ next, stepInfo }: Middleware.WrapStepArgs) {
      // Any async work before calling next() is enough to open the race,
      // e.g. a real middleware awaiting a decryption key lookup.
      await new Promise((r) => setTimeout(r, 10));
      const value = await next();
      tagged[stepInfo.options.id] = value;
      return { tagged: true, value };
    }
  }

  const client = createClient({ id: "test", middleware: [TagMiddleware] });
  let finalOutput: unknown;

  const fn = new InngestFunction(
    client,
    { id: "fn", triggers: [{ event: "evt" }] },
    async ({ step }: any) => {
      const [a, b] = await Promise.all([
        step.run("step-a", () => ({ from: "a" })),
        step.run("step-b", () => ({ from: "b" })),
      ]);
      finalOutput = { a, b };
      return finalOutput;
    },
  );

  const hashedA = _internals.hashId("step-a");
  const hashedB = _internals.hashId("step-b");

  // Simulate the request where both parallel steps are already memoized
  // (i.e. the backend has already run both and is asking the SDK to
  // replay the function to completion). This is a normal, everyday
  // request in the lifecycle of any function with parallel steps.
  const execution = fn["createExecution"]({
    partialOptions: {
      client,
      data: fromPartial({ event: { name: "evt", data: {} } }),
      runId: "run-1",
      stepState: {
        [hashedA]: { id: hashedA, data: { from: "a" }, fulfilled: true },
        [hashedB]: { id: hashedB, data: { from: "b" }, fulfilled: true },
      },
      stepCompletionOrder: [hashedA, hashedB],
      reqArgs: [],
      headers: {},
      stepMode: StepMode.Async,
    },
  });

  await execution.start();

  console.log(JSON.stringify(finalOutput));
  // Actual:   {"a":{"tagged":true,"value":{"from":"a"}},"b":{"from":"b"}}
  // Expected: {"a":{"tagged":true,"value":{"from":"a"}},"b":{"tagged":true,"value":{"from":"b"}}}
});
```

**Expected behavior**

Both `step-a` and `step-b` should have their values passed through `wrapStep`, regardless of the order they're replayed in or how they were originally scheduled. Expected output:

```
{"a":{"tagged":true,"value":{"from":"a"}},"b":{"tagged":true,"value":{"from":"b"}}}
```

Actual output from the repro above:

```
{"a":{"tagged":true,"value":{"from":"a"}},"b":{"from":"b"}}
```

`step-b`'s value is the raw `{ from: "b" }`. `wrapStep` never ran for it. `step-a`'s value is correctly wrapped.

Root cause, `packages/inngest/src/components/middleware/manager.ts`:

```ts
private readonly activeWrapStep = new Set<Middleware.BaseMiddleware>();   // line 81

buildWrapStepChain(handler, stepInfo) {                                   // line 293
  ...
      chain = () => {
        if (this.activeWrapStep.has(mw)) {                                // line 309
          return next();   // <-- silently skips wrapStep entirely
        }
        this.activeWrapStep.add(mw);
        const guardedNext = () => {
          this.activeWrapStep.delete(mw);
          return next().finally(() => { this.activeWrapStep.add(mw); });
        };
        return mw.wrapStep({ ctx: this.fnArg, fn: this.fn, next: guardedNext, stepInfo })
          .finally(() => this.activeWrapStep.delete(mw));
      };
  ...
}
```

`activeWrapStep` is a single `Set` on the `MiddlewareManager` instance, shared across every step processed by that instance, not scoped to an individual step's in-flight `wrapStep` call. The comment above its declaration describes it as an "infinite recursion guard" for a middleware calling `ctx.step.run()` from inside its own `wrapStep`, but the guard has no way to distinguish "the same middleware re-entering for the step it's currently wrapping" from "the same middleware being invoked for a completely unrelated sibling step that happens to be replaying at the same time."

- OS: N/A (reproduced via the SDK's execution engine directly, not OS-specific)
- npm package Version: `inngest@4.13.0` (verified against the published npm tarball's compiled `dist/`, the bug is present there too, not just on `main`)
- Framework: none in the repro below. It bypasses framework adapters entirely, drives `InngestFunction`/`InngestExecutionEngine` directly. We first noticed this via `@mastra/inngest` (Mastra's workflow engine) with `@inngest/middleware-encryption@2.0.0` registered, running two Mastra workflow steps in parallel
- Platform: N/A (Node v24.14.0 used for the repro, but the bug is a plain JS control-flow issue, not runtime-specific)

**Additional context**

The crux: `activeWrapStep` is keyed by middleware only, not by middleware and step. It should be tracking "is this middleware currently inside its own `wrapStep` for step X", but it can only answer "is this middleware currently inside a `wrapStep` at all", so two different steps going through the same middleware instance look identical to the guard.

This only shows up under a fairly specific combination. It needs 2+ steps invoked via `Promise.all` (sequential `await step.run('a'); await step.run('b');` doesn't race, since the second call only happens after the first settles), and it needs the affected steps to be in the "replay from memoized state" pass, which is the terminal request of essentially every function that uses parallel steps, not an edge case. It also needs the middleware's `wrapStep` to do some async work before calling `next()`. An immediate `next()` call doesn't race, which is likely why the SDK's own middleware test suite never caught this: none of its example `wrapStep` implementations do async setup before calling `next()`.

amh4r:
## Summary

Fixes a `wrapStep` middleware bug where async middleware could be skipped for one of multiple parallel `step.run()` calls during memoized replay.

The recursion guard for `wrapStep` is now scoped to the current async execution branch when `AsyncLocalStorage` is available. This keeps the intended protection for steps created inside a middleware's own `wrapStep`, while still allowing unrelated sibling steps in `Promise.all()` to run through the same middleware.

Also adds middleware architecture notes and AGENTS/CLAUDE routing files so future middleware changes have clearer lifecycle context.

Also add `--ignore-script=false` to the Dev Server test harness command, since the Dev Server uses a post install script.

Closes #1656.

## Checklist

### DIFF
--- a/.changeset/rotten-snails-punch.md
+++ b/.changeset/rotten-snails-punch.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Fix wrapStep middleware hook sometimes skipped for parallel steps
--- a/packages/inngest/src/components/execution/als.mocked.test.ts
+++ b/packages/inngest/src/components/execution/als.mocked.test.ts
@@ -86,6 +86,28 @@ describe("getAsyncCtx", () => {
     const store = await externalP;
     expect(store).toBeUndefined();
   });
+
+  test("should execute step tooling if node:async_hooks is not supported", async () => {
+    vi.spyOn(console, "warn").mockImplementation(() => {});
+
+    const { Inngest } = await import("../../index.ts");
+
+    const inngest = new Inngest({ id: "test" });
+    const fn = inngest.createFunction(
+      { id: "test", triggers: [{ event: "" }] },
+      async ({ step }) => {
+        return await step.run("step", () => "done");
+      },
+    );
+
+    // biome-ignore lint/suspicious/noExplicitAny: intentional
+    const t = new InngestTestEngine({ function: fn as any });
+
+    const { result, error } = await t.execute();
+
+    expect(error).toBeUndefined();
+    expect(result).toBe("done");
+  });
 });
 
 describe("getAsyncCtxSync", () => {
--- a/packages/inngest/src/components/execution/als.ts
+++ b/packages/inngest/src/components/execution/als.ts
@@ -168,6 +168,15 @@ export const getAsyncCtxSync = (): AsyncContext | undefined => {
   return getCache().resolved?.getStore();
 };
 
+export function mustGetAsyncContextSync(): AsyncContext {
+  const asyncContext = getAsyncCtxSync();
+  if (!asyncContext) {
+    throw new Error("Runtime does not support AsyncLocalStorage");
+  }
+
+  return asyncContext;
+}
+
 /**
  * Run a function within a specific async context if AsyncLocalStorage has
  * already been initialized.
--- a/packages/inngest/src/components/execution/als.ts
+++ b/packages/inngest/src/components/execution/als.ts
@@ -168,15 +168,6 @@ export const getAsyncCtxSync = (): AsyncContext | undefined => {
   return getCache().resolved?.getStore();
 };
 
-export function mustGetAsyncContextSync(): AsyncContext {
-  const asyncContext = getAsyncCtxSync();
-  if (!asyncContext) {
-    throw new Error("Runtime does not support AsyncLocalStorage");
-  }
-
-  return asyncContext;
-}
-
 /**
  * Run a function within a specific async context if AsyncLocalStorage has
  * already been initialized.
--- a/packages/inngest/src/components/execution/als.ts
+++ b/packages/inngest/src/components/execution/als.ts
@@ -169,6 +169,19 @@ export const getAsyncCtxSync = (): AsyncContext | undefined => {
 };
 
 /**
+ * Run a function within a specific async context if AsyncLocalStorage has
+ * already been initialized.
+ */
+export const runWithAsyncCtx = <R>(store: AsyncContext, fn: () => R): R => {
+  const als = getCache().resolved;
+  if (!als) {
+    return fn();
+  }
+
+  return als.run(store, fn);
+};
+
+/**
  * Get a singleton instance of `AsyncLocalStorage` used to store and retrieve
  * async context for the current execution.
  */
--- a/packages/inngest/src/components/execution/engine.ts
+++ b/packages/inngest/src/components/execution/engine.ts
@@ -82,8 +82,8 @@ import { Stream } from "../StreamTools.ts";
 import { validateEvents } from "../triggers/utils.js";
 import {
   getAsyncCtx,
+  getAsyncCtxSync,
   getAsyncLocalStorage,
-  mustGetAsyncContextSync,
   runWithAsyncCtx,
 } from "./als.ts";
 import {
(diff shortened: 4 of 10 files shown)


Linell (changes requested):
I think middleware that creates steps inside its own `wrapStep` loses recursion protection here?

The guard lookup uses whatever ALS context is active when the chain fires (manager.ts:335). For memoized steps that's the report batch (engine.ts:2604), usually the root context, and nothing ever stores a guard for root. With this shape, the middleware re-plans its injected steps on every replay and the run hangs at finalization, in both checkpointing modes:

```ts
override async wrapStep({ ctx, next, stepInfo }: Middleware.WrapStepArgs) {
  await ctx.step.run(`${stepInfo.options.id}-audit`, () => null);
  return next();
}
// fn body:
await Promise.all([step.run("a", () => 1), step.run("b", () => 2)]);
```

The old global Set caught this since it didn't care which context invoked the chain. Would capturing the declaring context on the `FoundStep` and running `wrappedHandler()` under it with `runWithAsyncCtx` work? And could the new test run under `matrixCheckpointing` like its neighbors?

Linell (approved):

PR #1671, opened 2026-08-03
feat: add SDK feature observations

amh4r:
## Summary

Add SDK feature observations so the JS SDK reports app-level feature readiness and configuration during app sync. This lets Inngest understand whether features such as AI metadata extraction, Extended Traces, and event sending are configured and ready.

There are no user-facing changes in this PR. We're only adding more data to the existing app sync payload.

https://github.com/inngest/inngest/pull/4700 contains the corresponding Dev Server changes.

## Changes

- Added feature observation protobufs and generated TS bindings.
- Added `sdkFeatureObservations` state/serialization in the SDK client.
- Reports observations through Connect config and app sync payloads.
- Records AI metadata and Extended Traces setup outcomes, including async OTel provider setup before app sync snapshots are sent.

## Context

[Spec](https://app.notion.com/p/inngest/SDK-feature-observability-3abb64753bbd80998bb5e9d4acb91bd5)


packages/inngest/src/components/execution/otel/provider.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  return {
+    path: setup.path,
+    providerFound,
+    providerSource,
+    addSpanProcessorAttempted,
+    spanProcessorAdded,
+    failure,
+  };
+}
```
scottnuma, line 22: It looks like you could use one of the generated fns for this:

```ts
function setupWith(
  setup: Readonly<Partial<OTelSetup> & Pick<OTelSetup, "path">>,
): OTelSetup {
  return OTelSetup.fromPartial(setup);
}
```

Less code and less drift!

packages/inngest/src/types.ts
```diff
@@ -1570,6 +1570,13 @@ export interface RegisterRequest {
    * Capabilities of the SDK.
    */
   capabilities: Capabilities;
+
+  /**
+   * Feature readiness and configuration facts observed by the SDK.
+   *
+   * Encoded with protobuf JSON mapping.
+   */
+  feature_observations?: unknown[];
```
scottnuma: We use snake case here, but camel case with its neighbors. Intentional break?

packages/inngest/src/components/sdkFeatureObservations.ts
```diff
@@ -0,0 +1,322 @@
...
+   * because registration snapshots are not refreshed when async setup, such as
+   * legacy Extended Traces provider creation, finishes later.
+   */
+  private async waitForReady(client: Inngest.Any): Promise<void> {
+    const pending = [...client[featureObservationStateSymbol].pending];
+    if (pending.length === 0) {
+      return;
+    }
+
+    await Promise.allSettled(pending);
```
scottnuma: My read is that this makes app registration and the Connect handshake wait on OTel provider setup, with no upper bound. With the `createProvider` option, auto-instrumentation now goes through the critical path for registration.

For the smallest change, we could add a short timeout.

Claude proposes a fix here which seems to thread some needle of trade offs. It's a bit complex, and I don't follow all of it, so the short timeout seems like the way to go, but including the idea:

------------------

I think there's a fix that removes the wait entirely rather than bounding it. `setGlobalTracerProvider` currently sits *behind* the dynamic imports (`util.ts:65`), which is the only reason "did the processor get added" is unknowable synchronously — but nothing about provider registration depends on those imports. `util.ts:29-33` already builds the processor and the `BasicTracerProvider` synchronously; only `registerInstrumentations` and `AsyncHooksContextManager` need the imported modules. Splitting it:

```ts
// synchronous — setup observation known immediately
function installProvider() {
  const processor = new InngestSpanProcessor();
  const p = new BasicTracerProvider({ spanProcessors: [processor] });
  const ok = trace.setGlobalTracerProvider(p);
  return { processor, setup: /* LEGACY_CREATE_PROVIDER, spanProcessorAdded: ok */ };
}

// fire-and-forget — attaches to the now-global provider
async function finishProviderSetup(instrumentations) { /* the four imports */ }
```

That lets `registerBody` and `prepareConnectionConfig` stay synchronous, reverting the async-signature churn in `config.ts` / `connect/index.ts` / `InngestCommHandler` rather than adding to it, and it retires `waitForReady` / `addPendingUpdate` / `state.pending` altogether. It also happens to fix a separate pre-existing bug: `metadataSpanProcessor.attach()` runs in the constructor (`Inngest.ts:466`) *before* the legacy provider exists, latches `NO_PROVIDER`, and is never retried (`attach()` has exactly one call site) — so AI metadata extraction is currently dead on the legacy path, and this PR would start reporting `OTEL_PROVIDER_MISSING` for those users. Registering the provider synchronously means the constructor finds it.

Tradeoff worth weighing: the provider goes live before `AsyncHooksContextManager` is enabled, so for the duration of the auto-instrumentations import, host-app spans created at module init are recorded but may be mis-parented — today they're dropped entirely. You can shrink that window to a single module resolution by awaiting only `@opentelemetry/context-async-hooks` (a hard dependency, tiny, no instrumentation side effects) before installing provider + context manager together, and letting the heavy import continue in the background.

Reporting a not-ready value is honest in that case — setup genuinely hadn't finished. Worth noting the two channels degrade differently: Connect re-sends on reconnect so it self-heals, but HTTP app sync fires once per deploy, so a first-sync miss is permanently wrong data.

scottnuma (approved):
This PR's quite large, so my grasp on the review is loose. Happy to focus in more on any specific parts if you like.

packages/inngest/src/components/execution/otel/provider.ts
amh4r: Thanks!

packages/inngest/src/types.ts
amh4r: Doh, good call. Fixed

packages/inngest/src/components/sdkFeatureObservations.ts
amh4r: I removed the async shenanigans in `sdkFeatureObservations.ts`. Ultimately it seemed overkill:
- Your concerns
- Provider creation is legacy (and will eventually be removed)
- The consequence of not having async seemed too theoretical. It should happen so fast that by the time we sync the app it'll be good

Linell (approved):

scottnuma (approved):

PR #1684, opened 2026-08-06
feat(sandbox): support shell command strings

jakobevangelista:
Lets sandbox commands accept shell strings, so the common case is just:

```ts
await sandbox.commands.run("git status && npm test");
```

Options still work:

```ts
await sandbox.commands.run("npm test", {
  cwd: "/workspace",
  environment: { CI: "true" },
  timeout: "5m",
});
```

And the API works the same way with a step ID:

```ts
await sandbox.commands.run("run-tests", "npm test", {
  cwd: "/workspace",
  timeout: "5m",
});
```

The SDK turns a string into `["/bin/sh", "-c", command]`. It does not split or parse it, so pipes, redirects, variables, quoting, and chained commands work normally. Existing argv arrays are passed through unchanged.

Managed process start accepts the same command strings:

```ts
await sandbox.processes.start({
  command: "npm run dev",
  cwd: "/workspace",
});
```

This is SDK-only. REST, middleware, and Simcity still receive argv arrays

### DIFF
--- a/packages/inngest/src/components/sandbox/types.ts
+++ b/packages/inngest/src/components/sandbox/types.ts
@@ -185,6 +185,10 @@ export interface SandboxProcessStartOptions {
 
 export interface SandboxProcessSignalOptions {
   signal: number;
+  /**
+   * Also signal the process's descendants. Defaults to `true` so shell-backed
+   * commands do not leave their child processes running.
+   */
   includeChildren?: boolean;
 }
 
--- a/packages/inngest/src/components/sandbox/types.ts
+++ b/packages/inngest/src/components/sandbox/types.ts
@@ -85,9 +85,17 @@ export interface SandboxListResult<TSandbox = SandboxRef> {
   fetchedAt: string;
 }
 
-export interface SandboxCommandOptions {
-  command: readonly string[];
+/**
+ * A shell command string or an argv array for direct execution.
+ *
+ * Strings run as `["/bin/sh", "-c", command]`. They are passed to the shell
+ * unchanged, so normal shell quoting, variables, pipes, redirects, and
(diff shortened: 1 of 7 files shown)


packages/inngest/src/components/sandbox/validation.ts
```diff
@@ -457,13 +471,17 @@ const normalizeProcessSpec = <
...
 };
 
 export const normalizeSandboxCommandOptions = (
-  options: SandboxCommandOptions,
-): SandboxCommandOptions & { timeoutMs: number } => {
-  const parsed = normalizeProcessSpec(options, true);
+  command: SandboxCommand,
+  options: SandboxCommandOptions = {},
+): SandboxCommandOptions & { command: string[]; timeoutMs: number } => {
+  const parsed = normalizeProcessSpec({ command, ...options }, true);
```
Linell: Doesn't `options` shadow `command` here? `{ command, ...options }` means a `command` key in the second arg wins over the first one, and the command you passed never runs. `options?: SandboxCommandOptions & { command?: never }` would make it a compile error instead.

packages/inngest/src/components/sandbox/types.ts
```diff
@@ -156,7 +164,7 @@ export interface SandboxProcessRef extends SandboxProcessResource {
 }
 
 export interface SandboxProcessStartOptions {
-  command: readonly string[];
+  command: SandboxCommand;
```
Linell, line 176: `sh -c "a && b"` sticks around as the parent now, I think? So a default SIGTERM just hits the shell and the real work keeps running. Is `includeChildren: false` still right? 🤔

Linell (approved):
🍰

PR #1696, opened 2026-08-10
chore: add ai middleware example app

Linell:
## Summary

Adds `examples/ai`, a TanStack Start app exercising `aiMiddleware()` from #1695. May not even be worth merging, but it was useful for testing!

Stacked on #1695.

Docs: inngest/website#1837.


scottnuma (approved):
Adding the example app seems useful. 

Can we a add a test that gives us an easy way of validating that the app works? Ideally without needing an OpenAI API key, though is fine if it does.

Doesn't need to run through CI, just a quick hand check that no bit-rot has occurred

PR #1700, opened 2026-08-11
fix(sandbox): return command output as text

jakobevangelista:
Captured sandbox command output should just be easy to use

This makes stdout and stderr UTF-8 strings for both direct and middleware-backed Exec. The REST and durable wire formats stay byte-safe/base64, and durable truncation still happens on the original bytes before decoding. Invalid or boundary-truncated UTF-8 uses replacement characters instead of throwing

also treat omitted empty `stdout` and `stderr` REST fields as empty strings, matching protobuf JSON's default-value omission (my bad, bug introduced from me)

### DIFF
--- a/.changeset/calm-sandboxes-speak.md
+++ b/.changeset/calm-sandboxes-speak.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Return captured sandbox command stdout and stderr as UTF-8 strings.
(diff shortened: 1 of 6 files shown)


Linell (approved):
Looks good, but it *does* mean we can't do binary output, right? Anything binary comes back as replacement characters with no way at the raw bytes. I think that's fine, just wanted to say it.

PR #1701, opened 2026-08-12
feat(sandbox): wait for running by default

jakobevangelista:
- Make direct and durable sandbox Create wait for `RUNNING` for up to 120 seconds by default.
- Allow `runningTimeout: false` to preserve the immediate `STARTING | RUNNING` response behavior.
- Preserve explicit duration overrides and reuse the existing readiness polling path.

Most callers use a sandbox immediately after creating it. Waiting by default removes an extra readiness step from the common path while retaining an explicit non-blocking opt-out.

Addresses [SYS-1121](https://linear.app/inngest/issue/SYS-1121/dx-sdk-create-should-be-blocking-by-default-w-runningtimeout-having-a).

### DIFF
--- a/.changeset/calm-sandboxes-wait.md
+++ b/.changeset/calm-sandboxes-wait.md
@@ -1,0 +1,5 @@
+---
+"inngest": minor
+---
+
+Wait for newly created sandboxes to reach RUNNING by default, with `runningTimeout: false` as an explicit opt-out.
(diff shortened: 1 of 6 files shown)


Linell (approved):
I think the `as SandboxDuration` cast in `normalizeSandboxCreateOptions` is avoidable? `normalizeDurationMs` already does the real runtime checking, so the unknown could flow straight through. Not blocking, just allergic to casts.

Issue #1440, opened 2026-04-04
[BUG] V1 Execution state is never cleared after completion, causing memory leak


fdarian:
# V1 Execution state is never cleared after completion, causing memory leak

**Describe the bug**

`V1InngestExecution` never clears its internal state after execution completes. The `_start()` finally block calls `this.state.loop.return()` but does not clear `this.state.steps`, `this.state.stepState`, `this.fnArg`, or `this.execution`. This means each execution instance — along with all step closures, arguments, and deferred promises — remains in memory indefinitely until the process is restarted.

For long-running services with cron-triggered functions, this produces a staircase memory growth pattern (~200–300 MB/hour in our case) that eventually causes OOM.

**Root cause**

Three retention mechanisms prevent `V1InngestExecution` from being garbage collected:

**1. `state.steps` Map is never cleared**

In `createStepTools()`, each step is stored in `this.state.steps` with its full closure arguments:

```typescript
// v1.ts — createStepTools()
const step = {
  rawArgs: fnArgs,       // full closure arguments (user code + framework runtime)
  fn: opts?.fn ? () => opts.fn?.(this.fnArg, ...fnArgs) : undefined,
  promise,               // deferred promise
  handle: () => { ... }, // closure over resolve, reject, this.state
};
this.state.steps.set(hashedId, step);
```

This Map is **never `.clear()`'d or nulled** after `_start()` completes. Every step's `rawArgs` and `fn` closures retain whatever the user's callback captured.

**2. `this.execution` memoization creates a self-reference cycle**

```typescript
// v1.ts — start()
this.execution = getAsyncLocalStorage().then((als) => {
  return als.run({
    app: this.options.client,
    execution: {
      ctx: this.fnArg,
      instance: this,  // ← strong reference back to the execution instance
    },
  }, async () => {
    return this._start().then(...).finally(...);
  });
});
```

`this.execution` is set once and never nulled. The ALS context includes `instance: this`, creating a cycle. Even after the Promise settles, the cycle keeps the instance reachable from itself.

**3. `createDeferredPromiseWithStack()` async generator survives `return()`**

```typescript
// helpers/promises.ts
const results = (async function* () {
  while (true) {
    const next = settledPromises.shift();
    if (next) yield next;
    else await new Promise<void>((resolve) => {
      rotateQueue = resolve;  // ← captured resolve function
    });
  }
})();
```

[...]

PR #1703, opened 2026-08-12
fix(otel): don't track non-recording spans in the metadata processor

### LINKED ISSUE #1702: [BUG] InngestMetadataSpanProcessor retains an execution engine per run when traces are sampled below 100%
# [BUG] `InngestMetadataSpanProcessor` retains an execution engine per run when traces are sampled below 100%

**SDK version:** 4.13.0 (verified present and unchanged through 4.18.0)
**Runtime:** Node 24 on DigitalOcean App Platform
**Tracing:** `@sentry/remix` with OTel, `tracesSampleRate` ~0.01

## Describe the bug

With `aiMetadata` enabled (the default), `metadataSpanProcessor` seeds its `#spanSinks` Map with a callback that captures the `InngestExecutionEngine`. The entry is only removed when the span **ends** — but OTel does not deliver `onStart`/`onEnd` to span processors for **non-recording** spans. `declareStartingSpan` has no sampling guard, so when the host application samples traces below 100%, entries are seeded for spans the processor will never be told about, and they are never removed.

The processor is attached to the process-global OTel tracer provider, so this is a process-lifetime root. **One execution engine is retained per run, indefinitely.**

We hit this in production: the service OOM-crashed every ~2 hours with `FATAL ERROR: Ineffective mark-compacts near heap limit`, growing ~9–10 MB/min.

## Evidence

Heap snapshot from a live 53-minute-old pod. **6,560 `InngestExecutionEngine` instances alive**, every one held by the same path:

```
global
 → Symbol(opentelemetry.js.api.1) → trace: ProxyTracerProvider
 → _delegate: BasicTracerProvider
 → _registeredSpanProcessors[1]: InngestMetadataSpanProcessor
 → #spanSinks: Map
 → closure onAIMetadata
 → (closure scope) → this: InngestExecutionEngine
```

Each retained engine drags its whole graph along. From the same snapshot:

| count | object |
|---:|---|
| 6,560 | `InngestExecutionEngine` |
| 85,243 | `wrappedMatchOp` (from `InngestStepTools`) |
| 26,232 | `AsyncGenerator` |
| 26,220 | retained request URLs (`/api/inngest?fnId=…`) |
| ~19,671 each | `steps-found`, `function-resolved`, `function-rejected`, `checkpointing-runtime-reached`, `checkpointing-buffer-interval-reached` |
| 1,015,514 | plain `Object` (50.6 MB) |
| 497,918 | closure `Context` (25.2 MB) |
| 118,088 | `Promise` |

Roughly 3 tracked spans per run, consistent with a root span plus children picked up by `onStart`.

## Root cause

`components/execution/engine.js` (`start()`) — the closure captures `this`:

```js
if (this.options.client.aiMetadataEnabled) metadataSpanProcessor.declareStartingSpan({
  span,
  traceparent: this.options.headers[headerKeys.TraceParent],
  onAIMetadata: (aiMetadata) => {
    if (!this.state.executingStep) return;          // <- captures the engine
    this.state.executingStepAIMetadata = this.state.executingStepAIMetadata
      ? aggregate(this.state.executingStepAIMetadata, aiMetadata)
      : aiMetadata;
  }
});
```

`components/execution/otel/metadataProcessor.js` — `declareStartingSpan` checks only `#attached` and `traceparent`, never whether the span is recording:

```js
declareStartingSpan({ span, traceparent, onAIMetadata }) {
  if (!this.#attached) return;
  if (!traceparent) return processorDevDebug(...);
  this.trackSpan(span, onAIMetadata);   // seeds #spanSinks unconditionally
}
```

### Why the two cleanup paths don't fire

1. **`onEnd()`** — never called. The span is a `NonRecordingSpan`, so OTel does not notify processors. `span.end()` in the engine's `.finally()` is a no-op for it.

2. **The `FinalizationRegistry` backstop** — cannot fire, because the retention is self-sustaining. The `#spanSinks` entry keeps the sink alive → the sink keeps the engine alive → the engine's pending execution keeps the span reachable. The span is therefore never collected, so the finalizer that would delete the entry never runs. The only thing that would free the span is deleting the entry, which is what the finalizer was supposed to do.

## To reproduce

1. Register an OTel/Sentry tracer provider with a sample rate below 1.0 (we use ~0.01).
2. Leave `aiMetadata` at its default (enabled).
3. Drive sustained function execution — ours was ~10 runs/sec/instance.
4. Watch `process.memoryUsage().heapUsed` climb linearly and never recover. Take a heap snapshot and inspect retainers of `InngestExecutionEngine`.

## Expected behaviour

Completed runs should be collectable. Entries seeded in `#spanSinks` should not outlive the run regardless of the host application's sampling configuration.

## Suggested fixes

1. **Guard on recording state** — the cheapest fix. If the span is not recording, the processor will never receive `onEnd` for it, so tracking it can only leak:

   ```js
   declareStartingSpan({ span, traceparent, onAIMetadata }) {
     if (!this.#attached) return;
     if (!traceparent) return;
     if (!span.isRecording()) return;   // processors get no lifecycle events for these
     this.trackSpan(span, onAIMetadata);
   }
   ```

   The same guard belongs in `trackSpan`, which `onStart` also calls.

2. **Break the finalizer cycle.** Even with (1), the `FinalizationRegistry` backstop is ineffective whenever the sink transitively retains the span. Holding the sink weakly, or storing something that doesn't reference the engine, would make the backstop actually work.

3. **Bound the Map** — a size or TTL cap would turn an unbounded leak into a bounded one as defence in depth.

## Workaround

Setting `aiMetadata: false` on the client stops it completely, for anyone not using the AI features:

```js
new Inngest({ id: "my-app", aiMetadata: false });
```

## Related

Possibly compounding, but distinct: #1440 (`V1InngestExecution` never clearing `state.steps` / `this.execution`). That issue explains why each retained engine is so expensive; this one explains what roots them globally. Fixing #1440 alone would not release these, since the retaining reference here is the global `#spanSinks` Map.

deanmalmgren:
Fixes #1702.

## Problem

`InngestMetadataSpanProcessor` seeds `#spanSinks` in `declareStartingSpan` with a callback that captures the `InngestExecutionEngine`, and only removes the entry when the span **ends**. OTel does not deliver `onStart`/`onEnd` to span processors for **non-recording** spans, so when the host application samples traces below 100% the entry is never removed — and because the processor is attached to the process-global tracer provider, **one execution engine is retained per run, for the lifetime of the process.**

The `#spanCleanup` `FinalizationRegistry` cannot recover it either. The retention is self-sustaining: the `#spanSinks` entry keeps the sink alive → the sink keeps the engine alive → the engine's pending execution keeps the span reachable. The span is therefore never collected, so the finalizer that would delete the entry never runs. The only thing that would free the span is deleting the entry, which is what the finalizer was supposed to do.

We hit this in production on 4.13.0 with `@sentry/remix` at `tracesSampleRate` ~0.01. The service OOM-crashed every couple of hours with `FATAL ERROR: Ineffective mark-compacts near heap limit`, growing ~9–10 MB/min. A heap snapshot from a live pod showed **6,560 `InngestExecutionEngine` instances alive** after 53 minutes, all held by:

```
global → Symbol(opentelemetry.js.api.1) → ProxyTracerProvider
      → BasicTracerProvider._registeredSpanProcessors[1]
      → InngestMetadataSpanProcessor → #spanSinks (Map)
      → closure onAIMetadata → this: InngestExecutionEngine
```

Each retained engine drags its whole graph with it — 85,243 `wrappedMatchOp`, 26,232 `AsyncGenerator`, 118,088 `Promise`, ~1M plain `Object`. Full analysis in #1702.

## Fix

`trackSpan` now skips non-recording spans.

Tracking them was never useful: the processor is never told when such a span starts or ends, so its sink could never fire and no AI metadata was ever collected for it. This only removes entries that could not have been acted on — for anyone sampling at 100%, behaviour is unchanged.

The guard lives in `trackSpan` rather than `declareStartingSpan` so it covers the `onStart` path too, though in practice OTel already filters that one.

## Tests

Three added to `metadataProcessor.test.ts`:

- **`does not track non-recording spans`** — the regression test. Uses `AlwaysOffSampler`, declares 100 roots, asserts nothing is retained. Without the fix this fails with `expected 100 to be +0`.
- **`tracks a declared root while it is open, and releases it on end`** — pins the normal lifecycle so the guard can't over-reach and break tracking for recording spans.
- **`does not accumulate entries across many completed runs`** — 100 full run cycles at 100% sampling leave the map empty.

Asserting this needs visibility into `#spanSinks`, which is a true private field, so I added a `trackedSpanCount` getter marked `@internal`. Happy to drop it for a different approach if you'd rather not widen the surface at all.

`pnpm test` on the file passes 14/14 and `biome check` is clean.

## Notes

- Present and unchanged in 4.13.0 through 4.18.0.
- Workaround for anyone affected before this lands: `new Inngest({ id, aiMetadata: false })`.
- Possibly compounding, but distinct: #1440 (`V1InngestExecution` not clearing `state.steps` / `this.execution`). That issue explains why each retained engine is so expensive; this one explains what roots them globally. Fixing #1440 alone would not release these.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/olive-donkeys-shave.md
+++ b/.changeset/olive-donkeys-shave.md
@@ -1,0 +1,7 @@
+---
+"inngest": patch
+---
+
+Stop `InngestMetadataSpanProcessor` retaining an execution engine per run when traces are sampled below 100%.
+
+Span processors only receive `onStart`/`onEnd` for recording spans, so tracking a non-recording span seeded an entry in `#spanSinks` that could never be removed, retaining the `AIMetadataSink` closure and the `InngestExecutionEngine` it captures for the lifetime of the process. `trackSpan` now skips non-recording spans.
--- a/packages/inngest/src/components/execution/otel/metadataProcessor.test.ts
+++ b/packages/inngest/src/components/execution/otel/metadataProcessor.test.ts
@@ -1,5 +1,8 @@
 import { type Attributes, context, trace } from "@opentelemetry/api";
-import { BasicTracerProvider } from "@opentelemetry/sdk-trace-base";
+import {
+  AlwaysOffSampler,
+  BasicTracerProvider,
+} from "@opentelemetry/sdk-trace-base";
 import type { AIMetadata } from "./aiExtractor.ts";
 import { InngestMetadataSpanProcessor } from "./metadataProcessor.ts";
 
@@ -256,5 +259,65 @@ describe("InngestMetadataSpanProcessor", () => {
     expect(processorsOf(provider).filter((p) => p === processor)).toHaveLength(
       1,
     );
+  });
+
+  /**
+   * Same wiring as {@link setup}, but with a sampler that records nothing —
+   * the shape of any host app tracing at less than 100%.
+   */
+  function setupUnsampled() {
+    const processor = new InngestMetadataSpanProcessor();
+    const provider = new BasicTracerProvider({
+      sampler: new AlwaysOffSampler(),
+    });
+    trace.setGlobalTracerProvider(provider);
+    processor.attach();
+    return { processor, tracer: provider.getTracer("test") };
+  }
+
+  test("tracks a declared root while it is open, and releases it on end", () => {
+    const { processor, tracer } = setup();
+    const { root } = declaredRoot(processor, tracer);
+
+    expect(processor.trackedSpanCount).toBe(1);
+
+    root.end();
+
+    expect(processor.trackedSpanCount).toBe(0);
+  });
+
+  test("does not track non-recording spans", () => {
+    const { processor, tracer } = setupUnsampled();
+
+    // Processors are never handed onStart/onEnd for non-recording spans, so a
+    // tracked entry here could never be removed: it would retain the sink, and
+    // through it the execution engine, for the lifetime of the process.
+    for (let i = 0; i < 100; i++) {
+      const root = tracer.startSpan("inngest.execution");
+      expect(root.isRecording()).toBe(false);
+      processor.declareStartingSpan({
+        span: root,
+        traceparent: TRACEPARENT,
+        onAIMetadata: () => {},
+      });
+      root.end();
+    }
+
+    expect(processor.trackedSpanCount).toBe(0);
+  });
+
+  test("does not accumulate entries across many completed runs", () => {
+    const { processor, tracer } = setup();
+
+    for (let i = 0; i < 100; i++) {
+      const { root } = declaredRoot(processor, tracer);
+      endChild(tracer, root, "llm", {
+        "gen_ai.request.model": "gpt-4.1-nano",
+        "gen_ai.usage.input_tokens": 1,
+      });
+      root.end();
+    }
+
+    expect(processor.trackedSpanCount).toBe(0);
   });
 });
--- a/packages/inngest/src/components/execution/otel/metadataProcessor.ts
+++ b/packages/inngest/src/components/execution/otel/metadataProcessor.ts
@@ -133,11 +133,37 @@ export class InngestMetadataSpanProcessor implements SpanProcessor {
    * are stamped on the span.
    */
   private trackSpan(span: Span, sink: AIMetadataSink): void {
+    // OTel only delivers `onStart`/`onEnd` to span processors for recording
+    // spans, so tracking a non-recording one can only ever leak: `onEnd` will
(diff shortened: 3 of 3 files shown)


jacobheric:
Thanks for the fix! Team merged a fix with some additional cleanup and defense [here](https://github.com/inngest/inngest-js/pull/1704). Since that's been merged I'll close this one.

PR #1704, opened 2026-08-12
fix: prevent ai metadata span retention

jacobheric:
## Summary

skip ai metadata tracking for spans that otel marks as non-recording. addresses: #1702

### DIFF
--- a/.changeset/calm-spans-rest.md
+++ b/.changeset/calm-spans-rest.md
@@ -2,4 +2,4 @@
 "inngest": patch
 ---
 
-Prevent AI metadata tracking from retaining executions when OpenTelemetry samples out their root spans.
+Prevent AI metadata tracking from retaining executions after an SDK request ends, including when OpenTelemetry samples out root spans.
--- a/.changeset/calm-spans-rest.md
+++ b/.changeset/calm-spans-rest.md
@@ -2,4 +2,4 @@
 "inngest": patch
 ---
 
-Prevent AI metadata tracking from retaining executions after an SDK request ends, including when OpenTelemetry samples out root spans.
+Prevent AI metadata tracking from retaining executions after an SDK request ends, including when OpenTelemetry excludes the root span from sampling.
--- a/.changeset/calm-spans-rest.md
+++ b/.changeset/calm-spans-rest.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Prevent AI metadata tracking from retaining executions when OpenTelemetry samples out their root spans.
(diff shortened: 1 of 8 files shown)


packages/inngest/src/components/execution/otel/metadataProcessor.ts
```diff
@@ -75,6 +96,32 @@ export class InngestMetadataSpanProcessor implements SpanProcessor {
    */
   #attachSetup: OTelSetup | undefined;
 
+  /**
+   * Start a logical scope for AI metadata tracking.
+   */
+  public startScope(): AIMetadataScope {
+    const scope = Symbol("inngest.ai-metadata-scope");
```
scottnuma: nit: for debugging, it might be useful to have a more descriptive symbol. I think runId and span context are available when we call startScope and could seed info for a unique string.

.changeset/calm-spans-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Prevent AI metadata tracking from retaining executions after an SDK request ends, including when OpenTelemetry samples out root spans.
```
scottnuma: I think OTel usually thinks of sampling as including the trace in reporting.

```suggestion
Prevent AI metadata tracking from retaining executions after an SDK request ends, including when OpenTelemetry excludes the root span from sampling.
```

packages/inngest/src/components/execution/otel/metadataProcessor.ts
```diff
@@ -132,21 +181,38 @@ export class InngestMetadataSpanProcessor implements SpanProcessor {
...
+      return;
+    }
+
     const { traceId, spanId } = span.spanContext();
     const key = spanSinkKey(traceId, spanId);
 
     this.#spanCleanup.register(span, key, span);
-    this.#spanSinks.set(key, sink);
+    this.#spans.set(key, { sink, scope });
+    this.#scopeSpanKeys.get(scope)?.add(key);
```
scottnuma: scopeSpanKeys is part of our back stop to clear out the spans.

What do you think of only setting `this.#spans.set(key, { sink, scope });` when `this.#scopeSpanKeys.get(scope)?.add(key);` will succeed?

The two will always succeed together today, but if we accidentally break something in the future, we currently bias towards keeping the sink vs. protecting against leaks. Similarly reporting data vs. OOMs.

I don't have strong takes here, more discussing.

```suggestion
    const keys = this.#scopeSpanKeys.get(scope);
    # bail if its missing
    if (!keys) return;

    const { traceId, spanId } = span.spanContext();
    const key = spanSinkKey(traceId, spanId);

    this.#spanCleanup.register(span, key, span);
    this.#spans.set(key, { sink, scope });
    keys.add(key);
```

scottnuma (commented):
getting things into a better state!

scottnuma (approved):

PR #1705, opened 2026-08-12
feat(sandbox): allow non-absolute commands

jakobevangelista:
- allow sandbox commands to use bare executable names
- allow cwd-relative executable paths
- retain empty, NUL, Unicode, and size validation

This exposes the execvp-style semantics implemented by inngest/simcity#45 through the cloud API change in inngest/monorepo#8209.

### DIFF
--- a/.changeset/calm-tools-exec.md
+++ b/.changeset/calm-tools-exec.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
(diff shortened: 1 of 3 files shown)


Linell (approved):
Approved, just a heads up: it looks like https://github.com/inngest/monorepo/pull/8209 needs to land before this is safe to release? Otherwise bare commands pass validation here and then 400 on the API.

PR #1707, opened 2026-08-15
fix(execution): throw NonRetriableError when step.invoke target is cancelled

### LINKED ISSUE #1694: [BUG] Cancelled function invoke tasks resolve with wrong type
**Describe the bug**

When functions are run via `await step.invoke(…)` and the run is cancelled in the Dev UI, the promise still resolves but with the wrong data.

**To Reproduce**

```ts
const doThing1 = inngest.createFunction(
  {
    id: `doThing1`,
    triggers: eventType(`doThing1.run`, { schema: z.object() }),
  },
  async ({ step }): Promise<void> => {
    const result = await step.invoke(`invoke another function`, {
      function: doThing2,
      data: {},
    });

    z.object({ text: z.string() }).parse(result);
  },
);

const doThing2 = inngest.createFunction(
  {
    id: `doThing2`,
    triggers: eventType(`doThing2.run`, { schema: z.object() }),
  },
  async ({ step }): Promise<{ text: string }> => {
    await step.sleep(`sleep for 1 minute`, 60_000);
    
    // Now go into the dev UI and cancel this task before it resolves.

    return { text: "hello world" };
  },
);
```

1. Run `doThing1`.
2. Wait for it to invoke `doThing2`.
3. Cancel the `doThing2` task before the 60s sleep expires.

**Expected behavior**
I expected that either the `await step.invoke()` would throw a `NonRetriableError` or the return type of `step.invoke()` would return a wrapper object like `{ data?: …, success: boolean }`, so that if it's cancelled you. can handle it appropriately.

Instead the `step.invoke()` resolves with an object like this:

```json
{
  "data": {
    "_inngest": {
      "status": "Cancelled"
    },
    "correlation_id": "01KZMH2MWC1TKB1MAFKVKB2PWB.cf3ce4fd5ffcd8bc52a0516436f293bf27d93e2f",
    "event": {
      "data": {
        "_inngest": {
          …
        }
      }
    }
  },
  "id": "01KZMH3G9VEC5FW3SPH723PJR2",
  "name": "inngest/function.finished",
  "ts": 1786321682747,
  "user": {}
}
```

Notice the `data._inngest.status`

- OS: Mac
- npm package Version: 4.13.0
- Framework: Bun

**Additional context**

marsyg:
## Summary
When an invoked child function was cancelled in the Dev UI or via API, `step.invoke(...)` previously resolved with the internal `inngest/function.finished` event payload (with `data._inngest.status === "Cancelled"`) instead of rejecting. This broke TypeScript return type assumptions and caused runtime crashes when trying to parse or use the return value.

This PR normalizes memoized step states in `engine.ts` (`normalizeMemoizedOp`) so that cancelled invocations set `data: undefined` and `error: { name: "NonRetriableError", message: "Invoked function was cancelled" }`. This ensures:
1. `step.invoke(...)` rejects with a `StepError` (`NonRetriableError`).
2. Parent functions can catch and handle cancellations cleanly using `try { await step.invoke(...) } catch (err)`.
3. Uncaught cancellations immediately fail the parent run without wasteful retries.

## Checklist
- [ ] ~~Added a docs PR that references this PR~~ N/A Bug fix aligning with existing NonRetriableError step behavior

## Related
Fixes #1694

### DIFF
--- a/.changeset/soft-walls-wonder.md
+++ b/.changeset/soft-walls-wonder.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Reject step.invoke with NonRetriableError when invoked function is cancelled
--- a/packages/inngest/src/components/execution/engine.ts
+++ b/packages/inngest/src/components/execution/engine.ts
@@ -136,6 +136,48 @@ function errorMessage(error: unknown): string {
   return String(error);
 }
 
+function normalizeMemoizedOp<T extends MemoizedOp>(op: T): T {
+  if (typeof op.data === "object" && op.data !== null) {
+    const dataObj = op.data as Record<string, unknown>;
+    const isFinishedEvent = dataObj.name === "inngest/function.finished";
+    const eventData = isFinishedEvent
+      ? (dataObj.data as Record<string, unknown> | undefined)
+      : undefined;
+
+    const status =
+      eventData?._inngest &&
+      typeof eventData._inngest === "object" &&
+      eventData._inngest !== null
+        ? (eventData._inngest as Record<string, unknown>).status
+        : dataObj._inngest &&
+            typeof dataObj._inngest === "object" &&
+            dataObj._inngest !== null
+          ? (dataObj._inngest as Record<string, unknown>).status
+          : undefined;
+
+    if (status === "Cancelled") {
+      return {
+        ...op,
+        data: undefined,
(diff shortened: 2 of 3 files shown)


marsyg:
Hi @amh4r @scottnuma would you be able to review this change?

PR #1714, opened 2026-08-29
Parse streamed SSE per the spec (CRLF + optional space)

eeshsaxena:
### What

`iterSse` (the SSE parser behind experimental durable-endpoint streaming) split events on `"\n\n"` and matched the literal prefixes `"event: "` and `"data: "`. Both assume LF line endings and a space after the colon, so two spec-valid shapes were dropped silently:

- **CRLF streams.** With `\r\n` line endings, events are separated by `\r\n\r\n`, which does not contain `\n\n`. `buffer.split("\n\n")` never splits, so the buffer grows and no event is ever yielded.
- **No space after the colon.** The space after an SSE field's colon is optional. `data:{"x":1}` fails `startsWith("data: ")` and is ignored, and `event:foo` falls back to the default `message` event, so the event is mis-typed and then dropped by `parseSseEvent`.

For a durable-execution stream, a dropped event can mean a lost step result.

### Repro

```ts
// current behaviour
[...iterSse('event: x\r\ndata: {"ok":1}\r\n\r\n')]  // []  (event lost)
[...iterSse('event:x\ndata:{"ok":1}\n\n')]          // [{ event: "message", data: "" }]
```

### Fix

Split events on `/\r?\n\r?\n/`, and parse each line as `field` + `value` with a single optional leading space stripped (and skip `:`-prefixed comment lines), per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation). Canonical `\n` + space payloads (what the SDK's own `buildSseEvent` emits) parse identically.

### Tests

`iterSse` had no test coverage; this adds `streaming.test.ts` covering canonical LF, CRLF, no-space fields, multi-line `data`, comment lines, and an event split across read boundaries. All pass under `vitest`, and `biome check` is clean.

### DIFF
--- a/.changeset/spec-compliant-iterSse.md
+++ b/.changeset/spec-compliant-iterSse.md
@@ -1,0 +1,5 @@
+---
+"inngest": patch
+---
+
+Parse streamed SSE responses per the Server-Sent Events spec: handle CRLF line endings and an optional (or absent) space after a field's colon, so `\r\n`-delimited streams and `data:value`/`event:value` lines are no longer silently dropped.
--- a/packages/inngest/src/components/execution/streaming.test.ts
+++ b/packages/inngest/src/components/execution/streaming.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, test } from "vitest";
+import { iterSse, type RawSseEvent } from "./streaming.ts";
+
+/**
+ * Build a `ReadableStream<Uint8Array>` from one or more string chunks, so a
+ * single logical SSE payload can be split across `reader.read()` boundaries.
+ */
+const streamOf = (...chunks: string[]): ReadableStream<Uint8Array> => {
+  const encoder = new TextEncoder();
+  return new ReadableStream({
+    start(controller) {
+      for (const chunk of chunks) {
+        controller.enqueue(encoder.encode(chunk));
+      }
+      controller.close();
+    },
+  });
+};
+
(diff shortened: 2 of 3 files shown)


Linell:
Thanks for the PR! Have you hit this against a real stream? A repro against an actual durable endpoint (or whatever sits in front of it) would help a lot. The SDK produces this stream itself, and the executor never emits CRLF or drops the space.

`iterSse` is intentionally bare since we control both ends and only need to match `buildSseEvent`. If a proxy or platform is rewriting the body somewhere, though, I'd want to know about it 🤔

eeshsaxena:
Honestly, no, I don't have a repro against a live durable endpoint. I came at this from the SSE spec rather than an observed failure, so if you control both ends today then yeah, it's defensive hardening rather than a fix for something you're actually seeing.

The case I had in mind isn't the executor, it's an intermediary. Some proxies/CDNs and a couple of the serverless platforms normalize line endings or re-chunk response bodies, and if `\r\n` ever ends up in the stream the `split("\n\n")` just quietly stops yielding. That's a silent hang rather than a loud error, which is the part that felt worth guarding against since it'd be painful to debug in the wild.

But if there's no non-inngest producer in the picture and you'd rather keep `iterSse` minimal, I'm happy to close it. Your call, no strong feelings either way.

PR #1723, opened 2026-09-13
fix(inngest): guard timeStr against NaN and invalid string duration inputs

Tyagiquamar:
In timeStr(), string inputs were passed to ms(input). When an unparseable or invalid duration string (e.g. 'invalid' or empty string) was passed, ms returned undefined or NaN.

Previously, timeStr did not explicitly check for NaN or non-number millisecond values before proceeding into periods.reduce, causing unexpected empty string returns or invalid arithmetic.

This fix adds an explicit check for typeof milliseconds !== 'number' || Number.isNaN(milliseconds) || milliseconds <= 0 to return empty string safely for invalid or non-positive durations, and floors fractional milliseconds before unit reduction.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

